PR #684, opened 2022-04-05
Fixing dependabot failures

anvabr:
Fixing for example this https://github.com/hashgraph/guardian/runs/5753909874?check_suite_focus=true. Using this explanation https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions and this doc here: https://github.com/EnricoMi/publish-unit-test-result-action

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,6 +88,9 @@ jobs:
         env:
           CI: true
       - name: Publish Unit Test Results
+        permissions:
+          checks: write
+          pull-requests: write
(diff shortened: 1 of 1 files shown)


anvabr:
This is not correct, the syntax is reported as wrong. Will use default permissions as described here: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

PR #768, opened 2022-04-20
Update send block property

Stepan-Kirjakov:
add 'Topic Owner' in sendToGuardianBlock 
shows which field determines the User who will own the topic of the document


Stepan-Kirjakov:
Update send block property

Stepan-Kirjakov:
add 'Topic Owner' in sendToGuardianBlock
shows which field determines the User who will own the topic of the document

PR #863, opened 2022-05-13
feat: upgrade typeorm, add healthcheck endpoint

samuraitruong:
Signed-off-by: Truong Nguyen <samuraitruong@hotmail.com>

What are changes:

When running guardian stack on production we need away to healthcheck the container and restart if need. It will be done by k8s live probe.  

- Add /healthcheck on all services, it will run by express and simple expose the http api. when the healthcheck got triggered, it will also test if the MQ is up and running
- During the refactor the server setup to core package and realised that we are using typeorm v2  so I decided to upgrade them to guardian v3. as result, v3 has mark all the global use of `getMongoRepository` being deprecated so I have update the code to using DataSource and mongoRepository create from DataSource instance. This is the good change and going forward we can have the code easier to maintain and update with lasted update

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
- typeorm v3 typing alway requires to pass filter condition under `where` property
- typeorm v3 remove findOne(string) function. I have udpated all the code but please review careful 

**Checklist**


prernaadev01:
@samuraitruong We have looked at this. Currently, typeorm supports MongoDB version 3 but we need MongoDB version 4. hence, we need to wait for the supported version. Thanks

anvabr:
To add to what @prernaadev01 said above, while the maintainers of the typeorm have supposedly [fixed](https://github.com/typeorm/typeorm/pull/8412) the [issue](https://github.com/typeorm/typeorm/issues/7907), they have not updated the 'supported' configuration files.

samuraitruong:
@anvabr I  have test the change and don't see nay issue with `find` function. however, I will update this PR to include the healthcheck change only and revert the typeorm. will update the code soon

prernaadev01:
@samuraitruong Thank you for the PR. Will check this and keep you posted.

samuraitruong:
@anvabr  are you talking about the server bootstrap function. Yes, that is intentional, at the moment if you look at the entry file of each service you see the duplicate code every where in main.ts. I just move all the function requires for the service into server.ts to clean up that duplication so we can maintain all service in 1 place. of course each service will have some extends plug-in after server start up and it still in the main.ts

For example , if you change the db connection or if you change the healthcheck if you dont need repeat it for 4-5 services.

anvabr:
@samuraitruong after careful consideration we decided not to accept refactored settings  as we believe this does not have the same benefits for the broader community. We are however are interested in the healthchecks - could you please separate them into a stand-alone PR? Thank you for your contributions!

PR #1037, opened 2022-06-27
fix: add queue to support horizontal scale

samuraitruong:
**Description**:
In production environment, to guarantee the HA system we run each service  with at least 2 instances(pod, nodes). Nats has design `Node Groups` to avoid the message being processed by multiple subscribers. 

Please refer to document here : 
https://github.com/nats-io/nats.js#queue-groups

**Related issue(s)**:

Fixes #

**Checklist**


samuraitruong:
@anvabr  this is very important for our production environment. can you please have a look. thanks

PR #1070, opened 2022-07-06
Fixes schema update documentation

justynspooner:
The current docs are incorrect for the update schema endpoint.

- `GET` should be `PUT`
- The `id` of the schema to be updated should be defined in the JSON body of the schema object, not in the URL parameters.

```
{
    "id": "<SCHEMA_ID>",
    ...schema
}
```

**Checklist**
I've not been able to verify the following but the changes are so small that I assume there will be no issues:
- [ ] Test that the updated swagger template strings render correctly in the UI


prernaadev01:
@justynspooner Thank you for correcting the documentation. We will consider your PR and integrate it. We will make these changes live in our upcoming releases.

PR #1774, opened 2023-02-07
Hotfix/fix logform dep

simvalery:
**Description**:
Preparing a hotfix resolving issues in https://github.com/winstonjs/logform/issues/245


wbt:
Note that the issue in Logform was patched in 2.5.1.

Issue #1661, opened 2023-01-10
Tagging of Guardian policy artefacts


anvabr:
### Problem description
Currently artefacts produced by Guardian instances are discoverable on-chain (tokens, topic messages) and in the decentralised storage (VCs/VPs). These systems are by definition immutable, therefore there is no way to additionally mark these artefacts beyond what has been produced at the production time. This makes it too rigid, given the long timeframes of their lifecycle it is highly likely that these artefacts will be used in situations/use-cases not currently envisioned. There is a need for a generally accepted practice of creating informative tags for various items, which can then power discoverability, analytics and reputation mechanics.

### Requirements
Create a tagging system with the following properties:
- tags can be created by any actor with a valid DID.
- created tags should be credentialed, i.e. it is clear who and when created or added to a particular tag.
- tags can have cumulative scores, i.e. two people independently creating/assigning the same tag to the same artefact results in both tagging actions recorded (credentialed etc), and the 'tag score' is then counted as 2.
- it is possible for users to untag the item, however this action does not remove the record of tagging in the first place it just records the action of untagging (also credentialed), and reduces to score by 1. 
- tags should be associate-able with any identifiable entity/thing/artefact that guardian uses and/or produces, such as:
   - actors, including SRs (i.e. DIDs),
   - schemas
   - policies and policy modules
   - VCs/VPs, 
   - tokens, 
   - smart contracts (i.e. addresses).
- tags can be created after the (immutable) artefacts are produced, therefore tags are external to artefacts
- it should be possibly for DIDs to create and/or follow tag ontologies containing the definition of description of tags
- tag registry[-ies] should be discoverable and useable by automated indexing systems 

### Definition of done

- Tagging system and tools are created, satisfying requirements above
- The design and usage workflow is documented
- Demo policies and various other related artefacts are tagged appropriately - to demonstrate the usage of the tags and also enable users to easily find/identify them as demo artifacts

### Acceptance criteria
Community actors can employ tagging for managing their relationships with various Guardian artefacts.

dubgeis:
@anvabr please fix spelling

prernaadev01:
@dubgeis Updated. Thanks

PR #1974, opened 2023-04-03
feat: api contract review

felipeneuhauss:
**Description**:

This PR aims to apply enhancements in the API such as the name convention for routes, the proper status code for the async processes, and improve code quality to manage errors and validations.  

The swagger file was updated to attend to the changes, and the tests were also updated.

Quick notes:

- Async functions are now returning 202
- Yup library was added to handle validations - In terms of bundle, the size is smaller than Joi's. 
- /trustchains route was changed to /trust-chains.
- Routes like that `/:tokenId/:username/grantKyc` was changed to `/:tokenId/:username/grant-kyc` for the sake of [name conventions. _**Topic 2.2.5**_](https://restfulapi.net/resource-naming/)
- A new handle code was created according to the Express conversions. 
- The **frontend** service links were changed to work properly with the new features


anvabr:
@felipeneuhauss thank you for submitting this PR, we reviewed its content and are generally happy with it - it did not break the 'old' API which, once the release is made, could be marked as deprecated and then retired appropriately. However we have merged code for #1661, which contained new API additions very soon after your PR was made. This new 'tagging' APIs are written to the old spec, which you were trying to change with your PR. Please can you add to your PR the adjustments required to make this new 'tagging' API conform to your naming design/convention. If done now it would avoid having to follow the deprecation policy (and have a deprecation period etc) as 'tagging' API has not been released yet.

Also could you please add a short document describing the naming convention/style which you chose for the APIs so all developers can follow it for all new API calls being added to the product.

It would be also great if you could write a test which proves that all 'old' API calls still exist (for now), and when time comes to discontinue them it would show that they have been removed. When changing the APIs it is usually a best practice to create an automated test first to be able to verify the transition.

felipeneuhauss:
@anvabr good question. We tried to apply the proper status code depending on the situation, being more specific. 

The 422 we added in most parts of the cases in which the message returned looked like a validation message error. In some cases, we are producing 403 when it requires authorization. 

And the 500 status code is returned only when we have exceptions that we can not identify. The 202 is specific for async functions when the request has been accepted for processing, but the processing has not been completed. It is still a success but indicates a pending result in the background.

I hope it clarified your question, and please, let me know if we can improve that even more.

anvabr:
Thank you for the explanation, it broke some of the automation tests we are working on - we'll fix them. Could you please submit a documentation PR with this explanation so the users understand the meaning of these codes

PR #2109, opened 2023-05-08
feature/demo-experience

### LINKED ISSUE #1618: Demo Experience
## Problem Description
Make the Demo Artifacts in the UI optional and remove the default demo login functionality from the header. There is existing authentication via JWT, which is not hooked up to any username/password combinations. The endpoints to create these logins already exist today.

## Requirements what needs to be done 
The demo experience menu on the login page will be removed and will only be available under specific criteria. 

This criteria for wether to use to remove these Demo Artifacts and endpoints should be taken care of before the deployment of Guardian and, in order to change this decision, Guardian should be built once more with the new configuration.

The ability to create accounts and other administrator operations should be restricted for system administrators (i.e. Standard Registries) outside of the "demo" mode.

## Definition of done
After we have chosen the best option and once the new demo page's UI is designed, built and accepted by QA.

## Acceptance criteria
The demo experience should no longer appear on the landing page. It should reproduce the same behaviour as previously being compliant with the accepted criteria of when the demo experience should or could be viewed.

PensoGlide:
**Description**:
This PR will segregate further the production from the demo experience of Guardian. This choice of experience is leveraged by the use of the variable `DEMO=true` on the root `~/.env` file and it will take action on the building phase of Guardian, meaning that a user will either intently deploy the demo or the production version. 

The orchestration works for building with Docker Compose. 

***A key thing to note is that when building with Docker Compose, if `DEMO` is present with a value other than `DEMO=true`,  the `auth-service`, `api-gateway`, `guardian-service` and `web-proxy` modules will still be built as if the demo is intended, so either use `DEMO=true` or delete the variable altogether.***

This PR, when built in **production** mode (when `DEMO` is not defined):
* does not create accounts on build besides one single StandardRegistry with the same credentials from the previous fixture file (in `./auth-service/src/app.ts` a different `fixtures.ts` file will be used depending on the presence of `DEMO=true`);
* the `api/v1/accounts/register` POST method is now guarded to be used exclusively by logged in Standard Registries;
* the `api/v1/demo/` methods are deactivated (again dependant on the presence of `DEMO=true`, as seen under `./api-gateway/src/app.module.ts`);
* the frontend will not show the "Demo Admin Panel" present on the login page until now.
Screenshot 2023-05-08 at 13 01 36
Screenshot 2023-05-08 at 13 02 32
*the frontend will not allow the user to generate Hedera credentials during the account creation process.
Screenshot 2023-05-12 at 13 42 24

This PR, when built in **demo** mode (when `DEMO=true`):
* creates all the same accounts as before on build;
* the `api/v1/accounts/register` POST method can be used by anyone even when logged out, as before;
* the `api/v1/demo/` methods are active;
* the frontend will show the "Demo Admin Panel" present on the login page until now, also with a superscript "Demo" on the side of "Guardian" and, upon login, the, until now, black header will have its color changed to gray;
Screenshot 2023-05-08 at 12 29 02
Screenshot 2023-05-08 at 12 29 20
*the frontend will allow the user to generate Hedera credentials during the account creation process.
Screenshot 2023-05-12 at 13 43 01

**Due to the guards implemented for the creation of accounts, there is now another way to achieve this (transversal to both Demo and Production modes).**
Up until now it was possible to generate new accounts on the “Create New Account” under the `/login` page, however, with these changes, this will only remain true if under a Demo build (`DEMO=true` defined on the `~/.env`). This requires a new way for Standard Registries to be able to create new accounts. The `api/v1/accounts/register` POST method is still available on both versions, exclusively for Standard Registries if on Production mode (no `DEMO` variable defined on `~/.env`), however Standard Registries also have a new menu option once logged in. Under the “Administration” menu one can find a “Create Accounts” option, that will lead to the same account creation process as before. This is depicted in the figures below.
Screenshot 2023-07-05 at 15 57 36
Screenshot 2023-07-05 at 15 57 48

It is also relevant to point out that the "demo" mode allows customisation in regards to the endpoints being used. These configurations can be found under the `./configs/.env.${GUARDIAN_ENV}.guardian.system` and the following will be the default used:
```
GATEWAY_HOST="api-gateway"
GATEWAY_PORT="3002"
GATEWAY_CLIENT_MAX_BODY_SIZE="1024m"
MRV_SENDER_HOST="mrv-sender"
MRV_SENDER_PORT="3005"
TOPIC_VIEWER_HOST="topic-viewer"
TOPIC_VIEWER_PORT="3006"
API_DOCS_HOST="api-docs"
API_DOCS_PORT="3001"
```

**Related issue(s)**:

Fixes #1618 

**Checklist**


anvabr:
@PensoGlide As I mentioned in slack there are already two builds for the FE in guardian: 

- demo build ([example for v2.12.2](https://console.cloud.google.com/gcr/images/hedera-registry/global/frontend@sha256:fe570f6922e91d70da9345af56e03da469ba1008ec0854d14d58ba3906757772/details))
- production build ([example for v2.12.2](https://console.cloud.google.com/gcr/images/hedera-registry/global/frontend-demo@sha256:dad92e4b9588ee6044f47aad9495fef5a683c4a02b11247e141990041f388a52/details))

These contain different features, the production one I believe already is restricted to not include the demo accounts. IN any case that changes that are required to achieve the objective of the requirement I believe to extend this approach into producing different builds (demo/prod) for all modules in the guardian, where production builds have the changes that your list above. This way if we every want to add/remove some demo features in different services of Guardian there is a standard and supported way of doing so. 

I don't believe what is done currently is the right way forward on this.

anvabr:
Hi @PensoGlide, as it currently stands this implementation still relies on the env variables to 'configure' the environment type at run time which does not resolve our earlier comment. As I mentioned before the best practice here would be for the two builds for Guardian to be compiled/packaged specifically for the environment in such a way that Production build does not actually have the code it should not have (not just 'disabled' execution with the 'if' statement checking an env parameter).

The easiest way to achieve this is to use a https://gulpjs.com builder, or a similar tool.

Issue #2064, opened 2023-04-26
Policy Viewer sub menu is not aligned properly on mobile version of guardian


varsha0592:
### Problem description
Policy Viewer sub menu is not aligned properly on mobile version of guardian

### Step to reproduce
1.Login as Standard Registry.
2. Click on the header and navigate to policies
3. Create/import existing policy. 
4. Go to policy viewer
5. Click on the header menu and select policies
6. Policy viewer sub menu is not displayed properly under policies. 

### Expected behavior
Policy Viewer sub menu should be aligned properly on mobile version of guardian

### Screenshots
<img width="517" alt="image" src="https://user-images.githubusercontent.com/118208225/234660350-98b12700-e9cc-499f-ab17-e8e4ae35db95.png">

PR #2333, opened 2023-06-26
hotfix/mobile-menu-alignment

PensoGlide:
**Description**:
This PR solves the issues where the "Policy Viewer" and the "Policy Editor" submenus would overlap with the rest of the menu on mobile.

Solves issue #2064 .

### DIFF
--- a/frontend/src/app/views/header/header.component.ts
+++ b/frontend/src/app/views/header/header.component.ts
@@ -37,6 +37,7 @@ export class HeaderComponent implements OnInit {
     displayDemoAccounts: boolean = environment.displayDemoAccounts;
     hederaAccountID: string | undefined;
     profileData: IUser | null = null;
+    mobileSubMenuOpen: boolean = false;
 
     public innerWidth: any;
     public innerHeight: any;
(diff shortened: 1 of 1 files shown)


frontend/src/app/modules/policy-engine/policies/policies.component.ts
```diff
@@ -681,6 +681,14 @@ export class PoliciesComponent implements OnInit {
         this.noFilterResults = false;
     }
 
+    mobileMenuFix(): void {
+        if (this.innerWidth <= 810) {
+            setTimeout(function () {
+                window.location.reload();
```
simvalery, line 687: location.reload is bad practice, need normal fix using css styles

simvalery:
Screenshot_20230627-212021_Chrome

simvalery:
Screenshot_20230627-212021_Chrome

Stepan-Kirjakov:
- Using JS for animation and hiding elements is not great especially if it can be done via css. Using css for such task is ‘best practice’ because it would be more maintainable and error free.
- Direct changes of the DOM model in angular are dangerous, this can affect the stability and correctness of angular itself. If it is completely unavoidable (which is not in this case) standard angular methods should be used for that.

I have implemented a fix for one of the ticket issues (https://github.com/hashgraph/guardian/commit/d73bf6ba3900e237eaf17ccb65e72209ca02a151), this is an example of the recommended approach - please take a look.

PR #2379, opened 2023-07-05
Update workers on docker-compose.yml

wolfwolker:
**Description**:

On cold start of docker compose, workers crash because ipfs is not ready yet. With this change the workers will wait til the ipfs sevice is READY based on the container health check.

**Related issue(s)**:

I haven't found any related open issue, but this behaviour is happening when using ipfs container.

**Steps to reproduce the issue**:

1. clone develop branch and configure the services to use containerized ipfs
2. docker compose up (after a while the workers will crash)
3. docker compose ls -a will show both workers are down
4. docker compose up will bring back the workers, and because now ipfs bootstrap is ready, they won't crash again

**After the fix**

After the fix, even on the first compose up, the workers don't crash

### DIFF
--- a/docker-compose-aws.yml
+++ b/docker-compose-aws.yml
@@ -50,8 +50,10 @@ services:
       context: .
       dockerfile: ./worker-service/Dockerfile
     depends_on:
-      - ipfs-node
-      - auth-service
+      ipfs-node:
+        condition: service_healthy
+      auth-service:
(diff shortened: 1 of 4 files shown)


simvalery:
Hi @wolfwolker , I can see that the docker-compose.yml file has been changed but there are other docker-compose*.yml files which should also be updated

Issue #2312, opened 2023-06-22
Verra VM0041 Webinar


prernaadev01:
### Problem description
Verra’s VM0041: Methodology for the Reduction of Enteric Methane Emissions from Ruminants through the Use of Feed Ingredients, v2.0 presents a set of guidelines for estimating the reduction of enteric methane (CH4) emissions resulting from the introduction of a specific feed ingredient in the diets of ruminant animals. By inhibiting methanogenesis, this methodology aims to mitigate the release of methane gas during the digestion process. The approach outlined in this methodology is designed to be applicable worldwide, allowing for consistent estimation and assessment of methane emission reductions across different regions and farming systems. By following these procedures, stakeholders can quantify and track the environmental benefits achieved through the implementation of this feed ingredient intervention. 

In this webinar we will go over the VM0041 methodology, policy creation, and have a demo of the Guardian. 

### Requirements
Design schemas for the VM0041 methodology, create a PowerPoint presentation, and conduct webinar. 

### Definition of done
Schema creation and approval
Creation of PPT for the webinar
Development of the policy
Testing the policy

### Acceptance criteria
100% working as mentioned as done definition

PR #3013, opened 2023-12-18
added vm0041 [skip ci]

artembuslaev:
#2312


artembuslaev:
**1702894551.809486643**

PR #3328, opened 2024-03-11
Move to es modules develop

ihar-tsykala:
**Description**:

This PR modifies commonjs modules to es modules
* changing in unit tests in api-tests common worker policy guardian services
* changing in gulp files in api auth worker guardian policy logger notification services
* changing in dev.js -> dev.mjs file
* changing in ecosystem config file
* changing configs micriservices for building esm
* added extention js in imports for esm
* removed module alias for corretc work esm

**Checklist**
- [X] Tested, builded app across docker-compose after changes
- [X] Tested, launched unit test scripts in local after changes
- [X] Tested, builded app across handle launching every microservice after changes
- [X] Tested, investigated ws conected and http requests in brouser dev tools


api-gateway/src/api/service/branding.ts
```diff
@@ -39,6 +39,7 @@ export class BrandingApi{
                 termsAndConditions
             };
 
+          console.log("hello");
```
Pyatakov, line 42: Needs to be cleaned up

PR #3545, opened 2024-04-15
Update documentation for IPFS provider filebase for use on gitbook

mattsmithies:
## Overview

This is a small PR that at the documentation for filebase into the docs folder, specifically for reading use in Gitbook.

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,8 +217,7 @@ The complete guide to using the new w3up web3.storage API is available at https:
 
 #### 4.3 Setting up IPFS Filebase Bucket:
 
-To configure the Filebase IPFS provider, set the following variables in the file *
-*./configs/.env.develop.guardian.system:**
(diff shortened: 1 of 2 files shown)


mattsmithies:
@anvabr Here's the updated docs, I noticed going through the code that the way I initially created this provider mechanism has changed somewhat to not be flexible for any IPFS provider.

Specifically, how the constructor signature for the _IpfsClientClass_ has now been overloaded with an additional key for filebase. Unfortunately, the implication of this is that your core class would need to be reworked to allow other implementations of clients as the class itself is dependent on multiple providers that will not be present at the same time for a given deployment of Guardian. 

In other words, the constructuor for _IpfsClientClass_ expects keys for both web3Storage and filebase for any instance of said client.

If you require some support to design this to be more generic and to potentially integrate interfaces/contracts, please let me know.
 
Screenshot 2024-04-15 at 15 13 15

PR #4021, opened 2024-08-13
fix: added evc to message report

artembuslaev:


rbarker-dev (changes requested):
It looks like something funky happened on this PR. 4838 commits and 1087 files changed for a report update?

@anvabr is this accurate?

PR #4854, opened 2025-04-25
feat: Standard Registry switcher

EnvBsh:
Implemented SR switcher


rbarker-dev (changes requested):
It looks like something funky happened on this PR. 5419 commits and 1475 files changed for a report update?

@anvabr is this accurate?

dubgeis:
@Pyatakov @swidnikk @Neurone can we confirm where this one is and the status?

Issue #4153, opened 2024-09-12
Issue when filling fields with test data


Celiant:
### Problem description
When filling the report with test data, the values of readonly fields are changed

### Step to reproduce
1. Import GCCM001 policy and open policy in Dry run
2. Open tab "Project Submission" and click on button "Add Verification Report" by Verifier
3. Click on "test" button

### Expected behavior
The value of the readonly field will be changed

### Actual behaviour
The value of the readonly field will not be changed

### Screenshots

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMi9QQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--246c98f37daf3374da597e9af8db7fda0b6d5b08/OXIfyJD.mp4)

PR #5016, opened 2025-06-16
fix: filling fields with test data [4153]

egorenvisionblockchain:
**Related issue(s)**:

Fixes https://github.com/hashgraph/guardian/issues/4153

**Checklist**


interfaces/src/helpers/generate-document.ts
```diff
@@ -94,16 +121,23 @@ export class DocumentGenerator {
...
+        option: GenerateOption,
+        rowPresets?: Record<string, any>,
     ): any {
         if (Array.isArray(field.examples) && field.examples[0]) {
             return field.examples[0];
         }
+
+        if (rowPresets?.[field.name] !== undefined) {
+            return rowPresets[field.name];
+        }
```
Pyatakov, line 139: These two needs to be swapped

Issue #5042, opened 2025-06-25
Guardian Form UI Improvements


gautamp8:
### Problem desctiption
For projects and policies of real world complexity existing Guardian UI tooling is not sufficiently powerful to make UX easy.

#### Original requirement

1) Support a table input(and csv import) field for multi-year data fields. Often methodologies require multi-year inputs(think 10 or 100 year) and it's not the easiest to visualize or input them with a typical form UI. It'd be good to have a way to have tabular inputs and a way to specify how they will be read within calculations code.

#### Requirements specification

1. Ability to manually input data in a Excel-like table view in a specific field in the form.
2. Ability to upload a `.csv` file (instead of the manual input as above), once imported the data are displayed in the Excel-like table view and are editable.
3. The .`csv` file is stored as a file (in IPFS), the 'parent' VC contains a CID reference to the file stored in IPFS.
4. Guardian Policy blocks, i.e. customLogicBlock and calculationBlock, have facilities which allow policy authors to use data from such 'csv' fields (i.e. data in these `.csv` files) in code and calculations using programming code and math.js native facilities correspondingly.

#### Other original suggestions (out of scope of this ticket - see comments below)

2) Support attachments to a field (justifications, report). Corresponding to a submitted field, a VVB or SR may ask for justification report which is usually a file. A similar issue is already there - https://github.com/hashgraph/guardian/issues/2076

_p.2 is covered by and will be implemented as part of #4548_

3) For large documents for example PDD, it'd be good to divide them into sections when user is submitting. We must allow draft saves and make dry-run testing feasible. Such a feature already exists(tabbed/navigation UI), but if document is too big, policy developer needs to click test data button multiple times to fill the entire form.

p.3: _The documents can be divided into sections - if a schema (which is driving the PDD form) contains subschemas those would be shown as separate sections by Guardian. Policy author best practice to compartmentalise fields into logical sections for enhances  end-user understanding and convenience._  

P.3: _Draft saves work in both dry-run (click save button) and in PROD policy execution mode._

4) GeoJSON/Shapefiles rendering - Allow file uploads of such types and they should be rendered with maps in best way in form UI. 

_p.4 covered by #5046_

PR #5261, opened 2025-09-07
feat: table field in form with calculation #5042

ihar-tsykala:
**Description**:

feat: table field in form with calculation #5042

**Checklist**


indexer-frontend/package.json
rbarker-dev, line 1: Please lock all versions in the package.json file (meaning no `^x.y.z` and no `~x.y.z` versions) for all dependencies, peer dependencies, and dev dependencies.

frontend/package.json
rbarker-dev, line 1: Please lock all versions in the package.json file (meaning no ^x.y.z and no ~x.y.z versions) for all dependencies, peer dependencies, and dev dependencies.

rbarker-dev (changes requested):

Issue #4988, opened 2025-06-04
[BOUNTY PROPOSAL] Capability to retire tokens by serial number


palindaa:
### Problem description
The current implementation of Hedera Guardian does not support retiring NFTs by a specific set of serial numbers, even though the Hedera SDK natively supports this operation. In applications such as carbon credit management, it is often necessary to retire selected tokens (e.g., specific serial numbers) rather than the entire supply or a fixed count.

### Requirements
Enhance Guardian to support NFT retirement by specifying a list of serial numbers. The implementation should:
- Allow users or policy workflows to pass a custom array of serial numbers.
- Call the appropriate Hedera SDK function to retire those specific serial-numbered NFTs.

### Definition of done
- Guardian updates the Wipe block to enable retiring selected NFTs by a list of serial numbers.
- When the policy executes the Wipe block, the specified serial numbers provided in the token block must be retired.

### Acceptance criteria
- Users can specify one or more serial numbers to retire.
- The system successfully retires only the specified NFTs
- Relevant documentation and example usage are added.

PR #5333, opened 2025-09-30
feat: Capability to retire tokens by serial number #4988

gayanath8:
### Issue Label
[https://github.com/hashgraph/guardian/issues/4988](https://github.com/hashgraph/guardian/issues/4988)

### Problem Description
The current implementation of Hedera Guardian does not support retiring NFTs by a specific set of serial numbers, even though the Hedera SDK natively supports this operation. 

In applications such as carbon credit management, it is often necessary to retire selected tokens (e.g., specific serial numbers) rather than the entire supply or a fixed count.

### Requirements
Enhance Guardian to support NFT retirement by specifying a list of serial numbers. The implementation should:
* Allow users or policy workflows to pass a custom array of serial numbers.
* Call the appropriate Hedera SDK function to retire those specific serial-numbered NFTs.

### Summary of Changes
By applying this enhancement:

* Users can select specific NFT serial numbers to retire.
* For NFT operations:

  * **Serial Numbers** field is required.
  * Guardian will generate an array of serial numbers and pass them to the Hedera SDK for the wipe operation.
* For FT operations:

  * **Rule** remains the required field.
  * No changes are introduced to the FT wipe process.

### Validation Rules
* The Serial Numbers field is required.
* The Serial Numbers expression may contain individual numbers or inclusive ranges (e.g., 1,2-5,10).
* If a range is used (e.g., 2-5), the end value must be greater than or equal to the start value.
* Rule is required for FT operations.

### NFT Wipe Flow
**_Schema-Driven Wipe Flow_**

Serial numbers can be dynamically provided via schema fields or policy formulas. This enables multiple wipe operations for the same token, with different sets of serial numbers, as needed. Can useful for workflows requiring repeated or conditional NFT retirement.

FireShot Capture 004 - Guardian -  3 93 78 104
image

* Example usage
  * Mint NFTs via the Mint Block, with the amount provided by the schema.
  * Add a separate schema for wipe operations.

image

  * Retire serial numbers 1, 3, 4, 5, 8, 12, 13, 14, 15, (entered dynamically in schema).

image
image

**_Single Wipe Flow (Per User, Per Token)_**

A constant value for Serial Numbers Expression (for example, 1,2-5,10) can be provided. This allows a one-time wipe of specific NFTs per user and per token throughout the policy lifecycle.

image

* Example usage
  * Mint 20 NFTs using the Mint Block.
  * (Minting amount must be ≥ highest serial number to be retired)
  * Retire serial numbers 10,12,13,14,18. 
  
image
image

Transaction History of Single NFT
image

### FT Wipe Flow

No changes to the existing FT token wipe flow.

image

* Example usage
  * Create an FT token using schema-driven Mint Block.
  * Use Wipe Block to retire 1 token (based on Rule).
  
  
image
image
image

### DIFF
--- a/common/src/helpers/workers.ts
+++ b/common/src/helpers/workers.ts
@@ -44,6 +44,7 @@ export const NON_RETRYABLE_HEDERA_ERRORS = [
     HederaResponseCode.CANNOT_WIPE_TOKEN_TREASURY_ACCOUNT,
     HederaResponseCode.TOKEN_HAS_NO_WIPE_KEY,
     HederaResponseCode.INVALID_WIPE_KEY,
+    HederaResponseCode.INVALID_NFT_ID,
 
     // Burn type errors
     HederaResponseCode.INVALID_TOKEN_BURN_AMOUNT,
--- a/docker-compose-build.yml
+++ b/docker-compose-build.yml
@@ -18,6 +18,8 @@ services:
     restart: always
     expose:
       - 27017
+    volumes:
+      - mongo_data:/data/db
 
   mongo-express:
     image: mongo-express:1.0.2-20
@@ -36,6 +38,8 @@ services:
     restart: always
     expose:
       - 6379
+    volumes:
+      - redis_data:/data
 
 #  seq:
 #    image: datalust/seq
@@ -328,6 +332,8 @@ volumes:
   grafana_data:
   ipfs_data:
   ipfs_export:
+  mongo_data:
+  redis_data:
 #  seq_data:
 
 networks:
--- a/docker-compose-build.yml
+++ b/docker-compose-build.yml
@@ -18,8 +18,6 @@ services:
     restart: always
     expose:
       - 27017
-    volumes:
-      - mongo_data:/data/db
 
   mongo-express:
     image: mongo-express:1.0.2-20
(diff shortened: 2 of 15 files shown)


Neurone:
Hi, @gayanath8 thanks for the submission, really appreciated.

As mentioned in our last live meeting, it is important for the users to have more flexibility in the definition of the tokens to be wiped.

An idea is having a field where users can enter a range of serial numbers, such as "1,2-5,10" or "field0,field1-field5,field10," rather than just the starting and ending serial numbers.

What do you think? Can you update the PR accordingly?

The splitBlock probably already contains the code to handle both field values and fixed numbers (we can already do things like "field0/1000" there) so that can be a good starting point if you need info about how those things are managed.

Neurone:
As a side note, it seems the PR contains some changes unrelated to the specific features (for example the changes to the docker-compose-build.yml file). Can you please remove this kind of changes and focus only on the relevant ones?

It would be still useful gathering all the side improvements you think we should apply to other part of the source code, just use a separate PR for that.

anvabr (changes requested):
Can you please ensure that there are no errors displayed on the PR:
- need to fix DCO
- PR title to conform to 'conventional commits' spec
Screenshot 2025-10-09 at 16 06 03

Issue #4548, opened 2025-01-16
Complex iterative review and approval workflows


anvabr:
### Problem description
Real world document review and approval processes are often nuanced and complex, with multiple iterations of exchanging comments, questions, change requests, other forms of feedback, and then corrections and resubmissions. Furthermore these processes can be of varying length - i.e. they can last from hours to several months or even years. At present Guardian workflow does not 'natively' support this - policy authors have to implement such cycles in every policy and at policy run time project participants would have to rely on limited communication capabilities  ('reason for approval').

See also: #3681, #2845, #3617

### Requirements
Introduce `reviewVcDocumentBlock` as a complement to the `requestVcDocumentBlock` into Policy definition language and corresponding Guardian UI tools which would facilitate complex iterative document review workflows supporting rich communications, namely the ability to:

- for forms, corresponding to each field, ability to add optional clarifications (text, links) or upload documents where submitters can share additional information on values entered
- send messages/descriptions associated with actions
- conduct stand-alone (not action-linked) message exchanges (questions, responses, general comms)
- send requests for providing additional documents/evidence, and provide facilities to upload/receive them
- ability to edit/correct previously submitted (but rejected) document and resubmit again
- forward messages to another user
- involve other users into 'conversations'

These actions, steps, message exchanges have to be verifiably traceable, i.e. recorded and published as VC/VP documents alongside the usual data-containing documents produce by the policy workflows.

#### Roles, permissions and data visibility
- Additional workflow data, i.e. messages (comments) and documents, can only exist in the context of specific documents (VCs/VPs), i.e. each is 'linked to' a document. They can't exist 'stand alone'.
- These comments/documents can have 3 types of visibility which configured at the creation time by their creator user:
  - public - visible to all users who have at least read access to the document
  - group - addressed to a specific role (i.e. group of users who have been assigned a particular role)
  - individual users - only named users have access to the data
- Overriding the above, there can be 'special' users which are assigned 'view all data' role which have access to all comments/documents regardless of the visibility configuration
- Comments and documents are part of the 'trust chain' and as such they are published to Hedera/IPFS under the following rules:
  - public comments are published as VCs (with the corresponding Hedera anchoring messages) signed by the user creating them
  - private comments (i.e. those assigned to individual users and/or groups): TBD

### Definition of done

- Functionality is implemented per requirements above
- Documentation is provided
- At least a single example of the usage of this functionality is implemented in any of the policies in the Guardian repository 

### Acceptance criteria
[...]

PR #5360, opened 2025-10-14
feat: Complex iterative review and approval workflows #4548

Stepan-Kirjakov:
#4548 Complex iterative review and approval workflows


topic-listener-service/package.json
rbarker-dev: We should remove all `^` and `~` versions. Ideally we have strict versioning for all dependencies, devdependencies, peerdependencies, etc.

rbarker-dev (changes requested):
We have a request from the hashgraph team to pin all dependency versions in the package.json files throughout all repositories using them. Can you please pin the dependency, dev-dependency, peer-dependency versions, etc.?

Pyatakov:
Superseded by #5400

Issue #3428, opened 2024-04-02
Accessible (non-color based) policy diff


anvabr:
### Problem description
While Guardian policy editor was implemented to support color blindness - by introducing additional icons to depict the 'highlighted' blocks - such aids are absent from the 'diff' functionality.

### Requirements
Introduce non-color based visual elements to the 'diff' UI such that users with impaired color vision can still adequately perceive the 'diff' functionality.

### Definition of done

- The functionality is implemented as per above
- Documentation is updated to highlight the elements/tools to be used instead of colors.

### Acceptance criteria
It is possible to identify similar/different/missing parts of the files in the diff UI without relying on color-coding.

mykytayefrueiv:
Hello, I have started working on this Epic.

PR #5574, opened 2025-12-16
feat: Added accessibile colors and icons for policy diff

mykytayefrueiv:
<p>Important note:</p>
<p>I am color blind :) I have deuteranomaly, so this feature is really close to me and I will do my best to make it as good as I can.</p>
<hr>
<p>This feature consists of two tracks. (#3428)</p>
<p>One is helping them perceive diffs without relying on affected colors, second one is to provide non-color based diff presentation.</p>
<h1>Color palette track</h1>
<h2>Color-blind types</h2>
<p>There are different types of color blindness conditions:</p>
<ul>
<li><strong>Protanomaly</strong>, which is a reduced sensitivity to red light</li>
<li><strong>Deuteranomaly</strong> which is a reduced sensitivity to green light (the most common form of color blindness)</li>
<li><strong>Tritanomaly</strong> which is a reduced sensitivity to blue light (extremely rare)</li>
</ul>
<h3>Picking colors</h3>
<p>To address this we will pick color theme which is accessible for people who are colorblind.</p>
<p>Multiple approaches exist to address this problem, but the core principle is to avoid placing colors that may appear identical to colorblind users adjacent to each other.</p>
<p>Our feature is intended to convey different meanings through colors, here are 4 of them:</p>
<ul>
<li>
<p>Blocks are equal, including their child blocks.</p>
<p>To address this blue-ish (#648fff) color was picked. Since it conveys “positive” meaning and we don’t want to use green color.</p>
</li>
<li>
<p>Blocks are equal, but their child blocks are different.</p>
<p>Combination of blue and orange (#ffb000) colors picked for this option, due to the high contrast of this pair it is easy to distinguish between others. Meaning, block are “sort of” equal, so blue and orange.</p>
</li>
<li>
<p>Blocks are of the same type and are partially equal.</p>
<p>Orange (#ffb000) was picked for this option for same reason as previously.</p>
</li>
<li>
<p>Blocks are absent in the other Policy.</p>
<p>And the final one is darker orange (#d55e00), again we don’t lean into red colors, but we want to convey “negative” meaning.</p>
</li>

image

</ul>
<p>These color combinations also meet WCAG 2.1 AA contrast requirements (<a href="https://www.w3.org/TR/WCAG21/#contrast-minimum">https://www.w3.org/TR/WCAG21/#contrast-minimum</a>)</p>
<h1>Icons track</h1>
<p>Apart from proper coloring it is required to convey the same meaning without relying on colors only. For that we can utilize icons.</p>
<p>We can seek inspiration for “diff” functionality on the code diff platforms, such as GitHub, GitLab, BitBucket, etc.</p>
<p>In general, we can distill all the non-color meaning of the diff to:</p>
<ul>
<li>“-” symbol near the deleted line</li>
<li>“+” symbol near the added line</li>
</ul>
<p>Now two things to take into account, that are crucial comparing guarding policy to commit diffs:</p>
<ul>
<li>Policy structure is different from commits. Policy is a tree, commit is a flat structure. From this property arises main challenging part. (We will dive into this soon)</li>
<li>Comparing commits one usually compares “old” (base) commit/branch and “new” (compare) commit/branch, which is not necessarily true about comparing policies. There are no “base” policy and “compare” policy.</li>
</ul>
<h3>Icons for trees</h3>
<ul>
<li>
<p>Blocks are equal, including their child blocks.</p>
<p>For this block we won’t be using any icon, same as in GitHub/Gitlab, so we won’t overwhelm user with variety of icons.</p>
</li>
<li>
<p>Blocks are absent in the other Policy.</p>
<p>For this block we can again utilize GitHub/Gitlab approach and use +/- symbols to convey the respective meaning.</p>
</li>
</ul>
<p>But the most challenging part is to determine and find icons that will represent similar, but yet different concepts:</p>
<ul>
<li>Blocks are equal, but children are different</li>
<li>Blocks of the same type and are partially equal</li>
</ul>
<p>Both of them are about equality, but on different levels.</p>
<p>To handle this I came up with such notation:</p>
<ul>
<li>To convey meaning that “something inside is different” I used <code>[≠]</code> , not equal sign inside of the brackets. Which exactly delivers necessary meaning. (Brackets are “container”, so “inequality inside of a container”)</li>
<li>For block partial equality, there already exists respective mathematical symbol – <code>≈</code></li>
</ul>
<p>To conclude, the solution consists of two main approaches:</p>
<ul>
<li>Accessible Color Selection: Using a carefully chosen color palette that avoids problematic red/green/blue combinations. The design uses blue (#648fff) and orange (#ffb000) tones to represent four different block states: equal blocks, equal blocks with different children, partially equal blocks, and absent blocks⁠⁠.</li>
<li>Icon-Based Indicators: Supplementing colors with symbols to convey meaning without relying solely on color. Drawing inspiration from code diff platforms, the design uses +/- symbols for additions/deletions, [≠] to indicate different children, and ≈ for partial equality</li>
</ul>

State | Color | Hex | Icon
-- | -- | -- | --
Equal (including children) | Blue | #648fff | (none)
Equal (different children) | Blue + Orange | #648fff / #ffb000 | [≠]
Partially equal | Orange | #ffb000 | ≈
Absent in other policy | Dark orange | #d55e00 | +/-

<p>Before:</p>
image

<p>After:</p>
<p>
image
</p>

Helpful links:
- https://atmos.style/blog/color-blindness-in-ui-design
- https://davidmathlogic.com/colorblind/#%23D81B60-%231E88E5-%23FFC107-%23004D40
- https://www.nceas.ucsb.edu/sites/default/files/2022-06/Colorblind%20Safe%20Color%20Schemes.pdf
- https://www.color-hex.com/color-palette/1044488
- https://www.w3.org/TR/WCAG20/
- https://marijohannessen.github.io/color-contrast-checker/

### DIFF
--- a/frontend/src/app/modules/analytics/compare-policy/compare-policy.component.scss
+++ b/frontend/src/app/modules/analytics/compare-policy/compare-policy.component.scss
@@ -257,7 +257,8 @@
 
 .report-item[type="LEFT"] .block-tree,
 .report-item[type="RIGHT"] .block-tree {
-    background: var(--left-right-color);
+	background: var(--absent-color);
+	color: lightblue;
 }
 
 .report-item[type="PARTLY"] .block-tree {
@@ -285,10 +286,9 @@
 }
 
 :host.colorblind-mode {
-    --full-color: #648fff;
-    --partly-color: #ffb000;
-    --left-right-color: #fe6100;
-    --absent-color: #d55e00;
+	--full-color: #b8ccff;
+	--partly-color: #d49e25;
+	--absent-color: #4a2c6a;
     --hover-color: rgb(229 229 229 / 50%);
 
 	 --full-color-bg: rgba(100, 143, 255, 0.15);
@@ -490,7 +490,7 @@
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
-	 font-weight: 600;
+	 font-weight: 500;
 }
 
 .block-icon {


anvabr:
@mykytayefrueiv thank you for the excellent PR! Please could you address the p1 (failed) test as reported above and we'd happily merge it.

PR #6011, opened 2026-04-30
chore: Increase initial standard registry balance

Pyatakov:
Bump INITIAL_STANDARD_REGISTRY_BALANCE from 13 to 20 in .github/workflows/api-after-commit.yml. This raises the default standard registry funds used by the CI workflow to avoid tests failing due to insufficient balance.


Neurone (approved):
LGTM

Issue #5795, opened 2026-03-06
[PSE - API] Simplify User Onboarding by Auto-Generating Hedera Account, DID, and Keys via API


palindaa:
### Labels
`enhancement`, `api`, `onboarding`, `high-priority`

### **Problem description**

Guardian user onboarding currently requires multiple separate steps: creating a Hedera account (and wallet), generating a DID, generating keys, and then completing the standard registration flow. This multi-step process is overly complex for both direct Guardian users and for external systems integrating with Guardian via API.

When integrating Guardian with another system, users must be mapped 1-to-1 with Guardian users along with their permissions. This becomes a significant hurdle because the onboarding steps are fragmented across multiple API calls and manual interventions. This feature can improve the quality of life for companies such as Allcot and Verra. 

### **Requirements**

*   Add new optional parameters to the existing user registration/onboarding API endpoint that allow the system to auto-generate a Hedera Account, DID, and cryptographic keys on behalf of the user.
    
*   If the caller does not provide a Hedera account, DID, or keys, Guardian should generate them automatically during onboarding (similar to the current internal Guardian flow, but exposed as a single API call).
    
*   If the caller provides their own Hedera account, DID, or keys, Guardian should accept and use those values instead.
    
*   The auto-generated values (Hedera Account ID, DID, keys) must be included in the API response payload.
    
*   All generated values must be securely stored in the system vault.
    
*   The existing API fields must remain backward-compatible to support user-created Hedera accounts.
    
*   A single API call should be sufficient to fully onboard a user, replacing the current multi-step process.
    

### **Definition of done**

*   A single API call to the onboarding endpoint can create a fully onboarded Guardian user with a Hedera account, DID, and keys — without requiring any additional API calls or manual steps.
    
*   The endpoint accepts optional parameters for Hedera account, DID, and keys; when omitted, they are auto-generated.
    
*   Response payload contains all generated credentials.
    
*   Generated secrets are persisted in the system vault.
    
*   Existing integrations that provide their own Hedera account continue to work without modification.
    

### **Acceptance criteria**

*   \[ \] Calling the onboarding endpoint without Hedera account/DID/key parameters returns a response containing the auto-generated Hedera Account ID, DID, and public key.
    
*   \[ \] Calling the onboarding endpoint with user-supplied Hedera account/DID/key parameters uses the provided values and does not auto-generate.
    
*   \[ \] Auto-generated keys are stored in the system vault and can be retrieved for subsequent operations.
    
[...]

PR #6035, opened 2026-05-07
feat: Simplify User Onboarding by Auto-Generating Hedera Account, DID, and Keys via API #5795

gayanath8:
### Issue Labels
[https://github.com/hashgraph/guardian/issues/5795](https://github.com/hashgraph/guardian/issues/5795)

### Requirements

- Add new optional parameters to the existing user registration/onboarding API endpoint that allow the system to auto-generate a Hedera Account, DID, and cryptographic keys on behalf of the user.
- If the caller does not provide a Hedera account, DID, or keys, Guardian should generate them automatically during onboarding (similar to the current internal Guardian flow, but exposed as a single API call).
- If the caller provides their own Hedera account, DID, or keys, Guardian should accept and use those values instead.
- The auto-generated values (Hedera Account ID, DID, keys) must be included in the API response payload.
- All generated values must be securely stored in the system vault.
- The existing API fields must remain backward-compatible to support user-created Hedera accounts.
- A single API call should be sufficient to fully onboard a user, replacing the current multi-step process.
- Hedera Account Reuse Restriction – When specifying an existing Hedera account in the API (Simplified Task), ensure that the account has not already been used for a prior user registration. The system will return an error if reused, as each user requires a unique DID key derived from the Hedera account’s private key.

### Summary of Changes

- **`POST /accounts/push/onboard`** — new async endpoint that accepts optional Hedera account, DID, and key fields; auto-generates any omitted values using the operator account. Returns a `taskId` immediately for polling.
- **`GET /tasks/onboard/{taskId}`** — new unauthenticated endpoint that returns onboarding task progress (sanitized — no credentials exposed). Only serves onboarding tasks; returns `401` for any other task type.
- `OnboardingDTO` — new DTO extending `RegisterUserDTO` with optional fields: `hederaAccountId`, `hederaAccountKey`, `parent`, `vcDocument`, `didDocument`, `didKeys`, `useFireblocksSigning`, `fireblocksConfig`. Includes format validation for Hedera account ID and key.
- `TaskAction.ONBOARD_USER` / `MessageAPI.ONBOARD_USER_ASYNC` — new enum values in `interfaces` for the async task pipeline.
- `guardian-service` — new `ONBOARD_USER_ASYNC` handler in `profile.service.ts` that runs the full onboarding flow: creates Hedera account (if needed), registers user, publishes DID/VC, stores keys in vault, and returns credentials in the task result.
- Task ownership transfer is assigned ,on completion, task ownership is automatically transferred to the newly created user, so only the new user can retrieve their sensitive credentials via the authenticated `GET /tasks/{taskId}` endpoint.
- New Docs added `docs/guardian/users/user-operations/account-apis/unified-onboarding.md` and updated `docs/SUMMARY.md`.

### Validation Rules

- `username` — required, non-empty string
- `password` — required, non-empty string
- `password_confirmation` — required, must match `password`
- `role` — required, must be one of `STANDARD_REGISTRY`, `USER`
- `hederaAccountId` — optional string, unique
- `hederaAccountKey` — optional string
- `parent` — optional string
- `didDocument`, `vcDocument`, `didKeys` — optional, typed objects/arrays

| Condition | Status |
|-----------|--------|
| Missing or invalid Bearer token (non-demo mode) | `401 Unauthorized` |
| Caller is not `STANDARD_REGISTRY` (non-demo mode) | `403 Forbidden` |
| `hederaAccountKey` provided but `hederaAccountId` format invalid (`\d+.\d+.\d+`) | `422 Unprocessable Entity` |
| `hederaAccountId` provided but `hederaAccountKey` format invalid (DER / raw hex) | `422 Unprocessable Entity` |
| `role = USER` with no `parent` provided | `400 Bad Request` |
| `parent` username not found in system | `400 Bad Request` |
| `parent` exists but is not a `STANDARD_REGISTRY` | `400 Bad Request` |
| Username already taken | `409 Conflict` |
| Hedera account '0.0.8682772' is already associated with an existing registration | `409 Conflict` |

### **1. SR — auto Hedera, auto DID, with VC**

- image
- image
- image

### **2. SR — manual Hedera, auto DID, with VC**

- image
- image
- image

### **3. Default User — auto Hedera, with parent registry**

- image
- image
- image

### **Validations**

- image
- image
- image
- image
- image
- image


api-gateway/src/api/service/account.ts
```diff
@@ -191,6 +191,160 @@ export class AccountApi {
...
+                HttpStatus.UNPROCESSABLE_ENTITY,
+            );
+        }
+
+        if (body.hederaAccountId && !HEDERA_ACCOUNT_KEY_REGEX.test(body.hederaAccountKey)) {
+            throw new HttpException(
+                'Invalid hederaAccountKey format. Expected a valid Hedera private key',
+                HttpStatus.UNPROCESSABLE_ENTITY,
+            );
+        }
```
Pyatakov: #### Hedera ID/Key format checks are cross-coupled

Each regex is gated on the other field's presence:
```
if (body.hederaAccountKey && !HEDERA_ACCOUNT_ID_REGEX.test(body.hederaAccountId)) { … }
if (body.hederaAccountId && !HEDERA_ACCOUNT_KEY_REGEX.test(body.hederaAccountKey)) { … }
```
A field's format is never actually tested against its own regex when only that field is supplied. The request still gets rejected (the missing field's regex fails on `undefined`), but the error message points at the wrong field. For example, supplying a malformed `hederaAccountId` with no key returns "Invalid hederaAccountKey format", which gives the caller no signal that their ID is the problem.

Please restructure so each field validates its own format independently, and express the "both-or-neither" rule as its own explicit gate, e.g.:
```
if (body.hederaAccountId || body.hederaAccountKey) {
    if (!body.hederaAccountId || !HEDERA_ACCOUNT_ID_REGEX.test(body.hederaAccountId)) { throw ... }
    if (!body.hederaAccountKey || !HEDERA_ACCOUNT_KEY_REGEX.test(body.hederaAccountKey)) { throw ... }
}
```

api-gateway/src/api/service/account.ts
```diff
@@ -191,6 +191,160 @@ export class AccountApi {
...
+                parentUser = null;
+            }
+            if (!parentUser) {
+                throw new HttpException('UNAUTHORIZED', HttpStatus.UNAUTHORIZED);
+            }
+            try {
+                await checkPermission(UserRole.STANDARD_REGISTRY)(parentUser);
+            } catch (error) {
+                await InternalException(error, this.logger, parentUser?.id);
+            }
```
Pyatakov: #### checkPermission 403 is rethrown as 500
`checkPermission` already throws `HttpException(FORBIDDEN)`. Wrapping it in `InternalException` rethrows as `500 Internal Server Error`, which contradicts the `403 Forbidden` row in this PR's documented validation table, which means non-SR callers will see 500 instead of 403.

Remove the try/catch and let the `HttpException` propagate.

api-gateway/src/api/service/account.ts
```diff
@@ -191,6 +191,160 @@ export class AccountApi {
...
+            );
+        }
+
+        if (body.hederaAccountId && !HEDERA_ACCOUNT_KEY_REGEX.test(body.hederaAccountKey)) {
+            throw new HttpException(
+                'Invalid hederaAccountKey format. Expected a valid Hedera private key',
+                HttpStatus.UNPROCESSABLE_ENTITY,
+            );
+        }
+
```
Pyatakov, line 276: #### `fireblocksConfig` requirement is documented but not enforced
The PR description's validation table and the `fireblocksConfig` DTO description both state it is required when `useFireblocksSigning` is `true`, but no code path enforces it. A request with `useFireblocksSigning: true` and a missing `fireblocksConfig` is accepted by the API and only fails later inside profile setup with an opaque error.

Add an explicit guard here alongside the existing 422 checks:
```
if (body.useFireblocksSigning === true && !body.fireblocksConfig) {
    throw new HttpException(
        'fireblocksConfig is required when useFireblocksSigning is true',
        HttpStatus.UNPROCESSABLE_ENTITY,
    );
}
```

api-gateway/src/middlewares/validation/schemas/accounts.ts
```diff
@@ -411,6 +412,56 @@ export class UserAccountDTO {
...
+    didKeys?: DidKeyDTO[];
+
+    @ApiProperty({ required: false, default: false, description: 'Use Fireblocks signing instead of local key.' })
+    @IsOptional()
+    @IsBoolean()
+    useFireblocksSigning?: boolean;
+
+    @ApiProperty({ required: false, description: 'Fireblocks configuration (required when useFireblocksSigning is true).' })
+    @IsOptional()
+    fireblocksConfig?: any;
```
Pyatakov: `fireblocksConfig?: any;` skips any structural validation of the nested object and renders nothing useful in Swagger. `IFireblocksConfig` already defines the shape in `guardian-service/src/api/helpers/profile-helper.ts`. Please introduce a `FireblocksConfigDTO` mirroring that interface and reference it via `type: () => FireblocksConfigDTO` in `@ApiProperty`. `@ValidateNested()` has one existing usage (policies.dto.ts:417) and can be applied here too.

guardian-service/src/api/profile.service.ts
```diff
@@ -168,6 +169,113 @@ export function profileAPI(logger: PinoLogger) {
...
+                const profile = {
+                    ...payload,
+                    hederaAccountId,
+                    hederaAccountKey,
+                    type: LocationType.LOCAL,
+                    entity: payload.role === UserRole.STANDARD_REGISTRY
+                        ? SchemaEntity.STANDARD_REGISTRY
+                        : SchemaEntity.USER,
+                    parent: payload.parent ?? parentUser?.did ?? null,
+                } as unknown as ICredentials;
```
Pyatakov: `{ ...payload, … } as unknown as ICredentials` turns off type checking at exactly the spot where the incoming request body becomes the credentials object that gets written to the vault and sent to workers. Because we spread the whole payload, fields like `username`, `password`, `password_confirmation`, `role`, and a string `parent` end up on an object typed as `ICredentials`, which doesn't have those fields. That keeps the plaintext password around longer than it needs to be, and means future changes to `ICredentials` won't be caught here at compile time.

Please build the `ICredentials` object explicitly, listing each field by name (no spread, no cast). That way, if `ICredentials` changes in the future, TypeScript will flag this.

Pyatakov (changes requested):
@gayanath8 please address the issues mentioned

api-gateway/src/api/service/account.ts
gayanath8: Fixed. Restructured into a single if (body.hederaAccountId || body.hederaAccountKey) block

image
image

api-gateway/src/api/service/account.ts
gayanath8: Removed the try/catch around checkPermission in both this and register endpoint as well

api-gateway/src/middlewares/validation/schemas/accounts.ts
gayanath8: Imported the existing FireblocksConfigDTO and update the OnboardingDTO

api-gateway/src/api/service/account.ts
gayanath8: Added an explicit 422 guard: if useFireblocksSigning === true && !body.fireblocksConfig

image

guardian-service/src/api/profile.service.ts
gayanath8: Fixed. Built ICredentials explicitly field by field — no spread, no wholesale cast.

Pyatakov (approved):

Issue #5800, opened 2026-03-06
[PSE - API] Add API Endpoints for Token Transfers


palindaa:
**Labels:** `enhancement`, `api`, `tokens`, `medium-priority`

### **Problem description**

Token transferring  are currently out of scope of Guardian policy workflows. Even if an external system integrates with Guardian up to the minting process (through policies), transferring or retiring tokens afterwards becomes too complex or too technical because Guardian manages the user's cryptographic keys internally. There is no API for authorized users or token owners to perform token transfers  when Guardian is the custodian of their wallets.

This limitation has been raised by NoviqTech and Xeptagon as a barrier to end-to-end token lifecycle management through Guardian. This feature can improve the quality of life for companies such as NoviqTech.

### **Requirements**

*   Add new API endpoints to Guardian that allow authorized users (token owners) to perform token transfers externally.
    
*   The endpoints must enforce proper authorization — only the owner of the tokens/wallets should be able to initiate transfers or retirements.
    
*   Since Guardian holds and manages user wallets and keys (especially with the CR-1 onboarding changes), the API must securely sign transactions on behalf of the user using their vault-stored keys.
    
*   Support standard Hedera token operations: transfer to another account.
    

### **Definition of done**

*   API endpoints exist for token transfer operations.
    
*   Authorized users can transfer tokens to other Hedera accounts tokens via API.
    
*   Transactions are signed using the keys managed by Guardian's vault on behalf of the token owner.
    
*   Proper authorization checks prevent unauthorized token operations.
    

### **Acceptance criteria**

*   \[ \] A POST endpoint exists for initiating a token transfer, accepting target account, token ID, and amount.
    
*   \[ \] Only the authenticated owner of the tokens/wallet can invoke these endpoints.
    
*   \[ \] Guardian signs the Hedera transaction using the owner's keys stored in the vault.
    
*   \[ \] Successful transfers are reflected on the Hedera network and can be verified via a Hedera explorer.
    
*   \[ \] Error responses are returned for insufficient balance, unauthorized access, and invalid parameters.
    
*   \[ \] API documentation covers both endpoints with request/response examples.

PR #6036, opened 2026-05-07
feat: Add API Endpoints for Token Transfers #5800

OshanM:
### **Issue Labels**:
[#5800](https://github.com/hashgraph/guardian/issues/5800)

### **Requirements**:

- Add new API endpoints to Guardian that allow authorized users (token owners) to perform token transfers externally.
- The endpoints must enforce proper authorization — only the owner of the tokens/wallets should be able to initiate transfers or retirements.
- Since Guardian holds and manages user wallets and keys, the API must securely sign transactions on behalf of the user using their vault-stored keys.
- Support standard Hedera token operations: transfer to another account.

### **API Changes**:
* Transfer APIs created for:
   * Synchronous execution (/tokens/{tokenId}/transfer)
   * Asynchronous execution (/tokens/push/{tokenId}/transfer)
* Token types supported:
    * FT - Amount has to be provided with target account ID
    * NFT - Specific serial numbers or amount can be given with target account ID. If amount is given, then the serials at the end will be transferred

### **Example POST requests (async)**:

**Fungible Token transfer:**
```bash
POST http://<ip>:<port>/tokens/push/{tokenId}/transfer                                                                          
Authorization: Bearer <your_jwt_token>
Content-Type: application/json                                                                                                                                                                                        {                                                                                                        
    "targetAccount": "0.0.12345",
    "amount": 10,
    "memo": "test transfer"
}
```

**NFT transfer by serial numbers:**
```bash
POST http://<ip>:<port>/tokens/push/{tokenId}/transfer                                                                          
Authorization: Bearer <your_jwt_token>
Content-Type: application/json                                                                                                                                                                                        {
    "targetAccount": "0.0.12345",
    "serialNumbers": [1, 2, 3],
    "memo": "test nft transfer"
}
```

**NFT transfer by amount (picks highest-numbered serials you own):**
```bash
POST http://<ip>:<port>/tokens/push/{tokenId}/transfer                                                                          
Authorization: Bearer <your_jwt_token>
Content-Type: application/json                                                                                                                                                                                        {
    "targetAccount": "0.0.12345",
    "amount": 2,
    "memo": "test nft transfer by amount"
}
```

#### **Response format (async)**:
The response will be a TaskDTO with a taskId you can poll to track progress:
```json
{
    "taskId": "...",
    "expectation": 4
}
```

#### **Async transfer examples**:

**Transfer NFTs given the amount**:

Request:
image

Response:
image

Hashscan:
image
image

**Transfer NFTs given serial numbers**:

Request:
image

Response:
image

Hashscan:
image
image

### **Example POST requests (Sync)**

**Fungible Token (FT) — transfer by amount:**

```bash
POST http://<ip>:<port>/tokens/0.0.8657911/transfer
Authorization: Bearer <your-jwt-token>
Content-Type: application/json                                                                                                                                                                                          {
    "targetAccount": "0.0.99999",                                                                          
    "amount": 10
}
```

**NFT - transfer by explicit serial numbers:**

```bash
POST http://<ip>:<port>/tokens/0.0.8657911/transfer
Authorization: Bearer <your-jwt-token>
Content-Type: application/json                                                                                                                                                                                          {
    "targetAccount": "0.0.99999",
    "serialNumbers": [1, 2, 3]
}
```

**NFT - transfer by amount (auto-resolves the last N serials you own):**

```bash
POST http://<ip>:<port>/tokens/0.0.8657911/transfer
Authorization: Bearer <your-jwt-token>
Content-Type: application/json                                                                                                                                                                                          {
    "targetAccount": "0.0.99999",
    "amount": 2
}
```

**Response format (FT transfer success):**

```json
{                                                                                                        
    "status": true
}
```

**Response format (NFT transfer success):**

```json
{                                                                                                        
    "status": true,
    "serials": [1, 2, 3]
}
```

**Sync transfer examples:**

**NFT serial transfer:**

Request/Response:
image

Hashscan:
image

**NFT transfer (given amount):**

Request/Response:
image

Hashscan:
image

__**Additionally, FT transactions are processed the same way.**__


guardian-service/src/api/token.service.ts
```diff
@@ -1401,4 +1401,418 @@ export async function tokenAPI(dataBaseServer: DatabaseServer, logger: PinoLogge
...
+     * @param targetAccount
+     * @param amount
+     * @param serialNumbers
+     * @param memo
+     * @param owner
+     * @param notifier
+     */
+    const NFT_TRANSFER_MAX_BATCH_SIZE = 10;
+
+    async function transferToken(
```
Pyatakov: `transferToken` and `transferTokenAsync` are ~80% identical. Only meaningful difference is the sync version enforces `NFT_TRANSFER_MAX_BATCH_SIZE = 10` while async chunks. Collapse into a single helper taking a `chunk: boolean` (or always chunk and use the limit only as an early-rejection check in the sync path). The duplication is ~200 lines and will compound every future fix.

guardian-service/src/api/token.service.ts
```diff
@@ -1401,4 +1401,418 @@ export async function tokenAPI(dataBaseServer: DatabaseServer, logger: PinoLogge
...
+            account.key = await (new Wallet()).getKey(user.walletToken, KeyType.KEY, user.did);
+        } else {
+            account.key = await (new Wallet()).getKey(user.walletToken, KeyType.RELAYER_ACCOUNT, `${user.did}/${account.account}`);
+        }
+        if (!account.key) {
+            throw new Error('Hedera account key not found');
+        }
+        notifier.completeStep(STEP_RESOLVE_ACCOUNT);
+
+        if (targetAccount === account.account) {
```
Pyatakov, line 1459: Self-transfer guard runs too late. The check happens after key resolution, please move it before the wallet lookup to fail fast.

guardian-service/src/api/token.service.ts
```diff
@@ -1401,4 +1401,418 @@ export async function tokenAPI(dataBaseServer: DatabaseServer, logger: PinoLogge
...
+            // FT transfer
+            notifier.startStep(STEP_RESOLVE_SERIALS);
+            if (!amount || amount <= 0) {
+                throw new Error('Amount is required for fungible token transfers and must be greater than 0');
+            }
+            notifier.completeStep(STEP_RESOLVE_SERIALS);
+
+            notifier.startStep(STEP_TRANSFER);
+            const decimals = token.decimals || 0;
+            const tokenValue = amount * Math.pow(10, decimals);
```
Pyatakov: Floating-point math here can produce non-integer results, e.g. `0.1 * Math.pow(10, 8)` evaluates to `10000000.000000002`, not `10000000`. Hedera requires an integer base-unit amount, so this either gets silently floored (off-by-one transfer) or rejected. Please reuse `PolicyUtils.tokenAmount(token, amount, 'round')` from `policy-service/src/policy-engine/helpers/utils.ts:192`, which already handles this correctly.

guardian-service/src/api/token.service.ts
```diff
@@ -1401,4 +1401,418 @@ export async function tokenAPI(dataBaseServer: DatabaseServer, logger: PinoLogge
...
+                }
+            }, {
+                priority: 20,
+                attempts: 3,
+                userId: user.id
+            });
+            notifier.completeStep(STEP_TRANSFER);
+
+            notifier.complete();
+            return { status: !!result, serials: resolvedSerials };
```
Pyatakov: When the NFT transfer fails on-chain, the worker returns null (see `worker-service/src/api/worker.ts:846`). The response correctly sets status: false, but still returns `serials: resolvedSerials`, making it look like those serials were transferred when nothing actually moved. Please return the serials only on success, e.g. `return { status: !!result, serials: result ? resolvedSerials : [] };.`

guardian-service/src/api/token.service.ts
```diff
@@ -1401,4 +1401,418 @@ export async function tokenAPI(dataBaseServer: DatabaseServer, logger: PinoLogge
...
+                    targetAccount,
+                    treasuryId: account.account,
+                    treasuryKey: account.key,
+                    tokenValue,
+                    transactionMemo: memo || '',
+                    payload: { userId: user.id }
+                }
+            }, {
+                priority: 20,
+                attempts: 3,
```
Pyatakov: `attempts: 3` on transfer tasks creates double-spend risk. The SDK helper handles `DUPLICATE_TRANSACTION` only inside one `_executeAndReceipt` call, so it doesn't help when `queue-service.ts:42-69` redispatches a failed task and worker rebuilds the `TransferTransaction` with a fresh auto-generated `TransactionId`, so a successful on-chain transfer whose receipt was lost in transit becomes a second on-chain transfer on retry. Set `attempts: 1` on `TRANSFER_FT` / `TRANSFER_NFT` here and let transient failures surface to the caller.
*Applies to all three call sites (`1506` FT-sync, `1578` NFT-sync, `1737` NFT-async).

guardian-service/src/api/token.service.ts
```diff
@@ -1401,4 +1401,418 @@ export async function tokenAPI(dataBaseServer: DatabaseServer, logger: PinoLogge
...
+            }
+            notifier.completeStep(STEP_RESOLVE_SERIALS);
+
+            notifier.startStep(STEP_TRANSFER);
+            const chunks: number[][] = [];
+            for (let i = 0; i < resolvedSerials.length; i += NFT_TRANSFER_MAX_BATCH_SIZE) {
+                chunks.push(resolvedSerials.slice(i, i + NFT_TRANSFER_MAX_BATCH_SIZE));
+            }
+
+            const chunkResults = await Promise.all(
```
Pyatakov: Consider running async NFT chunks sequentially. `Promise.all(chunks.map(...))` enqueues all chunks at once, so if chunk 2 fails, chunks 3–N still spend fees and retry. The reporting at `:1745` already shows which serials transferred, so the caller can retry the missing ones, meaning that this is not an asset-safety issue, just wasted fees on a bad path. A sequential `for…of await` with break on first failure would be cleaner.

api-gateway/src/api/service/tokens.ts
```diff
@@ -1116,6 +1116,157 @@ export class TokensApi {
...
+    @ApiInternalServerErrorResponse({
+        description: 'Internal server error.',
+        type: InternalServerErrorDTO
+    })
+    @ApiExtraModels(InternalServerErrorDTO)
+    @HttpCode(HttpStatus.OK)
+    async transferToken(
+        @AuthUser() user: IAuthUser,
+        @Param('tokenId') tokenId: string,
+        @Body() body: {
```
Pyatakov: Replace with a class-validator DTO (`TransferTokenDTO`) so the existing Nest `ValidationPipe` enforces `targetAccount` / `amount > 0` / `serialNumbers` integer constraints instead of the hand-rolled `if` blocks at `:1173-1178` and `:1251-1256.`
Same applies to `:1241`

Pyatakov (changes requested):
@OshanM, please address the mentioned issues. Additionally, please provide the required documentation.

guardian-service/src/api/token.service.ts
OshanM: Added a common `transferTokenCore` function and moved the duplicated codes there

guardian-service/src/api/token.service.ts
OshanM: Moved the guard check before the vault call

guardian-service/src/api/token.service.ts
OshanM: Since the `tokenAmount` function is in policy-service and is not a dependency of guardian-service, instead of using that, I have wrapped it around `Math.round(...)` which is what `tokenAmount` is doing internally as well.
I can move the `tokenAmount` to commons and then share it across other modules as well but it would require changes to other modules/files which are not covered in this PR.

guardian-service/src/api/token.service.ts
OshanM: Now the null values are filtered from `transferredSerials` array and this will not occur

guardian-service/src/api/token.service.ts
OshanM: Reduced the number of attempts from 3 to 1

guardian-service/src/api/token.service.ts
OshanM: Removed Promise.all() on NFT transfer and made it sequential to break on the first failure

api-gateway/src/api/service/tokens.ts
OshanM: Added TransferTokenDTO (with class validator decorators) and fixed the API bodies

Pyatakov (approved):

Issue #5799, opened 2026-03-06
[PSE - API] Add Block Completion Events to the Event Stream


palindaa:
**Labels:** `enhancement`, `api`, `events`, `high-priority`

### **Problem description**

When external systems integrate with Guardian and policy blocks are executed, these actions run asynchronously via workers. Currently, there is no reliable way for external systems to be notified when a specific block has completed execution (or failed). The event stream does not include block-level completion events, forcing integrators to resort to polling or guesswork to determine when a policy workflow step has finished.

This has been flagged as a high-priority issue by Xeptagon, as it prevents building reliable, event-driven integrations with Guardian policies. This feature can improve the quality of life for companies such as NoviqTech.

### **Requirements**

*   Add block completion events to Guardian's event stream for all policy block types.
    
*   Each block completion event should include: the block type, block tag/ID, policy ID, status (success/failure), relevant output data, and a timestamp.
    
*   Error events should include sufficient detail for external systems to diagnose and handle failures.
    
*   The event stream should support both success and error/failure scenarios for each block execution.
    

### **Definition of done**

*   Every policy block execution emits a completion event (success or failure) to the event stream.
    
*   External systems can subscribe to block-level events and react to individual workflow step completions in real time.
    
*   The events are documented (see CR-4).
    

### **Acceptance criteria**

*   \[ \] Block completion events are emitted to the event stream for all supported policy block types.
    
*   \[ \] Each event contains: block type, block tag/ID, policy ID, execution status (success/failure), output data, and timestamp.
    
*   \[ \] Failure events include error details sufficient for debugging.
    
*   \[ \] External systems can filter event streams by block type, policy ID, or status.
    
*   \[ \] Events are emitted in real time (not batched) so integrators can react promptly.
    
*   \[ \] Event payload schema is documented in the API reference (coordinated with CR-4).
    
*   \[ \] Integration tests verify event emission for at least 3 common block types in both success and failure scenarios.

PR #5979, opened 2026-04-21
docs: API & Event stream documentation and format changes #5798 #5801

gayanath8:
### Issue Labels
[https://github.com/hashgraph/guardian/issues/5798](https://github.com/hashgraph/guardian/issues/5798)
[https://github.com/hashgraph/guardian/issues/5801](https://github.com/hashgraph/guardian/issues/5801)

### Requirements

- Audit all existing Guardian API endpoints and identify documentation gaps.
- Improve documentation for every existing API endpoint, including: endpoint URL, HTTP method, authentication requirements, request payload schema with field descriptions, response payload schema, and error codes.
- Add documentation for any API endpoints that are currently undocumented.
- Generalize and clearly document each supported API block's execution payload structure.
- Keep provisions in the documentation for dynamic fields defined by users within policies.
- Provide example request/response pairs for common workflows.
- Update the Application Events Module documentation page with comprehensive, current information about all available event types, their payloads, and usage patterns.
- Fill in the External Events documentation page with all current event types supported by Guardian's event stream.
- Each event type should be documented with: event name, trigger condition, payload structure, and example usage.
- Include guidance on how external systems should subscribe to and consume these events.

### Summary of Changes

- Added documentation for previously undocumented API endpoints
- Introduced the Policy Block API Execution Payloads guide
- Reformatted all existing API documentation
- Filled documentation gaps and restructured event stream docs 
- Corrected and expanded the execution payloads guide with source-verified block entries 
- Updated SUMMARY.md to reflect the reorganized structure.

Issue #5797, opened 2026-03-06
[PSE - API] Address External Integration Challenges with Policy Filter Blocks


palindaa:
**Labels:** `enhancement`, `api`, `integration`, `needs-discussion`

### **Problem description**

Issue [5796](https://github.com/hashgraph/guardian/issues/5796) addresses read-only data access, but it does not resolve external integration issues with Guardian's filter blocks when actions need to be performed on filtered data. When a policy defines a Grid Container with a filter and then provides an action (e.g., "Approve") on the filtered grid, external parties using the read APIs must decode the policy structure to understand which actions can be performed on the listed data. This effectively forces external integrators to duplicate the policy logic in their own systems, which is error-prone, hard to maintain, and creates tight coupling to Guardian's internal policy design.

### **Requirements**

*   Investigate and design a solution that allows external systems to discover and execute actions defined within policy grid containers (e.g., approve, reject) without needing to decode and replicate the full policy logic.
    
*   External consumers should be able to:
    
    *   Retrieve the list of available actions for a given filtered dataset/grid.
        
    *   Execute those actions via API on specific records.
        
*   The solution should not require external integrators to have deep knowledge of the internal policy block structure.
    
*   **Note:** This item requires further discussion to determine the optimal approach, as it involves exposing policy workflow logic through APIs without breaking the abstraction.
    

### **Definition of done**

*   A design document or technical RFC has been reviewed and approved that outlines how external systems can discover and invoke actions on policy grid data.
    
*   An API (or set of APIs) is implemented that allows external systems to list available actions for a given policy grid and execute those actions on specific records.
    
*   External integrators no longer need to reverse-engineer policy block structures to perform actions on filtered data.
    

### **Acceptance criteria**

*   \[ \] An API endpoint (or endpoints) exists that returns the available actions for a given policy grid container and its filtered data.
    
*   \[ \] External systems can invoke actions (e.g., approve, reject) on specific records via API without decoding internal policy block logic.
    
*   \[ \] The API abstracts away the internal policy structure, exposing only the logical actions and their required parameters.
    
*   \[ \] Documentation explains how to discover and execute grid actions for a given policy.
    
*   \[ \] Integration test coverage demonstrates end-to-end action execution from an external system.

gayanath8:
Hi @Pyatakov 

We'd like to suggest the following implementation for this issue, a set of stable REST endpoints that allow external systems to discover and execute workflow actions on Guardian policy grids.
 
Endpoints will be  
[...]

PR #6043, opened 2026-05-08
feat: Add Block Completion Events to the Event Stream #5799

OshanM:
### Issue:
[#5799](https://github.com/hashgraph/guardian/issues/5799)

### Requirements:

- Add block completion events to Guardian's event stream for all policy block types
- Each block completion event should include: the block type, block tag/ID, policy ID, status (success/failure), relevant output data, and a timestamp
- Error events should include sufficient detail for external systems to diagnose and handle failures
- The event stream should support both success and error/failure scenarios for each block execution

### Changes done:

- A block execution now returns “trackingId” in the response which you can use to track the progress of the overall completion of the execution.
    - Each successful completion (this includes any downstream internal block executions as well) of a block, an external-`events.block_complete` event is emitted which can be listened using the webhook: `/api/events/subscribe` (ex: `curl -N http://<ip>:<port}>/api/events/subscribe`)

**_Note:_**
- This “trackingId” is not persistent. This uses the internal “recordActionId" which Guardian uses for each block execution session (a record session is started for each block execution, which is retained on memory). The “recordActionId" is returned as “taskId” in the response

### Block completion flow:

```mermaid
sequenceDiagram
    autonumber
    actor Caller as External System
    participant GW as API Gateway
    participant PS as policy-service<br/>(BlockTreeGenerator)
    participant RAS as RecordActionStep
    participant PL as PolicyLink<br/>(Downstream Blocks)
    participant W as Worker Service<br/>(IPFS / Hedera)
    participant GS as guardian-service
    participant AE as application-events
    participant WH as Webhook / SSE

    Caller->>GW: POST /policies/{id}/blocks/{blockId}
    GW->>PS: NATS: SET_BLOCK_DATA {user, blockId, data}

    PS->>PS: Validate block availability
    PS->>RAS: new RecordActionStep(completionCallback)
    PS->>PS: blockSetData() — runs block.setData()

    Note over PS,W: Block may await internal async ops (IPFS, Hedera)<br/>before returning — these complete before blockSetData() returns

    PS->>RAS: actionstep.finish() — arms 1-second quiesce timer

    PS-->>GW: MessageResponse { trackingId, ...blockOutput }
    GW-->>Caller: 200 OK { trackingId: "uuid-...", ... }

    Note over Caller: Caller stores trackingId for correlation

    loop For each downstream block triggered
        PS->>PL: triggerEvents() → PolicyLink.run()
        PL->>RAS: actionStatus.inc()
        PL->>PL: execute block callback (async)
        PL->>W: Worker task (ADD_FILE, token ops, etc.)
        W-->>PL: Task result (CID, txId, etc.)
        PL->>RAS: actionStatus.dec()
    end

    Note over RAS: counter reaches 0 → 1s timer fires

    RAS->>PS: completionCallback(id, timestamp, errors[])
    PS->>GS: NATS: policy-event-block-complete-broadcast<br/>{ trackingId, blockType, blockTag, policyId,<br/>  userId, status, outputData, timestamp }

    GS->>AE: NATS: external-events.block_complete { ...payload }

    AE->>WH: HTTP POST to registered webhook URL
    Note over AE,WH: OR streams via GET /api/events/subscribe (SSE)

    WH-->>Caller: Event received — match by trackingId
```

### Dev test examples:

**Block execution and response**:
image

**Event-stream events**:

Initial record persist step and recordActionId generation (which was returned as trackingId):
image

Policy event completion broadcast:
image

Policy events block complete (this marks the completion):
image


policy-service/src/policy-engine/block-tree-generator.ts
```diff
@@ -254,42 +287,59 @@ export class BlockTreeGenerator extends NatsService {
...
+                    blockType: block.blockType,
+                    blockTag: block.tag,
+                    blockId: block.uuid,
+                    policyId,
+                    userId: userFull.did,
+                };
+
+                const actionstep = new RecordActionStep(
+                    (recordActionId, actionTimestemp, errors) => {
+                        RecordUtils.RecordSetBlockData(policyId, userFull, block, data, recordActionId, actionTimestemp);
```
Pyatakov: typo, actionTimestemp -> actionTimestamp

Pyatakov (commented):
@OshanM please address the issues below.

policy-service/src/policy-engine/block-tree-generator.ts
Pyatakov, line 1: The two `SET_BLOCK_DATA*` handlers (lines 191–267 and 269–344) are ~60 lines of identical logic. Any future fix to the completion-event flow has to be made in two places, and one will eventually drift out of sync. The only thing that genuinely differs between the handlers is how the block is looked up (`GetBlockByUUID` vs `GetBlockByTag`); everything after the lookup is the same. Extracting a `private executeSetBlockData(block, userFull, data, policyId, syncEvents, history, waitRemotePolicy)` method on `BlockTreeGenerator` keeps the lookup decision in each handler (where it belongs) and gives a single place to evolve the completion-event behavior. Then both handlers collapse to ~10 lines each:
```
this.getPolicyMessages(PolicyEvents.SET_BLOCK_DATA, policyId, async (msg) => {
    const { user, blockId, data, syncEvents, history, waitRemotePolicy } = msg;
    const userFull = await this.getUser(policyInstance, user);
    const block = PolicyComponentsUtils.GetBlockByUUID<IPolicyInterfaceBlock>(blockId);
    return this.executeSetBlockData(block, userFull, data, policyId, syncEvents, history, waitRemotePolicy);
});

this.getPolicyMessages(PolicyEvents.SET_BLOCK_DATA_BY_TAG, policyId, async (msg) => {
    const { user, tag, data, syncEvents, history, waitRemotePolicy } = msg;
    const userFull = await this.getUser(policyInstance, user);
    const block = PolicyComponentsUtils.GetBlockByTag<IPolicyInterfaceBlock>(policyId, tag);
    return this.executeSetBlockData(block, userFull, data, policyId, syncEvents, history, waitRemotePolicy);
});
```

Pyatakov, line 1: typo `actionTimestemp` -> `actionTimestamp` – lines 222, 223, 299, 300

Pyatakov, line 1: #### Delete the four `as IBlockCompleteEvent` suffixes (lines 231, 249, 308, 326).
#### Rationale:
Each call to `blockCompleteEvent(...)` passes a freshly constructed object literal. Because `blockCompleteEvent` is already typed as `(event: IBlockCompleteEvent) => void`, TypeScript performs structural validation and excess-property checks on that literal in argument position and you get the exact same type guarantee `as IBlockCompleteEvent` is supposed to give you, for free. But the cast is not just redundant, it actively weakens the type system. A type assertion tells the compiler "trust me, this matches" and silences errors. So:
- If `IBlockCompleteEvent` later adds a required field, the cast hides the missing-property error and you get a runtime bug instead of a build break.
- If you misspell a field `(errorDetalis: ...)`, the cast silences the "object literal may only specify known properties" check. The misspelled property is silently included and the real one is silently `undefined`.

Pyatakov (changes requested):

policy-service/src/policy-engine/block-tree-generator.ts
OshanM: Moved the duplicated code block to `executeSetBlockData` function

OshanM: Fixed the typo here. However, this is reflected through the chain starting from `RecordActionStep.timestemp` property declaration typo, which is reflected in a couple of more files. I have not refactored those and left them as it is for now.

OshanM: Removed the `as IBlockCompleteEvent` cast and added value check for baseEvent creation time to pass empty strings if undefined

OshanM:
> @OshanM please address the issues below.

@Pyatakov all three are resolved now, please check

Pyatakov (approved):

Pyatakov:
@OshanM looks good now, thanks. The only thing left is documentation. Could you please add / update the necessary documentation in the /docs folder?

OshanM:
> @OshanM looks good now, thanks. The only thing left is documentation. Could you please add / update the necessary documentation in the /docs folder?

@Pyatakov I have added the `block_complete` event to the events table of the existing external events README. However, a new documentation needs to be created for the block and tag APIs to mention the overall usage/working of the API changes addressed here. 
Since this is being addressed in PR [#5979](https://github.com/hashgraph/guardian/pull/5979) which has added the base documentation required to be updated from this PR, I would like to know whether it is acceptable to be updated in another PR after #5797 is merged so we can avoid unnecessary conflicts (or hold this until #5797 is merged).

Pyatakov:
Thank you @OshanM. So if I understand correctly, the rest of the required documentation will be added with PR [5979](https://github.com/hashgraph/guardian/pull/5979). That's totally acceptable. Please confirm, and I'll merge the current PR.

OshanM:
> Thank you @OshanM. So if I understand correctly, the rest of the required documentation will be added with PR [5979](https://github.com/hashgraph/guardian/pull/5979). That's totally acceptable. Please confirm, and I'll merge the current PR.

Yes, if this PR is merged, we can add the document directly into the pending PR #5979

Issue #5796, opened 2026-03-06
[PSE - API] Provide Dynamic Read API Endpoint for Guardian Policy Data


palindaa:
**Labels:** `enhancement`, `api`, `data-access`, `high-priority`

### **Problem description**

Guardian policy filters are currently applied to grids internally and work according to the policy state within the Guardian UI. When external systems integrate with Guardian and need to replicate filtered views of policy data, they must reverse-engineer and replicate Guardian's internal filtering logic for grid/blocks externally. This is not straightforward and creates a significant barrier to integration.

There is no direct API to query the data committed by Guardian policies in a flexible, filterable manner. Integrators have to work around this limitation, leading to fragile and tightly-coupled integration code.

### **Requirements**

*   Provide a separate, dynamic API endpoint that allows external consumers to read data committed by Guardian policies.
    
*   The endpoint should execute queries against MongoDB and support dynamic filter fields provided by the API caller.
    
*   API users should be able to filter by policy IDs, schema names, and system field names.
    
*   Users can discover available policy IDs and schema names via the existing /policies and /schema endpoints.
    
*   All system field names that can be used for filtering must be documented so external users can reference them.
    
*   This endpoint addresses read-only data access concerns.
    

### **Definition of done**

*   A new API endpoint exists that returns policy-committed data with support for dynamic filtering.
    
*   The endpoint accepts query parameters for policy ID, schema name, and arbitrary system/custom field filters.
    
*   System field names are documented in the API reference.
    
*   External systems can retrieve filtered policy data without replicating Guardian's internal filtering logic.
    

### **Acceptance criteria**

*   \[ \] A new API endpoint is available that accepts dynamic filter parameters (policy ID, schema name, custom fields).
    
*   \[ \] The endpoint returns data committed by Guardian policies, filtered according to the provided parameters.
    
*   \[ \] Users can discover filterable policy IDs via /policies and schema names via /schema endpoints.
    
*   \[ \] All system field names available for filtering are documented in the API reference.
    
*   \[ \] The endpoint supports pagination for large result sets.
    
*   \[ \] Access control is enforced — only authorized users can query policy data.
    
*   \[ \] Response format is consistent and well-documented.

PR #6074, opened 2026-05-14
feat: Provide Dynamic Read API Endpoint for Guardian Policy Data #5796

gayanath8:
### Issue Labels
[https://github.com/hashgraph/guardian/issues/5796](https://github.com/hashgraph/guardian/issues/5796)

### Requirements

- Provide a separate, dynamic API endpoint that allows external consumers to read data committed by Guardian policies.
- The endpoint should execute queries against MongoDB and support dynamic filter fields provided by the API caller.
- API users should be able to filter by policy IDs, schema names, and system field names.
- Users can discover available policy IDs and schema names via the existing /policies and /schema endpoints.
- All system field names that can be used for filtering must be documented so external users can reference them.
- This endpoint addresses read-only data access concerns.

### Summary of Changes

- **`GET  /policy-data/query`** — Read-only endpoint that returns paginated Verifiable Credential (VC) documents committed by a published policy. The caller supplies a policy ID, schema name and optional field-level filters; the service resolves the schema IRI internally, applies safe MongoDB operator translation, and returns matching documents with pagination metadata.

> End User should have JWT Bearer token present and valid signature and hold one of POLICIES_POLICY_AUDIT, POLICIES_POLICY_MANAGE.

> The endpoint does not check whether the caller's DID is registered as a participant in the policy. However, POLICIES_POLICY_MANAGE callers are now restricted to policies owned by their own Standard Registry tenant, a policy belonging to a different tenant returns 403. POLICIES_POLICY_AUDIT (Auditor) is intentionally exempt from this restriction, since Auditors are a cross-organization role by design.

> The endpoint does not read permissions from documentsSourceAddon blocks. It returns all documents matching the filter. Callers who need own-only results must supply an owner filter themselves.

>  Revoked documents are included unless filtered out with hederaStatus ne REVOKE.

### Query Parameters

| Parameter | Required | Type | Description |
|-----------|----------|------|-------------|
| `policyId` | yes | string | MongoDB ObjectId of the published policy |
| `schemaName` | yes | string | Schema name as registered in the policy (e.g. `Installer Activity Report`) |
| `filters` | no | JSON string | Field-level filter map — see Filter Fields section |
| `sort` | no | string | Field to sort by. Prefix `-` for descending (e.g. `-createDate`) |
| `page` | no | number | 1-based page number (default: `1`) |
| `pageSize` | no | number | Results per page, max `200` (default: `20`) |

#### 1. System Fields (direct MongoDB document properties)

| Field | Type | Notes |
|-------|------|-------|
| `owner` | string | DID of the document owner |
| `assignedTo` | string | DID assigned to |
| `assignedToGroup` | string | Group assigned to |
| `group` | string | User group that created the doc |
| `hederaStatus` | string | `NEW` \| `ISSUE` \| `REVOKE` \| `SUSPEND` \| `RESUME` \| `FAILED` |
| `signature` | number | Document signature numeric code |
| `type` | string | Internal block document type |
| `draft` | boolean | Whether document is in draft |
| `edited` | boolean | Whether this version supersedes an earlier one |
| `disconnected` | boolean | Disconnected from policy |
| `policyId` | string | MongoDB ObjectId of owning policy |
| `schema` | string | Schema IRI (e.g. `#MySchema`) |
| `tag` | string | Policy block tag that produced the document |
| `messageId` | string | HCS message ID |
| `topicId` | string | Hedera topic ID |
| `hash` | string | Document content hash |
| `createDate` | string | ISO 8601 date |
| `updateDate` | string | ISO 8601 date |

#### 2. Option Fields

Any `option.*` sub-field stored by the policy engine.

> e.g. `option.status` (set by `sendToGuardianBlock` options array)

#### 3. Document Credential Fields

Any `document.credentialSubject[n].fieldX` path cached by the grid columns.

> e.g. `document.credentialSubject[0].field3`

Bracket notation `[0]` and dot notation `.0.` are both accepted.

### Filter Operators

| Token | MongoDB | Example |
|-------|---------|---------|
| `contains` | `$regex` | `{"op": "contains", "value": "Voluptatem"}` |
| `eq` | `$eq` | `{"op": "eq", "value": "Pass"}` |
| `ne` | `$ne` | `{"op": "ne", "value": "Fail"}` |
| `in` | `$in` | `{"op": "in", "value": ["Pass", "Conditional"]}` |
| `nin` | `$nin` | `{"op": "nin", "value": ["Germany"]}` |
| `gt` | `$gt` | `{"op": "gt", "value": 5000}` |
| `gte` | `$gte` | `{"op": "gte", "value": 80}` |
| `lt` | `$lt` | `{"op": "lt", "value": 100}` |
| `lte` | `$lte` | `{"op": "lte", "value": 70}` |

#### **1. No Filters Defaults**

> image

#### **2. Custom Page and Size**

> image

#### **3. Sort By Filed 3 (Number Field)**

> https://github.com/user-attachments/assets/ab501371-952d-4346-8e3b-df576cd03f80

#### **4. Different Filters**

image
image
image

#### **5. Multiple Filters**

image

#### **Validations**

> image
> image
> image
> image
> image
> image
> image

> (Page Size Set 200 Max)
> image

> (Page default to 1 when missing)
> image


gayanath8:
Hi @Pyatakov 
    Regarding the two concerns raised on permission and data querying we would like to get your feedback, also here are our proposed methods on those
    As for the permission concern we can apply that caller must be a valid assigned user or owner of the particular policy. Additionally, we can enforce role-based access if the caller has no active role within the policy, the request will be rejected.
    For data querying part we can Limit the number of filter entries per request (e.g. max 5) to prevent wide AND queries and also apart from can apply a standard API-gateway-level rate limit on the endpoint per user
    Also, If you would like to allow only a particular set of fields, please mention them and we will restrict the endpoint accordingly

api-gateway/src/api/service/policy-data.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    constructor(private readonly logger: PinoLogger) { }
+
+    /**
+     * Returns a paginated list of VC documents committed by the given policy,
+     */
+    @Get('/query')
+    @Auth(
+        Permissions.POLICIES_POLICY_READ,
+        Permissions.POLICIES_POLICY_EXECUTE,
```
Pyatakov: The endpoint lets any user holding POLICIES_POLICY_READ read every document of every published policy. Unlike existing policy data endpoints, it does not check whether the caller's DID is registered as a participant, and it does not honor `documentsSourceAddon` permission filtering. The PR description acknowledges this, but the constraint is enforced nowhere in code. `ownerDid` is plumbed through the service signature but only used for logging, which make it even more confusing.
Proposed solution:
- Add participant/owner scoping using the already-passed ownerDid (mirror what getVcDocuments does); or
- Restrict to POLICIES_POLICY_AUDIT/MANAGE only and document that this is intentional cross-policy access.

guardian-service/src/api/policy-data.service.ts
```diff
@@ -0,0 +1,215 @@
...
+            const _page = Math.max(1, page || 1);
+            const _pageSize = Math.min(200, Math.max(1, pageSize || 20));
+            const offset = (_page - 1) * _pageSize;
+
+            const sortOptions: Record<string, 'ASC' | 'DESC'> = {};
+            if (sortField) {
+                if (sortField.startsWith('-')) {
+                    sortOptions[normalisePath(sortField.slice(1))] = 'DESC';
+                } else {
+                    sortOptions[normalisePath(sortField)] = 'ASC';
```
Pyatakov: `sortField` is normalized and dropped directly into `orderBy` with no whitelist. A caller can sort by any field, including unindexed paths like `document.credentialSubject.0.field17`. On a large policy collection this forces an in-memory sort.

Proposed resolution:
Whitelist sortable fields. Reuse the filter whitelist (VC_DOCUMENT_SYSTEM_FIELDS, option.\*, document.\*) or  (preferably) restrict sort to indexed top-level fields (`createDate`, `updateDate`, `owner`, etc.). Reject the request with a 400 listing allowed values.

guardian-service/src/api/policy-data.service.ts
```diff
@@ -0,0 +1,215 @@
...
+            throw new Error(
+                `Unknown operator "${entry.op}" for field "${field}". Allowed: ${[...ALL_OPERATORS].join(', ')}.`
+            );
+        }
+
+        if (entry.op === 'contains') {
+            if (typeof entry.value !== 'string') {
+                throw new Error(`Operator "contains" requires a string value for field "${field}".`);
+            }
+            q[field] = { $regex: escapeRegex(entry.value), $options: 'i' };
```
Pyatakov, line 106: `escapeRegex` correctly escapes the input, but there is no length limit on the search string, and contains is allowed on every whitelisted field including unindexed credential-subject paths. A multi-KB regex on a non-indexed field triggers a collection scan.

Proposed resolution:
Cap value.length to a small constant (e.g. 256) and return 400 if exceeded.

guardian-service/src/api/policy-data.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        if (entry.op === 'in' || entry.op === 'nin') {
+            if (!Array.isArray(safeValue)) {
+                safeValue = typeof safeValue === 'string'
+                    ? safeValue.split(',').map((s: string) => s.trim())
+                    : [safeValue];
+            }
+        }
+
+        const mongoOp = OPERATOR_MAP[entry.op as Exclude<FilterOperator, 'contains'>];
+        q[field] = { [mongoOp]: safeValue };
```
Pyatakov, line 143: `safeValue` is typed unknown and passed through without checking that it is a primitive. Caller can submit `{op:"eq", value: {nested: "object"}}` and get confusing query semantics. For `$in`/`$nin` the array elements are also unconstrained.

Proposed resolution:
Validate values explicitly
- For `eq`/`ne`/`gt`/`gte`/`lt`/`lte`: require `typeof value `to be `string | number | boolean` (and `null` only for `eq`/`ne`).
- For `in`/`nin`: require every array element to be a primitive.
- Return a descriptive 400 listing the expected value type.

api-gateway/src/api/service/policy-data.ts
```diff
@@ -0,0 +1,317 @@
...
+import { Guardians, InternalException } from '#helpers';
+
+const MAX_PAGE_SIZE = 200;
+const DEFAULT_PAGE_SIZE = 20;
+
+/**
+ * Validates and parses the `filters` query-string value.
+ * Returns the parsed object, or throws with a 400.
+ */
+function parseFilters(
```
Pyatakov, line 44: Both layers (this and guardian-service/src/api/policy-data.service.ts:55 - buildMongoFilter) whitelist the same field set and operator set, but the gateway version skips operator-specific shape checks (`contains` requires a string; `in`/`nin` require an array). It also whitelists `rawField` directly, while the service first calls `normalisePath` and re-whitelists, so the two checks aren't equivalent on bracket-notation paths.

Proposed resolution:
Leave the gateway as a thin JSON parser, do all field/operator/value validation in `buildMongoFilter`.

guardian-service/src/api/policy-data.service.ts
```diff
@@ -0,0 +1,215 @@
...
+                return new MessageError(
+                    `Policy "${policyId}" is not published. Only published policy data is queryable.`,
+                    403
+                );
+            }
+
+            const schemas = await DatabaseServer.getSchemas(
+                { name: schemaName, topicId: policy.topicId },
+                { limit: 1 }
+            );
```
Pyatakov: Schema lookup is non-deterministic. If a topic has multiple versions of the same-named schema, the returned schema depends on insertion order.

Expected resolution:
Either order by version descending and take the latest, or accept the schema IRI directly as an alternative input. Document the tiebreaker in the endpoint reference.

api-gateway/src/api/service/policy-data.ts
```diff
@@ -0,0 +1,317 @@
...
+import {
+    VC_DOCUMENT_SYSTEM_FIELDS,
+    VC_DOCUMENT_OPTION_PREFIX,
+    VC_DOCUMENT_DOCUMENT_PREFIX,
+} from '@guardian/interfaces';
+import { AuthUser, Auth } from '#auth';
+import { Guardians, InternalException } from '#helpers';
+
+const MAX_PAGE_SIZE = 200;
+const DEFAULT_PAGE_SIZE = 20;
```
Pyatakov: `MAX_PAGE_SIZE` / `DEFAULT_PAGE_SIZE` are duplicated (the second one is at guardian-service/src/api/policy-data.service.ts:183). Two sources of truth for the same limits. If one is bumped without the other, behavior diverges silently.

Proposed resolution:
- Export `POLICY_DATA_MAX_PAGE_SIZE` / `POLICY_DATA_DEFAULT_PAGE_SIZE` from `@guardian/interfaces` and use them in both layers,  or
- treat the gateway-side ones as authoritative and drop the others (service-side).

Pyatakov (changes requested):

api-gateway/src/api/service/policy-data.ts
gayanath8: Updated the permissions on the endpoint only restricting to POLICIES_POLICY_AUDIT/MANAGE apart from that MANAGE callers are further scoped to their own Standard Registry's policies, AUDIT is exempt from that scoping, since Auditors audit across organizations by design.

guardian-service/src/api/policy-data.service.ts
gayanath8: Whitelisted sort to the same system-field set already used for filtering (owner, hederaStatus, createDate, etc.) — option.*/document.* paths are no longer sortable, since those aren't indexed and were the actual risk. Requesting any other field now returns 400 listing the allowed values.

guardian-service/src/api/policy-data.service.ts
gayanath8: Capped contains values at 256 characters requests over that limit now return 400

api-gateway/src/api/service/policy-data.ts
gayanath8: Added explicit type validation for every operators eq/ne accept string, number, boolean or null, and gt/gte/lt/lte accept string, number, or boolean but no null also in/nin require every array element to be a same string, number, or boolean.

guardian-service/src/api/policy-data.service.ts
gayanath8: picking the highest version via the existing ModelHelper.versionCompare helper. if multiple versions of the same schema name exist under the policy's topic, the latest version is used.

api-gateway/src/api/service/policy-data.ts
gayanath8: moved MAX_PAGE_SIZE/DEFAULT_PAGE_SIZE into @guardian/interfaces as POLICY_DATA_MAX_PAGE_SIZE/POLICY_DATA_DEFAULT_PAGE_SIZE

gayanath8:
Hi @Pyatakov 
All the changes and fixes you requested in this review have been addressed and also verified with our internal QA team. Please have a look and let me know if anything still needs adjustment.

guardian-service/src/api/policy-data.service.ts
Pyatakov: @gayanath8 just want to make sure this one was addressed as well

api-gateway/src/middlewares/validation/schemas/policy-data.dto.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    @ApiProperty({ type: Number, example: 62 })
+    totalPages: number;
+}
+
+export class PolicyDataQueryInfoDTO {
+    @ApiProperty({ type: String, example: Examples.DB_ID })
+    policyId: string;
+
+    @ApiProperty({ type: String, example: '#MySchema' })
```
Pyatakov: `schemaName` example is `#MySchema` (an IRI), but this param takes a human-readable name. Update the example to a name (e.g. `MRV Data`) to match the docs and the actual input.

api-gateway/src/api/service/policy-data.ts
```diff
@@ -0,0 +1,284 @@
...
+                filters = parseFilters(rawFilters);
+            } catch (e: any) {
+                return res.status(HttpStatus.BAD_REQUEST).send({
+                    statusCode: 400,
+                    message: e.message,
+                });
+            }
+
+            const page = Math.max(1, parseInt(rawPage, 10) || 1);
+            const pageSize = Math.min(POLICY_DATA_MAX_PAGE_SIZE, Math.max(1, parseInt(rawPageSize, 10) || POLICY_DATA_DEFAULT_PAGE_SIZE));
```
Pyatakov: `page`/`pageSize` are bounded in both the service (`policy-data.service.ts:232-233`) and again here. Seems redundant, consider validating in one place (the service, since it's the source of truth) and letting the controller pass values through

Pyatakov (changes requested):
No tests are included. Please add unit tests for `buildMongoFilter` (each operator, the rejection paths, regex escaping, `in` string-splitting) and for the handler branches (tenant mismatch -> 403, not published -> 403, policy/schema not found -> 404). This is the security-critical validation layer, so table-driven coverage here is important.

api-gateway/src/middlewares/validation/schemas/policy-data.dto.ts
gayanath8: Updated the example to MRV Data

api-gateway/src/api/service/policy-data.ts
gayanath8: Kept the validations service side and removed from the controller

gayanath8:
> No tests are included. Please add unit tests for `buildMongoFilter` (each operator, the rejection paths, regex escaping, `in` string-splitting) and for the handler branches (tenant mismatch -> 403, not published -> 403, policy/schema not found -> 404). This is the security-critical validation layer, so table-driven coverage here is important.

Hi @Pyatakov 

Added the test cases for buildMongoFilter  and handlePolicyDataQuery along with above two change requests. Please review them and let me know if anything needs to be change 

Test Result:
image

Pyatakov (approved):

PR #6075, opened 2026-05-14
fix: terminate worker threads after block evaluation in policy-service

vshvets-bc:
Three policy blocks spawn worker_threads Workers via `new Worker(...)` but never call worker.terminate(). 
Node Worker threads do not auto-exit when the script body returns — they keep the V8 isolate alive (~30 MB each) waiting for more messages.
Every formula / custom-logic / data-transformation invocation leaked one worker thread. A heavy policy could accumulate multi-GB of leaked workers.

Adds a small cleanup() helper that calls worker.terminate() (with a swallowed .catch() for already-exited workers, e.g. via OOM-kill). Called on the terminating message, on error, and on every reject path.

- math-block.ts: terminate on the single 'done' message + on error
- custom-logic-block.ts (JS branch): terminate after the final 'done' (data.final === true) so debug messages are still forwarded for multi-emit custom-logic runs. 
- data-transformation-addon.ts: terminate after final result + on error

Behavior unchanged for callers; resources are now released after each block evaluation.

### DIFF
--- a/policy-service/src/policy-engine/blocks/custom-logic-block.ts
+++ b/policy-service/src/policy-engine/blocks/custom-logic-block.ts
@@ -387,18 +387,26 @@ export class CustomLogicBlock {
                                 tablesPack
                             },
                         });
+                    // Terminate the worker after the final 'done' (or on error) so
+                    // the V8 isolate is released. Without this the worker thread
+                    // stays alive after the script body returns, leaking ~30 MB per
+                    // custom-logic invocation.
+                    const cleanup = () => { worker.terminate().catch(() => { /* noop */ }); };
                     worker.on('error', (error) => {
+                        cleanup();
                         reject(error);
                     });
                     worker.on('message', async (data) => {
                         try {
                             if (data?.type === 'done') {
                                 await done(data.result, data.final);
+                                if (data.final) cleanup();
                             }
                             if (data?.type === 'debug') {
                                 ref.debug(data.message);
                             }
                         } catch (error) {
+                            cleanup();
                             reject(error);
                         }
                     });
--- a/policy-service/src/policy-engine/blocks/custom-logic-block.ts
+++ b/policy-service/src/policy-engine/blocks/custom-logic-block.ts
@@ -387,11 +387,8 @@ export class CustomLogicBlock {
                                 tablesPack
                             },
                         });
-                    // Terminate the worker after the final 'done' (or on error) so
-                    // the V8 isolate is released. Without this the worker thread
-                    // stays alive after the script body returns, leaking ~30 MB per
-                    // custom-logic invocation.
-                    const cleanup = () => { worker.terminate().catch(() => {}); };
+                    // Release the worker's V8 isolate; without this each invocation leaks ~30 MB.
+                    const cleanup = () => { worker.terminate().catch(() => { }); };
                     worker.on('exit', (code) => {
                         cleanup();
                         if (code !== 0 && code !== null) {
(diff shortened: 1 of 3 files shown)


policy-service/src/policy-engine/blocks/data-transformation-addon.ts
```diff
@@ -57,23 +50,32 @@ export class DataTransformationAddon {
                 },
             });
 
+            // Terminate the worker after the final result / on error so the V8
+            // isolate is released. Without this the worker thread stays alive
+            // after the script body returns, leaking ~30 MB per transformation.
+            const cleanup = () => { worker.terminate().catch(() => { /* noop */ }); };
+
```
Pyatakov: No '`exit`' handler. Additional code block needs to be added in case if a worker exits without emitting '`done`' and without emitting '`error`' (e.g., `process.exit(0)` inside the worker, OOM-kill, or any unhandled internal throw that terminates the isolate) 

```suggestion
worker.on('exit', (code) => {
    cleanup();
    if (code !== 0 && code !== null) {
        reject(new Error(`Data transformation worker exited with code ${code}`));
    }
});
```

policy-service/src/policy-engine/blocks/data-transformation-addon.ts
```diff
@@ -57,23 +50,32 @@ export class DataTransformationAddon {
                 },
             });
 
+            // Terminate the worker after the final result / on error so the V8
+            // isolate is released. Without this the worker thread stays alive
+            // after the script body returns, leaking ~30 MB per transformation.
+            const cleanup = () => { worker.terminate().catch(() => { /* noop */ }); };
```
Pyatakov: Please remove '/* noop */' here and in the two other places

policy-service/src/policy-engine/blocks/data-transformation-addon.ts
```diff
@@ -57,23 +50,32 @@ export class DataTransformationAddon {
                 },
             });
 
+            // Terminate the worker after the final result / on error so the V8
+            // isolate is released. Without this the worker thread stays alive
+            // after the script body returns, leaking ~30 MB per transformation.
```
Pyatakov: One-line comment is enough, please replace here and in the other two places with:
`// Release the worker's V8 isolate; without this each invocation leaks ~30 MB.`

policy-service/src/policy-engine/blocks/math-block.ts
```diff
@@ -87,15 +84,21 @@ export class MathBlock {
         return new Promise<IPolicyDocument>(async (resolve, reject) => {
             const workerFile = path.join(path.dirname(filename), '..', 'helpers', 'workers', 'math-worker.js');
             const worker = new Worker(workerFile, { workerData });
+            // Terminate the worker once it finishes so the V8 isolate is released.
+            // Otherwise every formula evaluation leaks a worker thread (~30 MB).
+            const cleanup = () => { worker.terminate().catch(() => { /* noop */ }); };
```
Pyatakov:
```suggestion
            const cleanup = () => { worker.terminate().catch(() => {}); };
            worker.on('exit', (code) => {
                cleanup();
                if (code !== 0 && code !== null) {
                    reject(new Error(`Math worker exited with code ${code}`));
                }
            });
```

policy-service/src/policy-engine/blocks/custom-logic-block.ts
```diff
@@ -397,18 +387,26 @@ export class CustomLogicBlock {
                                 tablesPack
                             },
                         });
+                    // Terminate the worker after the final 'done' (or on error) so
+                    // the V8 isolate is released. Without this the worker thread
+                    // stays alive after the script body returns, leaking ~30 MB per
+                    // custom-logic invocation.
+                    const cleanup = () => { worker.terminate().catch(() => { /* noop */ }); };
```
Pyatakov:
```suggestion
                    const cleanup = () => { worker.terminate().catch(() => {}); };
                    worker.on('exit', (code) => {
                        cleanup();
                        if (code !== 0 && code !== null) {
                            reject(new Error(`Custom logic worker exited with code ${code}`));
                        }
                    });
```

Pyatakov (changes requested):

Pyatakov (approved):

PR #6076, opened 2026-05-14
fix: require authentication on GET /demo/registered-users (CVE-2026-45248)

Christbowel:
The `GET /api/v1/demo/registered-users` endpoint was publicly accessible without authentication, exposing all registered user accounts (usernames, DIDs, roles, policy assignments).

## Changes

**`api-gateway/src/api/service/demo.ts`**
- Added `@Auth(Permissions.DEMO_KEY_CREATE)` guard to `registeredUsers()`, matching the sibling `/random-key` and `/push/random-key` endpoints
- Injected `@AuthUser()` to pass authenticated user context
- Updated `InternalException` call to use `user.id` instead of `null`

**`e2e-tests/cypress/e2e/api-tests/001_demo/getRegisteredUsers.cy.js`**
- Updated positive test to authenticate via `Authorization.getAccessToken()`
- Added negative tests: no token, invalid token, empty token → expect `401 UNAUTHORIZED`

## References
-  CVE-2026-45248
- CWE-306 (Missing Authentication for Critical Function)

### DIFF
--- a/api-gateway/src/api/service/demo.ts
+++ b/api-gateway/src/api/service/demo.ts
@@ -16,6 +16,9 @@ export class DemoApi {
      * Returns list of registered users
      */
     @Get('/registered-users')
+    @Auth(
+        Permissions.DEMO_KEY_CREATE,
+    )
     @ApiOperation({
         summary: 'Returns list of registered users.',
         description: 'Returns list of registered users.',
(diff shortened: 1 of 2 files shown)


Pyatakov:
@Christbowel Thanks for the report, but this is a false positive.

The `/demo/*` endpoints, including `/demo/registered-users`, are intentionally unauthenticated – they exist solely to support local development and testing environments. They are not part of the production API surface.
During a production build, the entire `demo` module is removed from the compiled output by the build process. This means these endpoints are physically absent from any production deployment and cannot be reached or exploited in a live environment.
Because of this, adding authentication to a demo-only endpoint doesn't improve security in any meaningful way, but adds friction for developers running local test setups, which is the intended audience for these endpoints.
We'll be closing this PR as a false positive. If you believe there's a scenario where the demo module could end up in a production build, we'd be happy to revisit – please open a discussion with details.

PR #6101, opened 2026-05-20
fix: FLD undefined values from source document

dariy-rised:
**Description**:
* Add clone sub-schema fields to prevent overwritten paths
* Prioritise dataSource over legacy dataType in sendToGuardianBlock

**Related issue(s)**:

### DIFF
--- a/interfaces/src/helpers/schema-helper.ts
+++ b/interfaces/src/helpers/schema-helper.ts
@@ -361,9 +361,6 @@ export class SchemaHelper {
      * Clone a SchemaField array
      */
     private static cloneFields(fields: SchemaField[]): SchemaField[] {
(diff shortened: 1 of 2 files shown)


interfaces/src/helpers/schema-helper.ts
```diff
@@ -357,6 +357,22 @@ export class SchemaHelper {
         return results;
     }
 
+    /**
+     * Clone a SchemaField array
+     */
+    private static cloneFields(fields: SchemaField[]): SchemaField[] {
+        if (!Array.isArray(fields)) {
+            return fields;
+        }
```
Pyatakov: Redundant check

Pyatakov (changes requested):

PR #6123, opened 2026-05-26
test: add policy-service coverage and related service tests

### LINKED ISSUE #6102: Unit test coverage for guardian/policy-service
Scope

  1. Block-factory about contract tests. One test file per block. Each asserts the documented capabilities:
  post, get, children, control, defaultEvent, deprecated, input, output. Run as a static contract check — the
  test doesn't instantiate the block, it imports the factory and asserts the about object shape. Cost: ~62 block
   files × ~10 assertions each = ~620 quick assertions, runs in well under a second. Catches accidental flag
  flips during refactors.

  2. Per-block runtime tests with mocked dependencies. For each block type that has runtime behavior beyond
  about:
  - Mock the policy-engine context (PolicyComponentsUtils, PolicyUser, PolicyDocumentLoader).
  - Mock NATS, MongoDB, Hedera SDK (already done in some existing tests via sinon).
  - Assert the block's runAction/getData/setData against representative inputs.
  - Skip blocks that are pure orchestration (interfaceContainerBlock, etc.) — those are integration territory.

  Prioritize blocks that mutate persistent state:
  - mintBlock, mintDocumentBlock (touches Hedera)
  - tokenActionBlock, tokenConfirmationBlock
  - requestVcDocumentBlock (the entry-point block in most policies)
  - sendToGuardianBlock (cross-service write)
  - externalDataBlock (the DMRV ingestion path)
  - customLogicBlock, calculateBlock (formula evaluation)
  - reassigningBlock (document ownership transfer)
  - retirementDocumentBlock (token destruction)

  3. Policy-engine service tests. guardian-service/src/policy-engine/policy-engine.service.ts is currently
  undercovered. Target methods:
  - clonePolicyAsync — used heavily, no unit test
  - draft — state machine that returned 500 unexpectedly during demo prep (PUBLISH → DRAFT blocked, but error
  message was generic)
  - publishPolicy — the pre-publish validation chain
  - buildDocumentationUrls 

  4. Gateway resolution tests. api-gateway/src/api/service/dmrv.ts is the entry point for the entire top-level
  API. The PR controller logic (regex enforcement, alias → entry resolution, method matching including
  Both) should be unit-tested against a mocked PolicyEngine.getPolicy(). Approx 10 tests covering: valid nested
  alias, valid flat alias, uppercase rejection, double-slash rejection, missing entry, method mismatch,
  Both-as-GET, Both-as-POST, missing policy, missing token.

  5. Coverage instrumentation and CI gate. The directory already has run-coverage.mjs and coverage-report.mjs.
  Wire these into the CI pipeline (.github/workflows/build_mgs.yml) with a coverage-drop gate: PRs that decrease
   policy-engine coverage block on review. Set the threshold high enough to push improvement but not so high it
  becomes a permanent blocker.

mariana-andruk-qa:
## Summary

Adds unit and contract test coverage for `policy-service`, plus a couple of related tests in `guardian-service` and `api-gateway`.

- **policy-service**: unit tests for block validators, blocks, helpers, decorators, errors, and policy-engine utilities, plus V8/c8 coverage tooling (`run-coverage.mjs`, `coverage-report.mjs`, `COVERAGE_NOTES.md`) and a `block-about` contract test.
- **guardian-service**: new `build-documentation-urls` helper with accompanying unit tests.
- **api-gateway**: `dmrv` service test (alias resolution) using `esmock`.

## Test results

Run against freshly built `dist/` output:

| Service | Passing | Failing |
|---|---|---|
| api-gateway | 12 | 0 |
| guardian-service | 16 | 0 |
| policy-service | 1254 | 0 |

The policy-service tests that were previously failing are now fixed:

- **ORM-dependent validators** no longer need a live database. `schema-validator` (and its property test) mock the ESM module via `esmock.strict`, and `orchestrator-validators` neutralises `DatabaseServer` reads on the real singleton instead of the non-working module mock. A `.mocharc.json` registers the esmock loader before mocha starts so the mock survives mocha's re-spawn.
- **`PolicyValidator` argument mismatch**: dropped the stray leading id argument so `policy`/`isDryRun` line up with the constructor signature.
- **`CommonBlock`-delegating validators** with mandatory options (`TransformationUIAddon`, `HttpRequestUIAddon`) now use per-block fixtures that supply the required fields.
- **`PolicyUtils` expectations** match the implementation: `getDocumentType` returns the `DocumentType` enum values (`VerifiablePresentation`/`VerifiableCredential`/`DID`) and `DocumentSignature.NEW` is numeric `0`.

## Notes

- `yarn.lock` changes were intentionally excluded (unrelated dependency churn).

## Related issues
Closes #6102

### DIFF
--- a/api-gateway/tests/service/dmrv.test.mjs
+++ b/api-gateway/tests/service/dmrv.test.mjs
@@ -1,0 +1,184 @@
+import assert from 'node:assert/strict';
+import esmock from 'esmock';
+
+const DMRV_DIST = '../../dist/api/service/dmrv.js';
+
+let engineStub;
+let invalidatedKeys;
+
+class MockPolicyEngine {
+    constructor(tenantContext) {
+        this.tenantContext = tenantContext;
+    }
+    getPolicy(...args) { return engineStub.getPolicy(...args); }
+    setBlockDataByTag(...args) { return engineStub.setBlockDataByTag(...args); }
+    getBlockDataByTag(...args) { return engineStub.getBlockDataByTag(...args); }
+}
+
+const helpersMock = {
+    PolicyEngine: MockPolicyEngine,
+    EntityOwner: class { constructor(user) { this.user = user; } },
+    getCacheKey: (tags) => `cache:${tags.join('|')}`,
+    InternalException: async (error) => { throw error; },
+    CacheService: class {},
+};
+
+async function loadDmrv() {
+    return await esmock(DMRV_DIST, {
+        '#helpers': helpersMock,
+        '#auth': { Auth: () => () => undefined, AuthUser: () => () => undefined },
+        '#constants': { PREFIXES: { POLICIES: 'policies/' } },
+        '#middlewares': { InternalServerErrorDTO: class {} },
+        '@guardian/common': { PinoLogger: class {} },
+    });
+}
+
+function makeUser() {
+    return { id: 'user-1', did: 'did:hedera:user', tenantContext: { tenantId: 't1' } };
+}
+
+function makeReq({ alias, method = 'GET', url = '/api/v1/dmrv/p1/x' }) {
+    return { params: { '*': alias }, method, url };
+}
+
+function makeApi(DmrvApi) {
+    invalidatedKeys = [];
+    const cacheService = { invalidate: async (key) => { invalidatedKeys.push(key); } };
+    const logger = { error: () => undefined };
+    return new DmrvApi(cacheService, logger);
+}
+
+describe('DmrvApi.proxyByAlias — alias resolution', function () {
+    this.timeout(60000);
+
+    let DmrvApi;
+
+    before(async () => {
+        ({ DmrvApi } = await loadDmrv());
+    });
+
+    beforeEach(() => {
+        engineStub = {
+            getPolicy: async () => ({
+                policyDocumentation: [
+                    { alias: 'monitoring-reports/create', method: 'POST', target: 'tag_create' },
+                    { alias: 'monitoring-reports', method: 'GET', target: 'tag_list' },
+                    { alias: 'status', method: 'Both', target: 'tag_status' },
+                ],
+            }),
+            getBlockDataByTag: async (user, policyId, tag) => ({ ok: 'get', tag }),
+            setBlockDataByTag: async (user, policyId, tag) => ({ ok: 'set', tag }),
+        };
+    });
+
+    it('resolves a valid nested alias on a GET and forwards to getBlockDataByTag', async () => {
+        const api = makeApi(DmrvApi);
+        const res = await api.proxyByAlias(
+            makeUser(), 'p1', {}, undefined,
+            makeReq({ alias: 'monitoring-reports', method: 'GET' })
+        );
+        assert.deepEqual(res, { ok: 'get', tag: 'tag_list' });
+    });
+
+    it('resolves a valid flat alias', async () => {
+        engineStub.getPolicy = async () => ({
+            policyDocumentation: [{ alias: 'create', method: 'GET', target: 'tag_flat' }],
+        });
+        const api = makeApi(DmrvApi);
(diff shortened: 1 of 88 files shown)


dariy-rised:
1. `orchestrator-validators.test.mjs` in `@unit PolicyValidator`
Six tests call the constructor with a string 't-1' as the first argument.
But the real constructor signature is:                                                                                                                                                                             
`constructor(policy: Policy, isDruRun: boolean = false, ...)` → 2 tests fail

Fix: remove 't-1'

2. `orchestrator-validators.test.mjs` and `schema-validator.test.mjs`: Module._load doesn't intercept ESM imports → ORM hits real DB → 3 tests fail

Fix: replace the Module._load block with esmock

3. `policy-utils.test.mjs`: getDocumentType returns enum strings, not 'vp-document', 'vc-document', 'did-document'.
→ 3 tests fail

Fix: update all three expected values to match the actual enum 'VerifiablePresentation', 'VerifiableCredential', 'DID'

4. `policy-utils.test.mjs`: DocumentSignature.NEW is numeric 0, not string 'NEW' → 2 tests fail

real enum is:
```
 export enum DocumentSignature {                                                                                                                                                                                
      NEW,      // = 0                                                                                                                                                                                         
      VERIFIED, // = 1
      INVALID,  // = 2
  }
```

Fix: replace 'NEW' to 0

5. `common-only-validators-batch3.test.mjs`: two validators require mandatory fields that the shared refOK fixture omits → 2 tests fail

TransformationUIAddon: requires options.expression
HttpRequestUIAddon: requires options.url and options.method

Fix: use per-block fixtures for the validators that have mandatory fields

6. also I have 13th test fail in `http-request-block.test.mjs`: IPv6 bracket handling bug
This is the code bug and we will fix it soon.

dariy-rised:
1) **orchestrator-validators.test.mjs**

 This new stubs persist for the whole Mocha process, other test files silently inherit them and may get false passes. can you add before/after scopes the stubs for this case.

from:
```
    DatabaseServer.getSchemas  = async () => [];
    DatabaseServer.getModules  = async () => [];
    DatabaseServer.getTools    = async () => [];
    DatabaseServer.getTool     = async () => null;
    DatabaseServer.getArtifact = async () => null;
```

to:
```
let origGetSchemas, origGetModules, origGetTools, origGetTool, origGetArtifact;

before(() => {
    origGetSchemas  = DatabaseServer.getSchemas;
    origGetModules  = DatabaseServer.getModules;
    origGetTools    = DatabaseServer.getTools;
    origGetTool     = DatabaseServer.getTool;
    origGetArtifact = DatabaseServer.getArtifact;

    DatabaseServer.getSchemas  = async () => [];
    DatabaseServer.getModules  = async () => [];
    DatabaseServer.getTools    = async () => [];
    DatabaseServer.getTool     = async () => null;
    DatabaseServer.getArtifact = async () => null;
});

after(() => {
    DatabaseServer.getSchemas  = origGetSchemas;
    DatabaseServer.getModules  = origGetModules;
    DatabaseServer.getTools    = origGetTools;
    DatabaseServer.getTool     = origGetTool;
    DatabaseServer.getArtifact = origGetArtifact;
});
```

2) **common-only-validators-batch3.test.mjs**

minor: extract the inline option objects into named like refOK, keeps cases clean.

  ```
  const refOK = () => ({ options: {} });
  const refExpression = () => ({ options: { expression: '1 + 1' } });
  const refHttpRequest = () => ({ options: { url: 'https://example.com/api', method: 'get' } });

  const cases = [
      ['dataTransformationAddon',      DataTransformationAddon,      refOK],
      ['ipfsTransformationUIAddon',    IpfsTransformationUIAddon,    refOK],
      ['transformationUIAddon',        TransformationUIAddon,        refExpression],
      ['httpRequestUIAddon',           HttpRequestUIAddon,           refHttpRequest],
  ];
```

dariy-rised (approved):

Issue #5584, opened 2025-12-22
Improve Excel schema creation flow to reduce repetitive and time-consuming actions


ghost:
Creating schemas via Excel templates currently requires a large number of repetitive and time-consuming actions, especially when working with sub-schemas and enums.

As schema complexity grows, this approach becomes difficult to maintain and error-prone, particularly due to the increasing number of required tabs within a single workbook.

### Current challenges
**Sub-schemas**
- When creating a sub-schema inside a VC (main schema):
-A separate tab must be added for each sub-schema
-The sub-schema tab must repeat much of the same information already present in the VC tab

_This leads to:_
- Duplication of data
- Higher risk of inconsistencies
- Reduced readability of the workbook

**Enums**
- For each enum field:
-A separate tab must be created

_In larger schemas, this results in:_
- Dozens of enum tabs
- Difficult navigation within the workbook
- Increased maintenance effort

### Proposed improvements

**Sub-schema handling**
Instead of requiring a separate tab for each sub-schema:
The system could:
- Detect sub-schemas based on Field type = "Sub-Schema" in the main schema (VC) tab
- Take the sub-schema name from the Description column
- Automatically include all rows grouped below that belong to the sub-schema

_This would allow sub-schemas to be defined directly within the main schema tab, reducing duplication and complexity_

**Enum handling**
Instead of creating a separate tab for each enum:
- Use a single shared Enum tab with a table containing:
-Schema name
-Field name
-Enum parameter/value

- On the schema tab:
Fields would reference the enum definition from this shared Enum tab

- While schema name and field name would still need to be specified per enum value, this approach would:
-Greatly simplify workbook structure
-Improve navigation
-Reduce the number of required tabs
-Make enum management more scalable

### Benefits

- Reduced repetitive manual work
- Cleaner and more maintainable Excel templates
- Better scalability for complex schemas
- Improved user experience and lower error rate during schema creation

PR #6126, opened 2026-05-27
feat: 5584 Improve Excel schema creation flow

dariy-rised:
### Description
- Added parsing inline sub-schemas (previously required a separate tab per sub-schema)
- Added a shared "Enums" tab replacing per-field enum tabs
- Added error messages in the import modal window

### Related issue(s)
Resolves: [#5584](https://github.com/hashgraph/guardian/issues/5584)


common/src/xlsx/json-to-xlsx.ts
```diff
@@ -38,7 +57,7 @@ export class JsonToXlsx {
             _schemaCache.set(schema.iri, sheetName);
         }
```
Pyatakov, line 51: These could be folded into a single pass to avoid re-parsing each schema 3x.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6187, opened 2026-06-10
fix(policy-service): stop documents-source from over-fetching DocumentState.document

vshvets-bc:
## Problem

`documents-source` history loop calls `getDocumentStates(filter)` per outer document — a plain MikroORM `find()` with no projection. For methodologies whose `DocumentState.document` field carries large embedded payloads (e.g. VM0033 Tidal Wetland & Seagrass: ~1-2 MB per row of geospatial polygons + monitoring data), every hydrated entity is pinned by the per-policy EntityManager's IdentityMap until the child process exits. Only ~250 bytes of each row are actually read (`createDate`, `document.option.status`, `document.option.comment`).

The result: policy-service child processes running such methodologies retain tens to hundreds of MB of `document` field data unnecessarily.

## Fix

Add `DatabaseServer.getDocumentStateHistory(filters, documentSubpaths)` that runs a `$match + $project` aggregation. The aggregation:
- Projects only `createDate` plus the top-level `document.<key>` subtrees derived from the dot-paths the caller will read (defaults: `document.option`)
- Returns plain objects via MikroORM's aggregation path, so nothing lands in the IdentityMap
- Top-level key names are validated with `/^[A-Za-z_][A-Za-z0-9_]*$/` before being injected into the projection

`documents-source.ts` history loop now calls the new method with the two dot-paths actually read. Per-row payload drops from ~2 MB → ~300 bytes.

## Changes
- `common/src/database-modules/database-server.ts` — `+getDocumentStateHistory(...)`
- `policy-service/src/policy-engine/blocks/documents-source.ts` — use the new method

## Compatibility
- Behavior unchanged: same `labelValue` / `comment` / `created` returned per history entry.
- Dry-run routing preserved (`aggregate` already branches on `this.dryRun`).
- Old `getDocumentStates` left in place; this is purely additive.

### DIFF
--- a/common/src/database-modules/database-server.ts
+++ b/common/src/database-modules/database-server.ts
@@ -2688,15 +2688,16 @@ export class DatabaseServer extends AbstractDatabaseServer {
         filters: FilterObject<DocumentState>,
         documentSubpaths: string[],
     ): Promise<Array<{ createDate: Date, document: any }>> {
-        const topLevelKeys = Array.from(new Set(
-            documentSubpaths
-                .map(p => (p || '').split('.', 1)[0])
-                .filter(k => /^[A-Za-z_][A-Za-z0-9_]*$/.test(k))
-        ));
-        const projection: any = { _id: 0, createDate: 1 };
-        for (const k of topLevelKeys) {
-            projection[`document.${k}`] = 1;
+    const topLevelKeys = Array.from(new Set(
+        documentSubpaths.map(p => (p || '').split('.', 1)[0])
+    ));
+    const projection: any = { _id: 0, createDate: 1 };
+    for (const key of topLevelKeys) {
+        if (!key || key.startsWith('$') || key.includes('.') || key.includes('\0')) {
+            continue;
         }
+        projection[`document.${key}`] = 1;
+    }
         const pipeline: any[] = [
             { $match: filters },
             { $project: projection },
--- a/common/src/database-modules/database-server.ts
+++ b/common/src/database-modules/database-server.ts
@@ -2685,6 +2685,35 @@ export class DatabaseServer extends AbstractDatabaseServer {
     }
 
     /**
+     * Lightweight history projection for `documents-source` blocks.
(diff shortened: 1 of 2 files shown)


common/src/database-modules/database-server.ts
```diff
@@ -2684,6 +2684,26 @@ export class DatabaseServer extends AbstractDatabaseServer {
...
+    ): Promise<Array<{ createDate: Date, document: any }>> {
+        const topLevelKeys = Array.from(new Set(
+            documentSubpaths
+                .map(p => (p || '').split('.', 1)[0])
+                .filter(k => /^[A-Za-z_][A-Za-z0-9_]*$/.test(k))
+        ));
+        const projection: any = { _id: 0, createDate: 1 };
+        for (const k of topLevelKeys) {
+            projection[`document.${k}`] = 1;
+        }
```
Pyatakov: The `/^[A-Za-z_][A-Za-z0-9_]*$/` allow-list is stricter than what MongoDB permits for field names. A methodology that configures `timelineLabelPath/timelineDescriptionPath` with a top-level key containing a hyphen or leading digit (e.g. `field-1.status`) has that key silently dropped from the projection, `document.<key>` is then absent and `labelValue/comment` come back empty, where the old `find()` path returned the real value. Suggested swapping the allow-list for a deny-list that only blocks `$`-prefixed, dotted, empty, or null-byte keys (the only things unsafe in `document.${key}`). This keeps the injection guard while avoiding the silent data loss.
```suggestion
    const topLevelKeys = Array.from(new Set(
        documentSubpaths.map(p => (p || '').split('.', 1)[0])
    ));
    const projection: any = { _id: 0, createDate: 1 };
    for (const key of topLevelKeys) {
        if (!key || key.startsWith('$') || key.includes('.') || key.includes('\0')) {
            continue;
        }
        projection[`document.${key}`] = 1;
    }
```

Pyatakov (changes requested):

Pyatakov (approved):

PR #6283, opened 2026-06-26
feat: 6312 Add source conditions to validation addon block

dariy-rised:
### Description
- Added sourceValidations to documentValidatorBlock for cross-document validation against stored policy documents
- Added custom config component for documentValidatorBlock replacing the generic property editor
- Added cross-document condition evaluation supporting equal, not_equal, in, not_in, gt, gte, lt, lte operators
- Added source filter support with value (literal) and variable (resolved from input document) value types
- Added documentValidatorBlock as allowed child of requestVcDocumentBlockAddon
- Added unit tests for block validator

### Related issue(s)
Resolves: [#6312](https://github.com/hashgraph/guardian/issues/6312)


policy-service/src/policy-engine/helpers/utils.ts
```diff
@@ -313,7 +313,7 @@ export class PolicyUtils {
      */
     public static getSchemaContext(ref: AnyBlockType, schema: SchemaCollection): string {
         if (ref.dryRun) {
-            return `schema${schema.iri}`;
+            return `schema:${schema.iri.slice(1)}`;
```
Pyatakov, line 316: Why is this change to dry-run schema context? getSchemaContext is also called by `request-vc-document-block`, `request-vc-document-block-addon`, `split-block`, and `reassigning.block`, so this alters context generation for all of them in dry-run, not just the validator, and there's no test covering it. If it's required for source validation in dry-run, please explain it in the description and add a test; if it's an incidental fix, perhaps it should be split into its own commit (or even PR).

policy-service/src/policy-engine/helpers/utils.ts
```diff
@@ -420,6 +420,20 @@ export class PolicyUtils {
         return true;
     }
 
+    /**
+     * Coerce a value to the best comparable type: number → ISO Date → string.
+     * Strings that don't start with YYYY-MM-DD are never passed to new Date(),
+     * so free-form strings like "Jan 1 2024" are left as-is.
+     */
+    private static coerceComparable(v: any): any {
```
Pyatakov: **The doc comment describes a guard that doesn't exist**. Lines 425-426 claim "_Strings that don't start with YYYY-MM-DD are never passed to `new Date()`, so free-form strings like 'Jan 1 2024' are left as-is._", but the implementation passes every non-numeric string to `new Date(v)` with no such check. So "`Jan 1 2024`", "`March`", etc. get coerced to Dates, and `gt/gte/lt/lte` on free-form text fields behave inconsistently (Date-parseable strings compare as timestamps; others fall back to lexicographic). Either add the ISO-prefix guard the comment promises, or fix the comment to match the actual behavior.

policy-service/src/policy-engine/blocks/document-validator-block.ts
```diff
@@ -34,102 +34,228 @@ import { LocationType } from '@guardian/interfaces';
...
-        },
-        ]
+        defaultEvent: true
     },
     variables: [
         { path: 'options.schema', alias: 'schema', type: 'Schema' }
     ]
 })
 export class DocumentValidatorBlock {
+    private coerceValue(value: any): any {
```
Pyatakov, line 44: Two different coercion implementations. `coerceValue` here also maps `'null'/'true'/'false'` and returns `Date.getTime()` (a number, line 52), while `PolicyUtils.coerceComparable` (`policy-service/src/policy-engine/helpers/utils.ts:428`) returns a `Date` object and has no string -> boolean/null handling. Same-doc conditions (`checkDocumentField` -> `coerceComparable`) and cross-doc conditions (`evaluateCrossCondition` -> `coerceValue`) therefore coerce identical inputs differently for the same operators. Please consolidate into a single shared helper so both paths behave consistently.

frontend/src/app/services/handle-errors.service.ts
```diff
@@ -109,8 +109,9 @@ export class HandleErrorsService implements HttpInterceptor {
                 !this.excludeErrorCodes.includes(String(error.status)) &&
                 !this.excludeErrorTexts.includes(String(result.text))
             ) {
+                const displayText = result.text.replace(/\n/g, '<br>');
```
Pyatakov: `displayText` is built by replacing `\n` with `<br>` and then interpolated unescaped into the toastr body as `<div>${displayText}</div>`. `result.text` is server-supplied and now carries the new multi-line source-validation messages, so any HTML/script in it renders in the toast. Pre-existing, but this PR widens what flows through it, please HTML-escape result.text before the `<br>` substitution.

Pyatakov (changes requested):

policy-service/src/policy-engine/helpers/utils.ts
dariy-rised: The change fixes a bug in `checkDocumentSchema`: dry-run VCs write `schema:${subject.type}` into `credentialSubject['@context'] (vcjs.ts:447)`, but the old code returned `schema#abc (schema${schema.iri}` missing the colon, # not stripped), so indexOf always missed and schema validation silently failed in dry-run.

  The other callers (`request-vc, split-block, reassigning, request-vc-addon`) pass the result as `evidenceContext` into `vc.addContext()`. DraftSchemaContextLoader resolves it via `'#' + iri.substring(7)` both schema# and schema: are 7 characters, so both formats produce the same DB lookup. Those callers are unaffected by the format change.

The existing test asserting schema#abc was validating the buggy output and will be updated to schema:abc.

Pyatakov (approved):

PR #6328, opened 2026-07-03
fix: dry run active virtual user per session

### LINKED ISSUE #6327: Concurrent dry-run testers share one global "active virtual user" per policy, causing 422 "Block Unavailable" and identity swaps on refresh
## Summary

When two or more real users open the **same policy** in dry-run mode at the same time, switching the active virtual user (VU) as one user silently reassigns the acting identity for **every other user** viewing that policy. Form submissions then fail with `422 Unprocessable Entity` ("Block Unavailable"), and a hard refresh can land a user on a completely different virtual persona than the one they were using.

The dry-run "active virtual user" is a single pointer per `policyId`, shared globally across every real user/session/tab currently testing that policy. Whoever calls the persona switch last wins — for everyone.

## Steps to Reproduce

1. Two real users open the **same policy** in dry-run mode.
2. User A switches their active dry-run persona to `Virtual User 2` via the persona switcher.
3. User B (who already had the policy/form open) submits a document (e.g. clicks the create/validate button on a form block).
4. The request `POST /api/v1/policies/{policyId}/blocks/{uuid}` returns `422 Unprocessable Entity` with message `"Block Unavailable"`. The `sync-events` channel also shows failures.
5. User B does a hard refresh. Instead of returning to their own working state, they land on **User A's** last-selected persona (`Virtual User 2`).

## Expected Behavior

Each real user should be able to independently select and drive their own virtual persona within a shared dry-run policy, without affecting what other concurrent users see or are authenticated as.

## Actual Behavior

There is exactly **one** "active virtual user" per `policyId`, shared globally across every real user/session/tab currently testing that policy's dry run. The last persona switch overwrites the acting identity for all concurrent testers.

## Root Cause

The "who am I acting as" resolution for dry-run mode is keyed **only by `policyId`**, with no per-real-user scoping.

- `common/src/database-modules/database-server.ts`
  - `static setVirtualUser(policyId, did)` (~L5111): iterates **all** `VirtualUsers` rows for the policy and sets `active = (item.did === did)` — a single shared pointer:
    ```ts
    public static async setVirtualUser(policyId: string, did: string): Promise<void> {
        const items = (await new DataBaseHelper(DryRun).find({
            dryRunId: policyId,
            dryRunClass: 'VirtualUsers'
        }));
        for (const item of items) {
            item.active = item.did === did;
            await new DataBaseHelper(DryRun).save(item);
        }
    }
    ```
  - `static getVirtualUser(policyId)` (~L4109): reads that single `active: true` pointer back, filtered only by `dryRunId: policyId` + `dryRunClass: 'VirtualUsers'`.
- `guardian-service/src/policy-engine/policy-components-utils.ts`
  - `GetUserRoleByPolicy` (~L62) and `GetUserRole` (~L129) both override the caller's identity with the single shared pointer when the policy is in dry-run mode:
    ```ts
    if (PolicyHelper.isDryRunMode(policy)) {
        const activeUser = await DatabaseServer.getVirtualUser(policyId);
        if (activeUser && did !== activeUser.did) {
            did = activeUser.did;
            permissions = [];
        }
    }
    ```
    This is the path hit on page load / refresh (`policy()` info), which is why a hard refresh moves the user onto whichever persona was selected last.
- `policy-service/src/policy-engine/policy-components-utils.ts`
  - `GetPolicyUserByName` (~L1344) and `GetActiveVirtualUser` (~L1440) resolve the acting identity via `DatabaseServer.getVirtualUser(instance.policyId)`, again with no real user id in the lookup. The `422 "Block Unavailable"` is the block engine correctly refusing to serve a block to a role that doesn't match the now-swapped active identity — a symptom, not the bug.
- `api-gateway/src/api/service/policy.ts`
  - `POST /:policyId/dry-run/login` (~L6195) calls `loginVirtualUser` (~L6259), which calls the shared `setVirtualUser` above.

**Design note:** the dry-run feature appears to have been designed for a single tester driving multiple simulated actors sequentially, not for concurrent multi-user testing of the same policy. As collaborative dry-run testing becomes common, the single-active-user-per-policy model breaks down.

## Proposed Fix

Scope the "active persona" pointer to `(policyId, real user id)` instead of `policyId` alone, rather than duplicating the entire dry-run sandbox per user. Document/group/block state is already keyed by persona `did`, so per-user scoping of the active pointer is sufficient to isolate concurrent testers. No schema migration is required: the `DryRun` entity (`common/src/entity/dry-run.ts`) already has an unused `userId` field.

**Touch points:**
- `common/src/database-modules/database-server.ts`: `getVirtualUser`, `setVirtualUser`, `getVirtualUserByAccount`, `getVirtualUsers`, `countVirtualUsers` — add/filter by real `userId`.
- `policy-service/src/policy-engine/policy-components-utils.ts`: `GetPolicyUserByName`, `GetActiveVirtualUser` — thread the real user id through (already available in the call chain).
- `guardian-service/src/policy-engine/policy-components-utils.ts`: `GetUserRoleByPolicy` / `GetUserRole` (the role-resolution path hit on page load / refresh) need the same scoping.
- `api-gateway/src/api/service/policy.ts`: `/dry-run/login`, `/dry-run/user` — pass `user.id` down. Verify cache invalidation (`invalidedCacheTags` for `.../navigation`, `.../groups`) is scoped per real user so cached responses do not leak between concurrent testers.
- `frontend/src/app/modules/policy-engine/policy-viewer/policy-viewer/policy-viewer.component.ts`: `loadPolicyById` re-fetches `policy()` on every load/refresh and renders the server-resolved `userRole`; it will follow correct per-user scoping once the backend pointer is fixed.

**Estimated effort:** Medium. Self-contained backend change across `common`, `policy-service`, `guardian-service`, and `api-gateway`, with no DB migration or backfill. Main cost is careful testing of the dry-run flow across services (login, block execution, refresh, `sync-events`).

## Out of Scope / Follow-up

- **Two testers selecting the same virtual persona simultaneously** will still collide after this fix. Needs a product decision (block it, or accept as shared/intentional).
- **"Restart dry run"** currently wipes all virtual users for the entire `policyId`, affecting every concurrent tester. This remains a shared destructive action regardless of the fix above and may warrant its own guard/warning.

## References

- Upstream file: https://github.com/hashgraph/guardian/blob/main/guardian-service/src/policy-engine/policy-components-utils.ts

nazlo-climission:
**Description**:

This PR scopes the active dry-run virtual user (persona) per real user instead of per policy, in order to support concurrent dry-run testing by multiple users on the same policy without them interfering with each other's selected persona.

* Add optional `userId` parameter to `DatabaseServer.getVirtualUser`/`setVirtualUser`
* When `userId` is provided, read/write a new per-real-user pointer (`dryRunClass: 'ActiveVirtualUser'`, keyed by `dryRunId` + `userId`) instead of the single shared `active` flag on `VirtualUsers`
* Fall back to the previous shared-flag behavior when `userId` is omitted (e.g. background jobs)
* Pass `user.id`/`owner?.id`/`userId` through call sites: `PolicyComponentsUtils` (guardian-service and policy-service), `PolicyEngineService` (`SET_VIRTUAL_USER`, `RESTART_DRY_RUN`, `START_POLICY_TEST` handlers), and `BasicBlock.allAvailableUsers`
* Add unit tests for `DatabaseServer.getVirtualUser`/`setVirtualUser` covering both the new per-user pointer path and the legacy shared-flag fallback

**Related issue(s)**:

Fixes #6327 

**Notes for reviewer**:

Root cause: `getVirtualUser`/`setVirtualUser` resolved the "currently active" dry-run persona keyed only by `policyId`, via a single shared `active` flag. When two or more real users opened the same policy in dry-run mode at the same time, switching persona as one user silently reassigned the acting identity for every other user viewing that policy — causing block submissions to fail with `422 Unprocessable Entity` ("Block Unavailable"), and a hard refresh could land a user on a completely different persona than the one they were using.

No schema migration needed: the `DryRun` entity already had an unused `userId` field.

Out of scope (flagging for discussion, not fixed in this PR):
* Two real users deliberately selecting the *same* persona will still share it (this reads as a product decision, not a bug).
* "Restart dry run" still clears all personas for the whole policy, affecting every concurrent tester (pre-existing, unrelated behavior).

Added unit tests in `common/tests/unit-tests/database-server/database-server.test.mjs` (new `Active virtual user pointer (dry-run)` describe block) covering:
* `getVirtualUser`/`setVirtualUser` fall back to the legacy shared `active` flag when `userId` is omitted
* `getVirtualUser` resolves the per-user pointer then the matching persona, returns `null` when no pointer exists or the pointer has no `did`, and keeps two real users on independently selected personas
* `setVirtualUser` updates an existing per-user pointer in place, creates one when none exists, and doesn't let one real user's persona switch affect another's

Verified these tests fail against the pre-fix code (6 of 9 fail without the `userId` branch) and pass with it, so they're a genuine regression guard and not vacuous. No integration test was added for two concurrent dry-run sessions end-to-end — flagging in case reviewers want that as a follow-up.

**Checklist**

### DIFF
--- a/common/src/database-modules/database-server.ts
+++ b/common/src/database-modules/database-server.ts
@@ -4134,6 +4134,13 @@ export class DatabaseServer extends AbstractDatabaseServer {
      * Get All Virtual Users
      * @param policyId
      * @param savepointIds
+     * @param virtualKey
+     * @param document
+     * @param userId Real (authenticated) user requesting the list. When
+     * provided, each returned user's `active` field is recomputed against
+     * that real user's own selected persona (`ActiveVirtualUser` pointer)
+     * instead of the legacy shared `active` column, so concurrent dry-run
+     * testers each see their own persona highlighted.
      * @virtual
      */
     public static async getVirtualUsers(
@@ -4141,6 +4148,7 @@ export class DatabaseServer extends AbstractDatabaseServer {
         savepointIds?: string[],
         virtualKey?: boolean,
         document?: boolean,
+        userId?: string | null,
     ): Promise<DryRun[]> {
         const filter: any = {
             dryRunId: policyId,
@@ -4167,6 +4175,17 @@ export class DatabaseServer extends AbstractDatabaseServer {
                 createDate: 1
             }
         });
+
+        if (userId) {
+            const pointer = await new DataBaseHelper(DryRun).findOne({
+                dryRunId: policyId,
+                dryRunClass: 'ActiveVirtualUser',
+                userId
+            });
+            for (const user of users) {
+                user.active = !!pointer?.did && pointer.did === user.did;
+            }
+        }
 
         if (virtualKey) {
             for (const user of users) {
--- a/common/src/database-modules/database-server.ts
+++ b/common/src/database-modules/database-server.ts
@@ -4085,7 +4085,36 @@ export class DatabaseServer extends AbstractDatabaseServer {
      *
      * @virtual
      */
-    public static async getVirtualUser(policyId: string): Promise<DryRun | null> {
+    public static async getVirtualUser(policyId: string, userId?: string | null): Promise<DryRun | null> {
+        if (userId) {
+            const pointer = await new DataBaseHelper(DryRun).findOne({
+                dryRunId: policyId,
+                dryRunClass: 'ActiveVirtualUser',
+                userId
+            });
+            if (!pointer?.did) {
+                return null;
+            }
+            return await new DataBaseHelper(DryRun).findOne({
+                dryRunId: policyId,
+                dryRunClass: 'VirtualUsers',
+                did: pointer.did
+            }, {
+                fields: [
+                    'id',
+                    'did',
+                    'username',
+                    'hederaAccountId',
+                    'active'
+                ]
+            } as unknown as FindOptions<object>);
+        }
+        // Legacy fallback: single shared active pointer. Used only when no
+        // real (authenticated) user context is available, e.g. background
+        // jobs. Callers resolving a request on behalf of a real user should
+        // always pass `userId`, so that each real user driving a dry-run
+        // policy keeps their own independently selected virtual persona
+        // instead of sharing one global pointer with every other tester.
         return await new DataBaseHelper(DryRun).findOne({
             dryRunId: policyId,
             dryRunClass: 'VirtualUsers',
@@ -5087,7 +5116,25 @@ export class DatabaseServer extends AbstractDatabaseServer {
      *
      * @virtual
      */
-    public static async setVirtualUser(policyId: string, did: string): Promise<void> {
+    public static async setVirtualUser(policyId: string, did: string, userId?: string | null): Promise<void> {
+        if (userId) {
+            const pointer = await new DataBaseHelper(DryRun).findOne({
+                dryRunId: policyId,
+                dryRunClass: 'ActiveVirtualUser',
+                userId
+            });
+            if (pointer) {
+                pointer.did = did;
+                await new DataBaseHelper(DryRun).save(pointer);
+            } else {
+                await new DataBaseHelper(DryRun).save(DatabaseServer.addDryRunId({
+                    did,
+                    userId
+                }, policyId, 'ActiveVirtualUser', false));
+            }
+            return;
+        }
+        // Legacy fallback: see getVirtualUser() above.
         const items = (await new DataBaseHelper(DryRun).find({
             dryRunId: policyId,
             dryRunClass: 'VirtualUsers'
--- a/common/tests/unit-tests/database-server/database-server.test.mjs
+++ b/common/tests/unit-tests/database-server/database-server.test.mjs
@@ -415,5 +415,66 @@ describe('DatabaseServer', function () {
 				assert.equal(pointerB.did, 'did:b');
 			});
 		});
+
+		describe('getVirtualUsers', function () {
+			it('recomputes active per real user instead of trusting the stored shared flag when userId is given', async function () {
+				const users = [
+					{did: 'did:a', active: true},
+					{did: 'did:b', active: false},
+				];
+				findStub.resolves(users);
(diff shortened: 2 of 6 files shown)


common/src/database-modules/database-server.ts
```diff
@@ -5107,7 +5116,25 @@ export class DatabaseServer extends AbstractDatabaseServer {
...
+            if (pointer) {
+                pointer.did = did;
+                await new DataBaseHelper(DryRun).save(pointer);
+            } else {
+                await new DataBaseHelper(DryRun).save(DatabaseServer.addDryRunId({
+                    did,
+                    userId
+                }, policyId, 'ActiveVirtualUser', false));
+            }
+            return;
```
Pyatakov, line 5154: The per-user branch here only writes the `ActiveVirtualUser` pointer and returns early, it never sets `active = true` on the `VirtualUsers` record the way the legacy branch below does. Since all three write call sites now pass a `userId` (`SET_VIRTUAL_USER`, `RESTART_DRY_RUN`, `START_POLICY_TEST`), the `active` flag is effectively never set to `true` anymore for a normal dry-run session.

The frontend persona switcher relies on that flag to indicate the selected persona: `getVirtualUsers` returns each record's `active` field, and `policy-viewer.component.html` binds `[attr.active]="item.active"` to highlight the current persona. The `GET_VIRTUAL_USERS` handler receives no `userId`, so it can't recompute per-user active state either.

As a result, after a user selects a persona, block submissions route correctly, but no entry in the switcher is highlighted, and it stays unhighlighted after a reload. The user loses the visual confirmation of which persona they're acting as.
The switcher needs a per-user notion of "selected".

Options: have `GET_VIRTUAL_USERS` accept `userId` and mark the active persona by comparing each `did` against the user's `ActiveVirtualUser` pointer, or return the active `did` alongside the list so the frontend can highlight it. The shared `active` column can't represent per-user selection.

Pyatakov (changes requested):

common/src/database-modules/database-server.ts
nazlo-climission: pls check the fix

Pyatakov (approved):

PR #6329, opened 2026-07-04
feat: 6239 first steps onboarding

### LINKED ISSUE #6239: "First Steps" for new users (simplified) in side panel / right side drawer
When a new user logs (standard registry or default user), a welcome page guides first steps to using the Guardian. This welcome page content is authored/managed in Gitbook and content pulled into the app.

Content in Drawer Component

* display the content in a right side 'drawer' component so that user can follow along
* can be toggled open and closed
* https://primeng.dev/drawer
    

Designated Pages

* First step pages will be stored in gitbook there is a dedicated page for 'standard registry' and 'default user'
* https://guardian.hedera.com/docs/develop/first-steps/first-steps-with-digital-environmental-assets-simplified
    

Acceptance Criteria

* drawer opens by default when 'First Steps' is enabled
* page context pulled from a designated file in the repo (e.g. md file or html)
* user can disable first steps from the drawer
* can be re-enabled from the profile page
* instructions are for first time users of the Guardian

ihar-tsykala:
**Description**:

**Related issue(s)**:

Resolve #6239

**Notes for reviewer**:

**Checklist**


frontend/src/app/services/first-steps.service.ts
```diff
@@ -0,0 +1,111 @@
...
+        return this.role !== null && this.enabled$.value && this.open$.value;
+    }
+
+    public setOpen(value: boolean): void {
+        if (this.role === null || value === this.open$.value) {
+            return;
+        }
+        this.open$.next(value);
+        this.persistOpen();
+        this.applyPanelClass();
```
Pyatakov: When First Steps is toggled on for the first time, `open$` is already at its default `true`, so `setEnabled(true)` -> `setOpen(true)` hits the `value === this.open$.value` early-return and `applyPanelClass()` never runs. The drawer becomes visible (`isOpen()` is true), but the `first-steps-open` class is never added to `documentElement`, so `.content { right: … }` doesn't apply and the drawer overlaps the page content instead of resizing it. It only corrects itself after a reload.

Root cause: the panel class depends on both `enabled$` and `open$`, but `applyPanelClass()` only runs when `open$` changes. Suggest recomputing it unconditionally in `setOpen`:
```suggestion
        if (this.role !== null && value !== this.open$.value) {
            this.open$.next(value);
            this.persistOpen();
        }
        this.applyPanelClass();
```

Pyatakov (changes requested):

frontend/src/app/views/user-profile/user-profile.component.html
```diff
@@ -42,108 +40,141 @@ <h1>Profile</h1>
...
+                                        <span class="config-row-desc">Show the First Steps tutorial toggle in the side menu</span>
+                                    </div>
+                                    <p-toggleswitch [ngModel]="firstStepsEnabled" (onChange)="onFirstStepsToggle($event.checked)"></p-toggleswitch>
+                                </div>
+                                <div class="config-row">
+                                    <div class="config-row-label">
+                                        <span class="config-row-title">Two-factor authentication</span>
+                                        <span class="config-row-desc">Require an authenticator code at sign-in</span>
+                                    </div>
+                                    <p-toggleswitch [ngModel]="is2faEnabled" (onChange)="onToggle2fa($event.checked)"></p-toggleswitch>
```
Pyatakov: Replace with two-way binding:
```suggestion
                                    <p-toggleswitch [(ngModel)]="is2faEnabled" (onChange)="onToggle2fa($event.checked)"></p-toggleswitch>
```
so we don't need the hack at `user-profile.component.ts:1280-1282` anymore.

Same needs to be applied to `root-profile.component.html:224` and `root-profile.component.ts:872-875`.

Pyatakov (approved):

PR #6340, opened 2026-07-06
fix(common): serialize migrations across replicas with a Mongo lock

vshvets-bc:
## Problem

Guardian services run migrations at boot inside `Promise.all([...])` in each service's `app.ts`. In a multi-replica deployment, every replica calls the shared `Migration()` helper (`common/src/helpers/migration.ts`) concurrently, so `migrator.up()` runs on all replicas at once.

Data-backfill migrations that insert rows via raw collection writes (`this.getCollection(...).insertOne(...)`) enforce no unique constraint at insert time. When two replicas run such a migration concurrently, both can insert duplicate rows for the same key before the corresponding unique index exists. The parallel `MikroORM.init({ ensureIndexes: true })` then fails to build the unique index on the now-duplicate data, and the pod crash-loops with `E11000 duplicate key`.

## Fix

Add a Mongo advisory lock held around the migration run inside the shared `Migration()` helper, so only one replica of a service runs `migrator.up()` at a time. The lock document lives in the service's own database, so it scopes naturally to replicas of a single service with no cross-service contention.

- **New** `common/src/helpers/migration-lock.ts` - `withMigrationLock(db, fn)`: atomic acquire via `findOneAndUpdate` (takes the lock only when it is free or its lease has expired), a `.unref()`ed heartbeat that extends the lease while migrations run, poll-until-free waiters (2s poll, 5min cap), and guaranteed release in `finally`. A holder that crashes without releasing frees the lock after the TTL (stale takeover), so startup cannot deadlock.
- **Edit** `common/src/helpers/migration.ts` - wraps the `getExecutedMigrations` + old-migrations loop + `migrator.up()` block in `withMigrationLock`, obtaining the native `Db` via `orm.em.getDriver().getConnection().getDb()`.

No signature change - every service that calls `Migration()` gets serialization automatically. Code-only; no infra/deployment change. Late acquirers run `migrator.up()` after the leader finishes and find nothing pending (no-op).

## Verification

- `tsc` typecheck passes on `common`.
- Two concurrent service starts against the same database: exactly one acquires the lock and runs migrations; the other waits, then acquires and finds nothing pending. No `E11000`.

### DIFF
--- a/common/src/helpers/migration.ts
+++ b/common/src/helpers/migration.ts
@@ -2,6 +2,7 @@ import { MikroORM } from '@mikro-orm/core';
 import { MongoDriver } from '@mikro-orm/mongodb';
 import { Migrator } from '@mikro-orm/migrations-mongodb';
 import process from 'node:process';
+import { withMigrationLock } from './migration-lock.js';
 
 const DEFAULT_MIN_POOL_SIZE = '1';
 const DEFAULT_MAX_POOL_SIZE = '5';
@@ -23,18 +24,23 @@ export async function Migration(initConfig: any, migrations?: string[]) {
       },
     });
 
-    const migrator = orm.getMigrator();
-    const executedMigrations = await migrator.getExecutedMigrations();
-    const executeOldMigrations = async (name: string) => {
-        if (!executedMigrations.some((migration) => migration.name === name)) {
-            await migrator.up(name);
+    // Serialize migrations across replicas so concurrent runs can't insert
+    // duplicate rows before a unique index exists.
+    const db = orm.em.getDriver().getConnection().getDb();
+    await withMigrationLock(db, async () => {
+        const migrator = orm.getMigrator();
+        const executedMigrations = await migrator.getExecutedMigrations();
+        const executeOldMigrations = async (name: string) => {
+            if (!executedMigrations.some((migration) => migration.name === name)) {
+                await migrator.up(name);
+            }
+        };
+        if (migrations) {
+            for (const name of migrations) {
+                await executeOldMigrations(name);
+            }
         }
-    };
-    if (migrations) {
-        for (const name of migrations) {
-            await executeOldMigrations(name);
-        }
-    }
-    await migrator.up();
+        await migrator.up();
+    });
     return orm;
 };
--- a/common/src/helpers/migration.ts
+++ b/common/src/helpers/migration.ts
@@ -16,6 +16,13 @@ const RADIX = 10;
 export async function Migration(initConfig: any, migrations?: string[]) {
     const orm = await MikroORM.init<MongoDriver>({
       ...initConfig,
+      // Never build indexes during init(): MikroORM.init() runs ensureIndexes()
+      // itself before returning, outside the migration lock and against
+      // pre-migration data, so a unique-index build races across replicas and
+      // throws E11000 on not-yet-deduped rows. Force it off here (overriding
+      // whatever the caller passed) and build the indexes inside the lock,
+      // after the backfill migrations have run.
+      ensureIndexes: false,
       extensions: [Migrator],
       driverOptions: {
           minPoolSize: parseInt(process.env.MIN_POOL_SIZE ?? DEFAULT_MIN_POOL_SIZE, RADIX),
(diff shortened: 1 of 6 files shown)


common/src/helpers/migration.ts
Pyatakov, line 17: **The `ensureIndexes` step this PR blames for the `E11000` crash runs outside the lock.**

Callers pass `ensureIndexes: true` into `initConfig`, e.g. `guardian-service/src/app.ts:87`:
```ts
ensureIndexes: true,
 ```
In `@mikro-orm/core`, `MikroORM.init()` performs the index build *itself, before returning*.
`node_modules/@mikro-orm/core/MikroORM.js:58-59`:
```js
if (orm.config.get('connect') && orm.config.get('ensureIndexes')) {
    await orm.getSchemaGenerator().ensureIndexes();
```
`connect` defaults to `true`, so line 59 runs during `await MikroORM.init(...)`. That `ensureIndexes()` is the unique-index build that throws `E11000` on duplicate data, `node_modules/@mikro-orm/mongodb/MongoSchemaGenerator.js:86` → `createUniqueIndexes()` (defined at line 175):
```js
promises.push(...this.createUniqueIndexes(meta));
```

In `migration.ts` the lock is only acquired *after* `init()` has already returned:
```ts
17      const orm = await MikroORM.init<MongoDriver>({    // ensureIndexes() runs here, unguarded
...
29      const db = orm.em.getDriver().getConnection().getDb();
30      await withMigrationLock(db, async () => {           // lock starts here — too late
31          const migrator = orm.getMigrator();
...
43          await migrator.up();
44      });
```
Two consequences:
1. **The blamed step is not serialized** — all replicas still run `ensureIndexes()` concurrently during `init()`, outside the lock.
2. **The ordering is backwards** — indexes are built (line 16) *before* the backfill migrations (line 43) run inside the lock, so the index build sees pre-migration data on every boot.

The lock does correctly serialize `migrator.up()` (a real concurrency win against double-inserts / migration-bookkeeping races), but that's a different problem from the E11000-on-unique-index path in the description.

**Suggested fix:** drop `ensureIndexes: true` from the init config at `guardian-service/src/app.ts:87` (and other places, if any) and run the index build inside the lock after `migrator.up()`:
```ts
await withMigrationLock(db, async () => {
    // ... migrator.up() ...
    await orm.getSchemaGenerator().ensureIndexes();
});
```
so both the dedup migration and the unique-index build are serialized and correctly ordered.

Pyatakov (changes requested):

common/src/helpers/migration.ts
vshvets-bc: Good catch — you're right, thanks. Fixed in 41560eb0a.

`Migration()` now:
- forces `ensureIndexes: false` at `MikroORM.init()` so init no longer builds indexes eagerly (outside the lock, against pre-migration data), and
- builds them via `orm.getSchemaGenerator().ensureIndexes()` **inside `withMigrationLock`, after `migrator.up()`** — so the unique-index build is both serialized across replicas and correctly ordered after the dedup/backfill migrations. It's idempotent on re-runs.

I also dropped the now-dead `ensureIndexes: true` from the configs that pass it into `Migration()`: `guardian-service`, `analytics-service`, `indexer-service`, `indexer-worker-service`.

One thing I left out intentionally: `auth-service`, `topic-listener-service` and `notification-service` carry `ensureIndexes: true` on a **separate `MikroORM.init`** that runs concurrently (same `Promise.all`) with `Migration()`, not inside the Migration config. That's a different code path than this PR, and safely disabling it depends on `Migration()` discovering the same entities via `COMMON_CONNECTION_CONFIG` to build those indexes in-lock. Happy to fold that in here or track it as a follow-up — let me know which you'd prefer.

Pyatakov: @vshvets-bc  While the `ensureIndexes: true` was correctly removed from other services, it needs to be kept for `indexer-service` and `indexer-worker-service` because they have their own `indexer-common` package and aren't affected by changes to the shared `common` package.

vshvets-bc:
Good catch, fixed in `8be459d`. Restored `ensureIndexes: true` for **`indexer-service`** and **`indexer-worker-service`** - you're right, they import `Migration` from `@indexer/common`, which this PR doesn't touch, so its `Migration()` still relies on the eager index build at init. The removal stays only for the services on the shared `@guardian/common` `Migration()` (guardian-service, analytics-service), which now builds indexes inside the lock after `migrator.up()`.

Pyatakov (approved):

PR #6341, opened 2026-07-06
fix(policy-engine): fast-fail policy messages when policy is not loaded

vshvets-bc:
## Problem
Policy-directed calls (`checkIfPolicyAlive`, `sendPolicyMessage`, `sendBlockMessage`) use a custom `publish` + shared-reply mux with **no "no responders" detection**. When a policy is not loaded on any policy-service instance (no subscriber on its subject), the call **waits out the full timeout interval** rather than failing fast — 1s / 100s / 5min respectively. This is most visible during rollouts, when a policy is mid-reload and calls stall the whole interval before returning `Block Timeout` / `null`.

## Change
- **common** — add `NatsService.requestOrThrow`, which uses a NATS **core `request()`** over a dedicated inbox. A subject with zero subscribers fails fast via the server's "no responders" reply instead of running the timeout out. It throws distinguishable errors:
  - `NO_RESPONDERS` — nobody subscribed (request **not** delivered → safe to retry)
  - `REQUEST_TIMEOUT` — a responder exists but was too slow (**may** be delivered → do not retry)
  - `MessageError(code)` — responder returned an error
- **guardian-service** — route the three senders through it:
  - `checkIfPolicyAlive` — any failure ⇒ `false`.
  - `sendPolicyMessage` — preserves the soft contract (`NO_RESPONDERS`/`REQUEST_TIMEOUT` ⇒ `null`; responder error propagates).
  - `sendBlockMessage` — retries **only** on `NO_RESPONDERS` (non-delivery ⇒ no double-apply) with a short backoff until the deadline, then `Block Timeout` (504); `REQUEST_TIMEOUT` ⇒ `Block Timeout` (no retry).

The responder side already replies via `msg.respond()`, so core `request()` is transport-compatible; `nats@2.x` supports no-responders.

## Tests
- `common` — `requestOrThrow` outcome mapping + header/subject wiring (7).
- `guardian-service` — the three senders' contracts, incl. `sendBlockMessage` retry-on-`NO_RESPONDERS`-then-success and no-retry-on-`REQUEST_TIMEOUT` (13).

## Notes
Requires policy-service to cleanly unsubscribe a policy's subjects on unload (`UnregisterPolicy`), so no-responders reliably fires — worth a reviewer's eye.

### DIFF
--- a/common/src/mq/nats-service.ts
+++ b/common/src/mq/nats-service.ts
@@ -261,6 +261,54 @@ export abstract class NatsService {
     }
 
     /**
+     * Core NATS request over a dedicated inbox: a subject with no subscribers
+     * fails fast ("no responders") instead of waiting out the timeout. Throws
+     * Error{code:'NO_RESPONDERS'} (not delivered - safe to retry),
+     * Error{code:'REQUEST_TIMEOUT'} (maybe delivered - do NOT retry) or
+     * MessageError(code); otherwise returns the response body.
+     */
+    public async requestOrThrow<T>(
+        subject: string,
+        data?: unknown,
+        timeout: number = 1000,
+        extraHeaders?: Record<string, string>
+    ): Promise<T> {
+        const head = headers();
+        head.append('messageId', GenerateUUIDv4());
+        if (extraHeaders) {
+            for (const [key, value] of Object.entries(extraHeaders)) {
+                head.append(key, value);
+            }
+        }
+        const token = await JwtServicesValidator.sign(subject);
+        head.append('serviceToken', token);
+
+        let msg;
+        try {
+            msg = await this.connection.request(subject, await this.codec.encode(data), { timeout, headers: head });
+        } catch (error: any) {
+            // nats: NoResponders -> '503', Timeout -> 'TIMEOUT'
+            if (error?.code === '503' || /no responders/i.test(error?.message || '')) {
+                const e = new Error(`No responders for "${subject}"`);
+                (e as any).code = 'NO_RESPONDERS';
+                throw e;
+            }
+            if (error?.code === 'TIMEOUT' || /timeout/i.test(error?.message || '')) {
+                const e = new Error(`Timeout for "${subject}"`);
+                (e as any).code = 'REQUEST_TIMEOUT';
+                throw e;
+            }
+            throw error;
+        }
+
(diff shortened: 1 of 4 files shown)


common/src/mq/nats-service.ts
```diff
@@ -260,6 +260,54 @@ export abstract class NatsService {
...
+            }
+            if (error?.code === 'TIMEOUT' || /timeout/i.test(error?.message || '')) {
+                const e = new Error(`Timeout for "${subject}"`);
+                (e as any).code = 'REQUEST_TIMEOUT';
+                throw e;
+            }
+            throw error;
+        }
+
+        const message = (await this.codec.decode(msg.data)) as IMessageResponse<T>;
```
Pyatakov: Unguarded reply decode reintroduces the directLink error-leak the sibling `init()` path was explicitly hardened against.

guardian-service/src/helpers/guardians.ts
```diff
@@ -40,8 +39,14 @@ export class GuardiansService extends NatsService {
...
      */
     public async checkIfPolicyAlive(policyId: string): Promise<boolean> {
-        const exist = await this.sendPolicyMessage<boolean>(PolicyEvents.CHECK_IF_ALIVE, policyId, {}, 1000)
-        return !!exist
+        // Fast fail: no subscriber on the subject -> immediate "no responders".
+        const subject = [policyId, PolicyEvents.CHECK_IF_ALIVE].join('-');
+        try {
+            const alive = await this.requestOrThrow<boolean>(subject, {}, 1000, { policyId });
+            return !!alive;
+        } catch {
```
Pyatakov: bare `catch { return false }` reports a live policy as dead on any non-503 error, and `generateModel` reacts by issuing a spurious `GENERATE_POLICY`.

common/src/mq/nats-service.ts
Pyatakov: unlike the `replySubject` handler in `init()` (which calls `JwtServicesValidator.verify(serviceToken)` on the reply before trusting it), `requestOrThrow` decodes and returns the body with no verification, so with `QM_VERIFICATION` enabled this drops the reply-side auth check.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6347, opened 2026-07-07
perf(policy-service): parallelize bulk DID generation in customLogicBlock and reuse batch setup (#6346)

### LINKED ISSUE #6346: customLogicBlock with ID type DID processes bulk results sequentially — hours to register hundreds of devices
## Problem description

When a `customLogicBlock` returns an array of documents and the block is configured with **Document Id Type = DID**, the engine generates and anchors a DID for every output document **strictly sequentially**, and repeats expensive invariant setup for every item.

Real-world case: a metered-devices (cookstove) policy where an admin approval explodes a device batch into one registered-device VC per device, each with its own DID. Batches of 5–10 devices complete in seconds; batches of a few hundred take **hours**. Target scale for this use case is tens of thousands of devices.

## Root cause

1. `custom-logic-block.ts` processes array results with a serial `for … await` loop (`policy-service/src/policy-engine/blocks/custom-logic-block.ts`, `done()` handler) — concurrency is 1.
2. Each item with `idType: 'DID'` calls `GenerateDID.local()` (`policy-service/src/policy-engine/policy-actions/generate-did.ts`), which **per item**:
   - re-fetches the root topic (same topic for every item)
   - re-loads user credentials + Hedera credentials + relayer account (same owner for every item — the batch metadata is computed once from the first document)
   - constructs a new `MessageServer`
   - uploads the DID document to IPFS, then submits a `TopicMessageSubmitTransaction` and waits for consensus (~2–5 s), strictly in series (`common/src/hedera-modules/message/message-server.ts`, `sendMessage`)

Nothing is CPU-bound: total time is N × (IPFS round-trip + Hedera consensus wait + credential loading + DB writes) with no overlap. At ~3–5 s per item, 1 000 devices ≥ 1 hour.

`SendMessages.local` (`policy-actions/send-messages.ts`) already hoists credentials/`MessageServer` out of its loop, but still submits sequentially — same pattern.

## Suggested fix

1. Process array results with bounded concurrency (Hedera has no per-account nonce ordering — each transaction carries its own transactionId/valid-start, so concurrent submits from one operator account are safe). Must fall back to sequential processing when a record/replay controller is active, because recording consumes UUID/DID sequences positionally.
2. Resolve the batch-invariant setup (root topic, user credentials, `MessageServer`) once per batch and share it across items.

A PR implementing both is linked below.

## Additional scaling note

`customLogicBlock.getSources()` calls `documentsSourceAddon.getFromSource(user, null)` with no pagination, so a policy that uses sources to build an allowlist (e.g. filtering MRV rows to registered devices) loads **every** matching VC into memory and structured-clones the full set into the sandbox worker on every submission. At 10k+ documents this becomes the next bottleneck after DID generation is fixed. Possibly worth a separate issue.

mariana-andruk-qa:
Fixes #6346

## Problem

When a `customLogicBlock` returns an array of documents with **Document Id Type = DID**, the engine processes them strictly sequentially and repeats invariant setup (root topic fetch, credential loading, `MessageServer` construction) for every item. Each item then pays an IPFS upload + a Hedera consensus wait in series. Registering a few hundred devices this way takes hours; the target use case (metered-device MRV) needs tens of thousands.

## Changes

**`custom-logic-block.ts`** — replaces the serial `for … await` loop over array results with `processItems()`, a bounded-concurrency pool:
- default 10 in flight, tunable via `CUSTOM_LOGIC_CONCURRENCY`
- falls back to strictly sequential, in-order processing when a recording/replay controller is active (`ref.components.runAndRecordController`) — record/replay consume UUID/DID sequences positionally, so concurrent generation would break deterministic replay
- result order matches input order
- on the first failure no further items are scheduled; items already in flight complete (awaited via `Promise.allSettled`) before the first error is rethrown, so no side effects continue in the background after rejection. Note this differs from the old serial loop, which never started items after the failed one — with concurrency, up to `limit - 1` already-started items may still complete.

**`policy-actions/generate-did.ts`** — `GenerateDID.local()` accepts an optional `IGenerateDidBatch` context holding promise-memoized batch invariants (root topic, user credentials, `MessageServer`). Concurrent callers share the in-flight promise, so setup is resolved once per batch instead of once per item. Without a batch the behavior is unchanged.

**`policy-actions/utils.ts`** — `generateId()` threads the batch through to `GenerateDID.local` on the LOCAL path only. The remote request/response path and the other `generateId` callers (`request-vc-document-block`, `request-vc-document-block-addon`) are untouched.

## Safety

- Hedera has no per-account nonce ordering — each transaction carries its own transactionId/valid-start, so concurrent submits from one operator account are safe.
- Shared `MessageServer` under concurrency: the constructor is pure field assignment and `setTopicObject` always re-sets the same root topic; `sendMessage` holds no per-call instance state.
- Produced artifacts are unchanged: one DID message + one signed VC per item.
- Record/replay determinism is preserved via the sequential fallback.

## Tests

New `policy-service/tests/unit-tests/blocks/custom-logic-bulk-did.test.mjs` (14 cases, all passing):
- pool: order preservation under out-of-order completion, default cap of 10 verified via in-flight tracking, `CUSTOM_LOGIC_CONCURRENCY` override, invalid-value fallback, sequential + in-order under record/replay, rejection propagation, no new items scheduled after a failure, in-flight items awaited before rejection
- batch reuse: setup resolved exactly once across sequential and concurrent calls, distinct DIDs per item, per-call setup preserved without a batch, pre-seeded context honored
- threading: batch object identity reaches `GenerateDID.local` for DID ids; UUID path ignores it

### DIFF
--- a/policy-service/src/policy-engine/blocks/custom-logic-block.ts
+++ b/policy-service/src/policy-engine/blocks/custom-logic-block.ts
@@ -13,6 +13,7 @@ import { PolicyUtils } from '../helpers/utils.js';
 import { ExternalDocuments, ExternalEvent, ExternalEventType } from '../interfaces/external-event.js';
 import { fileURLToPath } from 'node:url';
 import { PolicyActionsUtils } from '../policy-actions/utils.js';
+import { IGenerateDidBatch } from '../policy-actions/generate-did.js';
 import { BlockActionError } from '../errors/index.js';
 import { collectTablesPack, hydrateTablesInObject, loadFileTextById } from '../helpers/table-field.js';
 import { RecordActionStep } from '../record-action-step.js';
@@ -219,6 +220,7 @@ export class CustomLogicBlock {
                     }
                     metadata = await this.aggregateMetadata(documents, user, ref, userId);
                 }
+                const didBatch: IGenerateDidBatch = {};
                 const done = async (result: any | any[], final: boolean) => {
                     if (!result) {
                         await triggerEvents(null);
@@ -240,14 +242,11 @@ export class CustomLogicBlock {
                         if (options.unsigned) {
                             return await this.createUnsignedDocument(json, ref, actionStatus?.id);
                         } else {
-                            return await this.createDocument(json, metadata, ref, userId, actionStatus?.id, user);
+                            return await this.createDocument(json, metadata, ref, userId, actionStatus?.id, user, didBatch);
                         }
                     }
                     if (Array.isArray(result)) {
-                        const items: IPolicyDocument[] = [];
-                        for (const r of result) {
-                            items.push(await processing(r))
-                        }
+                        const items = await this.processItems(result, processing, ref);
                         await triggerEvents(items);
                         if (final) {
                             try {
@@ -490,6 +489,39 @@ export class CustomLogicBlock {
     }
 
     /**
+     * Process result items with bounded concurrency
+     * @param items
+     * @param task
+     * @param ref
+     */
+    private async processItems(
+        items: any[],
+        task: (json: any) => Promise<IPolicyDocument>,
+        ref: IPolicyCalculateBlock
+    ): Promise<IPolicyDocument[]> {
+        // Recording and replay consume UUID/DID sequences in order, so they require sequential processing.
+        let limit = 1;
+        if (!ref.components.runAndRecordController) {
+            const configured = parseInt(process.env.CUSTOM_LOGIC_CONCURRENCY, 10);
+            limit = Number.isFinite(configured) && configured > 0 ? configured : 10;
+        }
+        limit = Math.min(limit, items.length);
+        const results = new Array<IPolicyDocument>(items.length);
+        let index = 0;
+        const workers: Promise<void>[] = [];
+        for (let i = 0; i < limit; i++) {
+            workers.push((async () => {
+                while (index < items.length) {
+                    const current = index++;
+                    results[current] = await task(items[current]);
+                }
+            })());
+        }
+        await Promise.all(workers);
+        return results;
+    }
+
+    /**
      * Generate document metadata
      * @param documents
      * @param user
@@ -576,7 +608,8 @@ export class CustomLogicBlock {
         ref: IPolicyCalculateBlock,
         userId: string | null,
         actionStatusId: string,
-        user?: PolicyUser
+        user?: PolicyUser,
+        didBatch?: IGenerateDidBatch
     ): Promise<IPolicyDocument> {
         const {
             owner,
@@ -623,7 +656,7 @@ export class CustomLogicBlock {
             user: owner,
             relayerAccount,
             userId
-        }, actionStatusId);
+        }, actionStatusId, didBatch);
         if (newId) {
             vcSubject.id = newId;
         }
--- a/policy-service/src/policy-engine/blocks/custom-logic-block.ts
+++ b/policy-service/src/policy-engine/blocks/custom-logic-block.ts
@@ -508,16 +508,26 @@ export class CustomLogicBlock {
         limit = Math.min(limit, items.length);
         const results = new Array<IPolicyDocument>(items.length);
         let index = 0;
+        let failed = false;
         const workers: Promise<void>[] = [];
         for (let i = 0; i < limit; i++) {
             workers.push((async () => {
-                while (index < items.length) {
+                while (!failed && index < items.length) {
                     const current = index++;
-                    results[current] = await task(items[current]);
+                    try {
+                        results[current] = await task(items[current]);
+                    } catch (error) {
+                        failed = true;
+                        throw error;
+                    }
                 }
             })());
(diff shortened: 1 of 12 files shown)


policy-service/src/policy-engine/blocks/custom-logic-block.ts
```diff
@@ -489,6 +488,39 @@ export class CustomLogicBlock {
...
+     */
+    private async processItems(
+        items: any[],
+        task: (json: any) => Promise<IPolicyDocument>,
+        ref: IPolicyCalculateBlock
+    ): Promise<IPolicyDocument[]> {
+        // Recording and replay consume UUID/DID sequences in order, so they require sequential processing.
+        let limit = 1;
+        if (!ref.components.runAndRecordController) {
+            const configured = parseInt(process.env.CUSTOM_LOGIC_CONCURRENCY, 10);
```
Pyatakov, line 505: New env var `CUSTOM_LOGIC_CONCURRENCY` isn't reflected in any `.env` file and docs.

policy-service/src/policy-engine/blocks/custom-logic-block.ts
```diff
@@ -489,6 +488,39 @@ export class CustomLogicBlock {
...
+        const workers: Promise<void>[] = [];
+        for (let i = 0; i < limit; i++) {
+            workers.push((async () => {
+                while (index < items.length) {
+                    const current = index++;
+                    results[current] = await task(items[current]);
+                }
+            })());
+        }
+        await Promise.all(workers);
```
Pyatakov: On failure, sibling workers keep draining the whole queue, this diverges from the old serial semantics.

When `task` rejects for one worker, only that worker's async function stops. The other `limit - 1` workers keep pulling from the shared index and run task on every remaining item to completion. `Promise.all` rejects on the first error, but the surviving workers continue their side effects in the background (DID generation, IPFS upload, Hedera submit).

The previous `for … await` loop broke on the first failure and never touched later items, so the PR's "_first item failure rejects the whole call (same semantics as before)_" isn't accurate for a non-transactional operation: a mid-batch failure now still commits messages for most of the remaining items.

Pyatakov (changes requested):

policy-service/src/policy-engine/blocks/custom-logic-block.ts
mariana-andruk-qa: Added — `CUSTOM_LOGIC_CONCURRENCY="10"` is now in all four `configs/.env.*.guardian.system` templates next to the other custom-logic variables, and documented in a new "Bulk processing" section of `docs/available-policy-workflow-blocks/customlogicblock.md`.

policy-service/src/policy-engine/blocks/custom-logic-block.ts
mariana-andruk-qa: You're right, thanks — the description overstated the parity. Fixed in e06b91e: the first failure sets a flag so no worker pulls another item, and the pool awaits every in-flight task (`Promise.allSettled`) before rethrowing the first error, so no side effects keep running in the background after it rejects. Items already in flight when the failure happens still complete — that's inherent to concurrent execution — but nothing new starts. Added two regression tests pinning both behaviors and corrected the PR description.

policy-service/src/policy-engine/blocks/custom-logic-block.ts
Pyatakov: @mariana-andruk-qa those under the `policy-service/configs` path also need to be updated

mariana-andruk-qa: Done in b1dbb5a, added to all three files under policy-service/configs (.env.policy, .env.policy.develop, .env.policy.template).

Pyatakov (approved):

dubgeis:
@swidnikk can you please confirm when docs are done on this? I don't see them.

swidnikk:
@dubgeis I don't have the full context for this item. Would you please take an initial pass and then let me know when it is ready? I can then help to ensure consistency.

dubgeis:
@mariana-andruk-qa @danielnorkin I didn't put this together, but we'd appreciate your help if you have time to draft some docs on Monday to help with this.

dubgeis:
@gautamp8 if you're taking advantage of this feature please let us know if it makes sense to use as an example.

swidnikk:
Yes an example would be helpful along with a few bullet points of guidance for a concept, reference, task, or guide article.

Please consult the new doc templates found under community guidelines for initial drafts and I can finalize.

gautamp8:
> @gautamp8 if you're taking advantage of this feature please let us know if it makes sense to use as an example.

Yes, we reported this issue where MGS was getting stuck when multiple devices were being registered - each with a single DID. I'm not sure what's the ask here @dubgeis for docs, if you need an example - we're using it to register active devices in MECD v2.0

PR #6353, opened 2026-07-09
feat(frontend): save request form draft to a JSON file and restore from file

vshvets-bc:
## Summary

Adds the ability to export a `requestVcDocumentBlock` form draft to a local JSON file and restore a draft from such a file, filling the form exactly like a regular server/autosave draft. Frontend-only — no backend or DB changes.

## Motivation

Drafts today live either server-side (Save Draft) or in browser IndexedDB autosave. This adds a portable draft users can save to disk, share, and reload into the same form (e.g. across machines/sessions).

## Changes

`requestVcDocumentBlock`, both render surfaces:

- Dialog render (`request-document-block-dialog`): the footer gets two standalone buttons, *Save draft to file* / *Restore from draft file*, alongside the existing *Save Draft* button (previously a single split button with a menu).
- Page render (`request-document-block`): the same two standalone buttons above the form (page mode has no *Save Draft* button).
- Both buttons and their hidden file input are implemented once in a shared `request-draft-file-actions` component, bound via `[schema]` / `[policyId]` / `[blockId]` / `[dataForm]` and a `(draftImported)` output — the dialog and page no longer duplicate the button markup or the export/import handlers (~130 duplicated lines removed).

## Behaviour

- **Export**: serializes the raw form value into an envelope `{ type, version, schema (iri), policyId, blockId, document }` and downloads `draft_<name>.json`, where `<name>` is the schema `iri` (falling back to schema `name`), sanitized to safe filename characters so schemas with spaces/special characters in their name can't produce a broken download.
- **Restore**: reads the file and reuses the exact regular-draft restore path (`tablePersist.restoreTablesFromDraft` + `preset`); accepts the envelope or a bare form object. If restoring persisted tables fails, the user now gets a warning toast (*"Draft loaded, but some table data could not be restored"*) instead of a console-only failure masked as success.
- **Guards**: on schema/policyId/blockId/type mismatch, a confirmation dialog lists the mismatches; the user can Continue anyway or Cancel (non-blocking).
- **Override protection**: if the form already contains data, a confirmation warns it will be replaced before loading.

## Notes

- Export/import/mismatch-check/confirm-dialog/apply logic now lives in a single `RequestDraftFileService`, reused by both render surfaces instead of being duplicated per-component.
- Drops `SplitButtonModule` — the split-button/menu is replaced by two plain buttons per surface, still using `CustomConfirmDialogComponent` for confirmations.
- Exports the un-prepared form value so it round-trips through `preset()` identically to autosave/regular-draft restore.


frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block/dialog/request-document-block-dialog.component.ts
```diff
@@ -404,6 +404,138 @@ export class RequestDocumentBlockDialog {
         }
     }
 
+    public onExportDraft(): void {
```
Pyatakov: The entire draft file feature (`onExportDraft`, `triggerImportDraft`, `onImportDraft`, `getDraftMismatches`, `confirmDraftImport`, `formHasData`, `hasMeaningfulValue`, `applyImportedDraft`) is duplicated almost verbatim between the page component (lines 482–611) and the dialog component (lines 407–537). The only differences are `this.preset `vs `this.parent.preset` and the extra `this.dataSaved = false` in the dialog. Please extract this into a shared service (e.g. `RequestDraftFileService`) or a common base, passing the preset callback in. Two copies of ~130 lines will drift and double the maintenance surface.

frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block/dialog/request-document-block-dialog.component.ts
```diff
@@ -404,6 +404,138 @@ export class RequestDocumentBlockDialog {
...
+            schema: this.schema?.iri || null,
+            policyId: this.policyId,
+            blockId: this.id,
+            document: formValue,
+        };
+        const blob = new Blob([JSON.stringify(envelope, null, 2)], { type: 'application/json' });
+        const url = URL.createObjectURL(blob);
+        const a = document.createElement('a');
+        a.href = url;
+        a.download = `draft_${this.schema?.name || 'document'}.json`;
```
Pyatakov: `schema.name` is authored text and may contain spaces,/, or other characters that produce awkward or broken filenames. Consider shortening the name, or falling back to `schema.iri`.

frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block/dialog/request-document-block-dialog.component.ts
```diff
@@ -404,6 +404,138 @@ export class RequestDocumentBlockDialog {
...
+            return Object.values(value).some((v) => this.hasMeaningfulValue(v));
+        }
+        return true;
+    }
+
+    private async applyImportedDraft(doc: any): Promise<void> {
+        try {
+            await this.tablePersist.restoreTablesFromDraft(doc);
+        } catch (e) {
+            console.error(e);
```
Pyatakov: A thrown `restoreTablesFromDraft` is caught and only `console.error`-logged, yet execution continues to `preset(...)` and still shows the `'Draft loaded from file.' / 'Import complete'` success toast. If table restore fails the user is told the import succeeded. Please surface a warning toast in the catch branch instead of silently continuing to a success message.

Pyatakov:
@vshvets-bc @nazlo90 Could you please address the 3 issues noted in the review, plus these two:

1. The description says the Save Draft button "becomes a split button" reusing `SplitButtonModule`. The actual code uses two plain `guardian-button-secondary` buttons and removes `SplitButtonModule` from `policy-engine.module.ts`. The removal itself is fine,  but the description should be corrected so change log aren't misled.
2. `let parsed: any` (`dialog...component.ts:439`) and `getDraftMismatches(parsed: any)` (`dialog...component.ts:468`) lose all type safety on the parsed file. Define an explicit `interface RequestVcDraftEnvelope { type; version; schema; policyId; blockId; document }` and type the parse/mismatch helpers against it. This (1) documents the on-disk format and (2) catches shape mistakes, and it pairs naturally with extracting the shared service.

Pyatakov (changes requested):

frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block/draft-file-actions/draft-file-actions.component.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * "Save draft to file" / "Restore from draft file" button pair, shared between
+ * the request document page and dialog variants to avoid duplicating markup
+ * and the export/import handling.
+ */
+@Component({
+    selector: 'request-draft-file-actions',
+    templateUrl: './draft-file-actions.component.html',
+    styleUrls: ['./draft-file-actions.component.scss'],
```
Pyatakov, line 13: The new component lacks explicitly set `standalone: false`.
This project is on Angular 21, where `standalone` defaults to `true`. The new `DraftFileActionsComponent`'s `@Component` decorator omits the flag, so it's `standalone`, yet it's added to `PolicyEngineModule`'s `declarations`.
This fails to compile:
`Error: Component DraftFileActionsComponent is standalone, and cannot be declared in an NgModule.`

Pyatakov:
@nazlo-climission The last one comment is left unaddressed, could you take a look?

`let parsed: any` (`request-draft-file.service.ts:66`) and `getDraftMismatches(parsed: any)` (`request-draft-file.service.ts:101`) lose all type safety on the parsed file. Define an explicit `interface RequestVcDraftEnvelope { type; version; schema; policyId; blockId; document }` and type the parse/mismatch helpers against it.

Pyatakov (approved):

PR #6361, opened 2026-07-13
fix(web-proxy): prevent stale SPA after redeploy via index.html no-cache

vshvets-bc:
## Problem

The Angular frontend is served by nginx with **no cache-control headers**, so after a redeploy a browser can keep serving a cached `index.html`. Since `index.html` is the one fixed-name file that references the build's content-hashed bundles (`main.[hash].js`), a stale `index.html` keeps pointing at the previous build - users stay on the old app until they manually hard-refresh, which produces confusing "stale UI" reports that are easily misattributed to the backend.

## Fix

Add explicit cache policy to every nginx config that serves the SPA:

- **`index.html`** → `Cache-Control: no-cache, no-store, must-revalidate` so every load revalidates and a new deploy is picked up immediately.
- **hashed assets** (`js`/`css`/fonts/images) → `Cache-Control: public, immutable`, `expires 1y` - safe because Angular content-hashes filenames per build (`outputHashing: all`), so a new build yields new filenames.
- SPA `try_files … /index.html` fallback preserved.

Applied consistently to:
- `web-proxy/configs/default.conf`
- `web-proxy/configs/https-default.conf`
- `web-proxy/configs/demo.conf`
- `web-proxy/configs/image.conf`
- `k8s-manifests/1-config/frontend-service-nginx-conf.yaml`

## Verification

- `curl -I https://<host>/` → `Cache-Control: no-cache, no-store, must-revalidate`
- `curl -I https://<host>/main.<hash>.js` → `Cache-Control: public, immutable`
- After a redeploy, a reload loads the new build with no manual cache clear.

## Notes

- No functional/proxy changes; caching headers only.
- Zero risk to the hashed assets (immutable is correct for content-addressed files); the only always-revalidated file is `index.html` (tiny).

Signed-off-by: Volodymyr Shvets <Volodymyr.Shvets@envisionblockchain.com>

### DIFF
--- a/k8s-manifests/1-config/frontend-service-nginx-conf.yaml
+++ b/k8s-manifests/1-config/frontend-service-nginx-conf.yaml
@@ -8,6 +8,22 @@ data:
       listen  [::]:80;
       server_name  localhost;
 
+      # Hashed build assets (main.[hash].js etc.) - safe to cache forever
+      location ~* \.(?:js|css|woff2?|ttf|eot|svg|png|jpg|jpeg|gif|ico)$ {
+          root   /usr/share/nginx/html;
+          expires 1y;
+          add_header Cache-Control "public, immutable";
+          try_files $uri =404;
+      }
+
+      # index.html - never cache so a redeploy is picked up on next load
+      location = /index.html {
+          root   /usr/share/nginx/html;
+          add_header Cache-Control "no-cache, no-store, must-revalidate";
+          add_header Pragma "no-cache";
+          expires 0;
+      }
+
       location / {
           root   /usr/share/nginx/html;
           try_files $uri $uri/ /index.html;
--- a/k8s-manifests/1-config/frontend-service-nginx-conf.yaml
+++ b/k8s-manifests/1-config/frontend-service-nginx-conf.yaml
@@ -8,20 +8,11 @@ data:
       listen  [::]:80;
       server_name  localhost;
 
-      # Hashed build assets (main.[hash].js etc.) - safe to cache forever
-      location ~* \.(?:js|css|woff2?|ttf|eot|svg|png|jpg|jpeg|gif|ico)$ {
-          root   /usr/share/nginx/html;
-          expires 1y;
-          add_header Cache-Control "public, immutable";
-          try_files $uri =404;
-      }
-
       # index.html - never cache so a redeploy is picked up on next load
       location = /index.html {
           root   /usr/share/nginx/html;
           add_header Cache-Control "no-cache, no-store, must-revalidate";
           add_header Pragma "no-cache";
-          expires 0;
       }
 
       location / {
--- a/web-proxy/configs/default.conf
+++ b/web-proxy/configs/default.conf
@@ -100,20 +100,11 @@ server {
         proxy_redirect off;
     }
 
-    # Hashed build assets (main.[hash].js etc.) - safe to cache forever
-    location ~* \.(?:js|css|woff2?|ttf|eot|svg|png|jpg|jpeg|gif|ico)$ {
-        root   /usr/share/nginx/html;
-        expires 1y;
-        add_header Cache-Control "public, immutable";
-        try_files $uri =404;
-    }
-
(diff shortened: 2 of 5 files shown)


k8s-manifests/1-config/frontend-service-nginx-conf.yaml
```diff
@@ -8,6 +8,22 @@ data:
       listen  [::]:80;
       server_name  localhost;
 
+      # Hashed build assets (main.[hash].js etc.) - safe to cache forever
+      location ~* \.(?:js|css|woff2?|ttf|eot|svg|png|jpg|jpeg|gif|ico)$ {
```
Pyatakov: In nginx, regex locations take precedence over plain-prefix locations (`location /api-docs/v1/`, `location /mongo-admin`, etc.). Because this regex is not path-scoped, any request to a proxied backend whose URL ends in one of these extensions will now be intercepted and served from `/usr/share/nginx/html` → 404, instead of being proxied.

Concretely broken:
`/api-docs/v1/` (all 5 configs) -> Swagger UI. Its assets (`swagger-ui.css`, `swagger-ui-bundle.js`, favicon `.png`) end in `.css/.js/.png` and will 404, which results to API docs page renders unstyled/non-functional.
`/mongo-admin` -> mongo-express static `.css/.js` assets break.

Fix option:
Add the `^~` modifier to the proxied prefix locations (`location ^~ /api-docs/v1/`, `^~ /mongo-admin`, `^~ /topic-viewer/`, `^~ /mrv-sender/`), this tells nginx to skip regex evaluation once the prefix matches.

web-proxy/configs/default.conf
```diff
@@ -100,6 +100,22 @@ server {
         proxy_redirect off;
     }
 
+    # Hashed build assets (main.[hash].js etc.) - safe to cache forever
+    location ~* \.(?:js|css|woff2?|ttf|eot|svg|png|jpg|jpeg|gif|ico)$ {
+        root   /usr/share/nginx/html;
+        expires 1y;
```
Pyatakov: Combining the `expires` with a manual `add_header Cache-Control` emits two (duplicate) `Cache-Control` headers:
assets: `expires 1y` -> `Cache-Control: max-age=31536000`, plus `add_header` -> `Cache-Control: public, immutable`
index.html: `expires 0` -> `Cache-Control: no-cache`, plus `add_header` -> `Cache-Control: no-cache, no-store, must-revalidate`
While this is valid HTTP (comma-combined by clients) and works in practice, but is sloppy and some intermediaries/tools may read only the first value. Cleaner to drop `expires` and put everything in one header, e.g. assets: `add_header Cache-Control "public, max-age=31536000, immutable"`; and rely solely on `add_header` for index.html.

Pyatakov (changes requested):

Pyatakov:
@vshvets-bc please take a look on these comments. They needs to be applied to every `.conf` file

vshvets-bc:
Thanks @Pyatakov. Rather than guard every proxied prefix with `^~`, I've **removed the asset-cache location entirely** and kept only the `index.html` no-cache rule (which is the actual goal of this PR - pick up a redeploy on next load). Pushed in `c0dd2a6`.

This resolves both comments:

1. **Regex precedence / proxied assets** - the `location ~* \.(js|css|png|…)$` block is gone, so it can no longer intercept Swagger UI (`/api-docs/v1/`) or mongo-express (`/mongo-admin`) static assets. No `^~` guards needed.
2. **Duplicate `Cache-Control`** - dropped the redundant `expires 0` from the `index.html` location, so it now emits a single `Cache-Control: no-cache, no-store, must-revalidate`.

Applied across every affected file: `web-proxy/configs/{default,demo,https-default,image}.conf` and `k8s-manifests/1-config/frontend-service-nginx-conf.yaml`. The net change per file is now just the `location = /index.html` no-cache block.

Note: with immutable asset caching removed, assets fall back to nginx defaults (revalidated), so there's no stale-asset risk - and Angular's content-hashed filenames already make long-term asset caching a separate, optional optimization we can revisit later if desired.

PR #6367, opened 2026-07-13
fix(frontend): show "step unavailable" message instead of infinite spinner in step block

vshvets-bc:
## Problem
The **step block** (`interfaceStepBlock`) in the policy viewer spins **forever** when the workflow advances to a step the current user cannot access (for example, the next step belongs to a VVB/validator role after the proponent submits a project).

The container serializes a non-permitted active child as `null`, so `getBlockData` returns **200** with `blocks[index] === null`. The old getter:

```ts
get loading() { return !this.blocks || !this.blocks.length || !this.activeBlock; }
```

stayed `true` forever whenever there was no viewable active child - there is **no pending request and no error**, so nothing ever clears the spinner and the user is stuck.

## Fix (frontend only)
- `loading` now tracks a real `loaded` flag, set once any response is processed, so the spinner always clears.
- New `unavailable` getter (`loaded && !activeBlock`) renders a short, friendly message instead of the spinner, explaining the step is handled by another participant/role.
- `_onError` stops the spinner on any error (not just 503) and falls through to the same message rather than hanging.

## Behavior

| Backend returns | Before | After |
|---|---|---|
| active child present | renders block | renders block (unchanged) |
| active child `null` (permission/role-gated) | ∞ spinner | friendly "not available" message |
| 503 (block closed) | cleared | cleared (unchanged) |
| other error | ∞ spinner | spinner stops → message |

## Notes
- Frontend-only change scoped to `step-block.component.{ts,html,scss}`; no API changes. The backend returning a null active child for a non-permitted step is left as-is (it's correct) - this only fixes how the UI presents it.
- Verified the equivalent change builds under Angular AOT and passes step-block unit tests in a downstream build.

### DIFF
--- a/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
+++ b/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
@@ -9,7 +9,7 @@
     //width: fit-content;
 }
 
-.step-unavailable {
+.step-message {
     display: flex;
     flex-direction: column;
     align-items: center;
@@ -18,22 +18,42 @@
     max-width: 460px;
     margin: 48px auto;
     padding: 32px 24px;
-    color: #64748b;
+    color: var(--guardian-grey-color, #64748b);
 
     &__icon {
-        color: #94a3b8;
+        color: var(--guardian-grey-color, #94a3b8);
         line-height: 0;
     }
 
     &__title {
         font-size: 16px;
         font-weight: 600;
-        color: #334155;
+        color: var(--guardian-font-color, #334155);
     }
 
     &__text {
         font-size: 14px;
         line-height: 1.5;
+    }
+
+    &--error &__icon {
+        color: var(--guardian-failure-color, #d32f2f);
+    }
+
+    &__retry {
+        margin-top: 12px;
+        padding: 8px 20px;
+        font-size: 14px;
+        font-weight: 500;
+        cursor: pointer;
+        border: 1px solid var(--guardian-primary-color, #4d6df5);
+        border-radius: 6px;
+        background: var(--guardian-primary-color, #4d6df5);
+        color: var(--guardian-on-primary-color, #fff);
+
+        &:hover {
+            opacity: 0.9;
+        }
     }
 }
 
--- a/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
+++ b/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
@@ -18,10 +18,10 @@
     max-width: 460px;
     margin: 48px auto;
     padding: 32px 24px;
-    color: var(--guardian-grey-color, #64748b);
+    color: var(--guardian-grey-color-4, #6C7791);
 
     &__icon {
-        color: var(--guardian-grey-color, #94a3b8);
+        color: var(--guardian-grey-color-4, #6C7791);
         line-height: 0;
     }
 
--- a/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
+++ b/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
@@ -9,6 +9,34 @@
     //width: fit-content;
 }
 
+.step-unavailable {
+    display: flex;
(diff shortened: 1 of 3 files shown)


frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
```diff
@@ -9,6 +9,34 @@
     //width: fit-content;
 }
 
+.step-unavailable {
```
Pyatakov: This block uses literal hex values (`#64748b`, `#94a3b8`, `#334155`), while the existing `.loading` rule right below uses `var(--guardian-background, #fff)`. These won't adapt to a dark theme. Prefer the CSS-variable convention already present in the file.

frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.ts
```diff
@@ -83,8 +95,14 @@ export class StepBlockComponent implements OnInit {
 
     private _onError(e: HttpErrorResponse) {
```
Pyatakov, line 111: In `_onError`, any error that isn't 503, sets `loaded = true` and falls through to the same panel that says "The workflow has moved to a step handled by another participant or role." That statement is factually wrong for a genuine server/network failure, and it silently hides real outages behind a reassuring message. The old behavior was also bad, but at least visibly broken. Consider distinguishing a true error state (generic "something went wrong / try again") from the permission-gated `null` case.

Pyatakov (changes requested):

vshvets-bc:
Thanks for the review @Pyatakov - both points addressed in `254367e`:

**1. Theme-aware styles.** Replaced the literal hex values with the `var(--guardian-*, fallback)` convention already used in this stylesheet: `--guardian-font-color` (title), `--guardian-grey-color` (text/icon), `--guardian-failure-color` (error icon), `--guardian-primary-color` / `--guardian-on-primary-color` (retry button). So it now adapts to the dark theme.

**2. Real errors vs. permission-gated steps.** Good catch - conflating them was misleading. The two cases are now distinct:

- **Permission-gated** (`503`, or a successful `200` whose active child is `null`): the "not your turn / another participant is handling this step" info message.
- **Genuine failure** (any other error): a separate `errored` state with a neutral **"Couldn't load this step - something went wrong… please try again"** message and a **Try again** button (`retry()`), so a real outage is no longer disguised as a normal workflow message.

`unavailable` now excludes the error case (`loaded && !hasError && !activeBlock`); `errored` is `loaded && hasError`.

Verified the equivalent change builds under Angular AOT and passes step-block unit tests (error-state + retry cases included) in a downstream build.

frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
```diff
@@ -9,6 +9,54 @@
...
+.step-message {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    text-align: center;
+    gap: 8px;
+    max-width: 460px;
+    margin: 48px auto;
+    padding: 32px 24px;
+    color: var(--guardian-grey-color, #64748b);
```
Pyatakov: In this design system `--guardian-grey-color` is a surface/background grey, not a text grey.
Light theme: `#EFF3F7` (near-white) rendered on `--guardian-background: #FFFFFF` makes body text and the info icon effectively invisible.
Dark theme: `#303746` (very dark slate) on `--guardian-background: #171B24`, dark-on-dark, also nearly invisible.
The readable fallback `#64748b` never applies because the variable is defined.

Please use `--guardian-grey-color-4` (`#6C7791` light / `#AAB4C5` dark), which give real contrast in each theme.

frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.scss
```diff
@@ -9,6 +9,54 @@
...
+
+    &__icon {
+        color: var(--guardian-grey-color, #94a3b8);
```
Pyatakov: Please use `--guardian-grey-color-4` (`#6C7791` light / `#AAB4C5` dark)

Pyatakov (changes requested):

Pyatakov (approved):

PR #6370, opened 2026-07-15
fix(queue-service): add GridFS connection to queue-service startup

nazlo-climission:
**Description**:
This PR connects GridFS at queue-service startup so the `TaskEntity` large-payload offload actually works - without it `DataBaseHelper.gridFS` is `undefined` and publishing any policy whose task payload exceeds `TASK_DATA_GRIDFS_LIMIT` (5 MB) fails with `Cannot read properties of undefined (reading 'openUploadStream')` (AMS-III.F v5 with Tools 05/21; VM0051 hits the same crash).

* Add `DatabaseServer.connectGridFS()` call after `DatabaseServer.connectBD(db)` in `guardian/queue-service/src/app.ts` (same bootstrap pattern as guardian-service and policy-service)
* Add `tests/startup-gridfs.test.mjs` - source-level regression guard asserting queue-service bootstrap connects the DB, connects GridFS, and connects GridFS after the DB (app.ts is a top-level IIFE that dials Mongo/NATS on import, so wiring is asserted at source level instead of importing it)

**Notes for reviewer**:
Repro before the fix: publish Tool 05 & Tool 21 → import `AMS-III.F v5.policy` → publish → async task fails with the toast `Publish policy: Cannot read properties of undefined (reading 'openUploadStream')`, no failed HTTP requests (publish returns 202; error arrives via task status). After the fix the publish succeeds and payloads > 5 MB are offloaded to GridFS (`dataFileId` set) and restored on load. Also unblocks `__restoreData()` / `__deleteDataFile()` on the same `gridFS` handle. One-line runtime change; blast radius is queue-service only, no migration, rollback = revert.

**Checklist**


Pyatakov:
@nazlo-climission There's an odd filename issue, the added tests won't be picked up because the filename contains UTF-8 bytes at the end (`\342\200\216`):
`diff --git "a/queue-service/tests/startup-gridfs.test.mjs\342\200\216" "b/queue-service/tests/startup-gridfs.test.mjs\342\200\216"`

Here's how you can verify this locally:
```
# Fetch the PR branch, then:
ls queue-service/tests/ | cat -v          # shows: startup-gridfs.test.mjsM-bM-^@M-^N
ls queue-service/tests/*.test.mjs         # -> "no matches" (glob won't match it)
```
Fix: delete and recreate the file with a plain ASCII name (retype the name rather than copy-pasting, since the stray mark rides along on paste)

PR #6372, opened 2026-07-15
fix(policy-service): close Custom Logic JS sandbox escape via outer-realm intrinsics (TM-006)

vshvets-bc:
## Problem

The Custom Logic Block JavaScript worker (`policy-service/src/policy-engine/helpers/workers/custom-logic-worker.ts`) runs user-supplied policy code in a `node:vm` context and copies host-realm intrinsics onto the sandbox:

```ts
sandbox.JSON = JSON;
sandbox.Math = Math;
sandbox.Date = Date;
// … Array, Object, String, Number, Boolean, RegExp, Map, Set, Promise, …
```

`node:vm` is **not a security boundary** (per the Node.js docs). Each host intrinsic assigned onto the sandbox carries the **host realm's** `Function` constructor. A malicious policy author can walk the prototype chain from any of them back to the outer realm and execute arbitrary code in the policy-service process:

```js
JSON.constructor.constructor('return process')().env.VAULT_TOKEN
```

`Function('return process')` here is created in the **outer** realm (because `JSON.constructor.constructor` resolved there), so the context's `codeGeneration: { strings: false }` restriction does not apply, and the returned `process` is the worker thread's real Node global — exposing `process.env` (Vault tokens, DB creds), `require`, and the filesystem.

This corresponds to **TM-006 / TM-029 (Critical)** and mitigation **M-005** in the April 2026 Guardian Secure Design Review & Threat Model.

## Fix

Stop copying host intrinsics onto the sandbox. `vm.createContext` already gives user code the **context's own** native intrinsics (`JSON`, `Math`, `Object`, `Array`, …), whose `constructor.constructor` chain terminates at the **context's** `Function` — which `codeGeneration.strings: false` blocks. Removing the host-intrinsic assignments closes the escape while leaving the documented API intact.

- `mathjs` / `formulajs` remain exposed as **module-namespace** objects — their `.constructor` is `undefined`, so the chain breaks at step 1 — and they still run in the host realm, so features that internally use `new Function` (e.g. `math.evaluate`) keep working.
- Only the documented surface (`done`, `debug`, `user`, `documents`, `mathjs`, `formulajs`, `artifacts`, `sources`, `table`, `console`) is exposed.

## Verification

| Vector | Result |
|---|---|
| `process` / `global.process` / `globalThis.process` | blocked (before & after) |
| `require('fs')`, `eval()`, `new Function()`, `WebAssembly.compile` | blocked (before & after) |
| sandbox-created `(()=>{}).constructor('…')()` | blocked (before & after) |
| **`JSON.constructor.constructor('return process')()`** (+ Math/Date/Array/Object/String/Number/Boolean/RegExp/Map/Set/Promise) | **before: LEAKS outer `process.env` → after: blocked** |

Added `policy-service/tests/unit-tests/helpers/custom-logic-worker-sandbox.test.mjs`: it spawns the real worker with a poisoned env var and asserts 20+ escape payloads never leak it, plus a functional smoke suite confirming context-native `JSON`/`Math`, `mathjs.evaluate`, `formulajs`, and passed-in `documents` still work after the change.

## Notes

- Frontend/API unchanged; scoped to the JS worker + its new test.
- No dependency changes — this stays on `node:vm` (no native addon), so it does not introduce `isolated-vm` and cannot fault the policy-service process.

### DIFF
--- a/policy-service/src/policy-engine/helpers/workers/custom-logic-worker.ts
+++ b/policy-service/src/policy-engine/helpers/workers/custom-logic-worker.ts
@@ -1,33 +1,47 @@
 import { workerData, parentPort } from 'node:worker_threads';
 import * as mathjs from 'mathjs';
-import * as formulajs from '@formulajs/formulajs'
+import * as formulajs from '@formulajs/formulajs';
 import { buildTableHelper } from '../table-field-core.js';
 import * as vm from 'node:vm';
 
 /**
- * Execute user-supplied JavaScript in a sandboxed V8 context.
+ * Execute user-supplied JavaScript in a hardened `node:vm` context.
  *
- * Previously, user code ran via Function() in the same Node.js context as
- * the policy-service, giving access to process.env, require(), and the
- * full filesystem. The Python worker already uses Pyodide (WASM sandbox)
- * for isolation; this change brings the JavaScript worker to the same
- * security level using Node.js vm module with code generation disabled.
+ * Isolation model (TM-006 / M-005):
+ *  - The sandbox global is `Object.create(null)` and is contextified by
+ *    `vm.createContext`, so user code sees the CONTEXT's own native intrinsics
+ *    (`JSON`, `Math`, `Object`, `Array`, ...). We deliberately do NOT copy the
+ *    host realm's intrinsics onto the sandbox — that was the escape vector: a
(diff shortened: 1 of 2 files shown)


vshvets-bc:
Superseded by https://github.com/hashgraph/guardian/pull/6373 (cleaned up commit history and description). Closing this one.

PR #6402, opened 2026-07-21
perf: send a schema reference in get-block-data instead of the full schema

vshvets-bc:
## Problem
`requestVcDocumentBlock` and `requestVcDocumentBlockAddon` `getData()` embed the **full schema** (`document` + `context`) in every form block. For policies with large schemas the same schema is serialized once per form block, so `get-block-data` responses can reach **tens of MB**. That payload is `JSON.parse`/`JSON.stringify`-d as it is relayed between services (which do not read it), kept in memory on the client, and re-fetched on every block update — the repeated multi-MB serialization blocks the single Node event loop and slows unrelated requests.

## Change
Emit a lightweight schema **reference** `{ id, iri, uuid, name, version }` instead of the full `document`/`context`. The frontend resolves the full schema **once by id** (cached and de-duplicated) via the existing `GET /schema/:id`.

- Much smaller `get-block-data` responses for schema-heavy forms.
- **Backward-compatible:** the frontend uses a full schema (with `document`) exactly as before, so the client change is a no-op until the backend sends a reference. The change is sequenced so the frontend can ship first.

## Details
- **policy-service:** `request-vc-document-block.ts`, `request-vc-document-block-addon.ts` — `getData` returns the reference.
- **frontend:** `SchemaService.getSchemaById` + cached/de-duplicated/error-evicting `resolveSchemaById`; `request-document-block(-addon)` `_onSuccess` resolves a reference to the full schema before the existing render flow.

## Tests
- **policy-service** (new `request-vc-getdata-reference.test.mjs`): `getData` emits a reference, not `document`/`context`. Passing.
- **frontend** (new `schema.service.spec.ts`): endpoint, concurrent-subscriber de-dup, cache-hit, and error-eviction/retry.

## Notes for reviewers
- The frontend package was not built in this environment; the policy-service change and its test build and pass.
- Worth confirming there is no non-frontend consumer of `get-block-data` that relies on the embedded `document`/`context` (e.g. remote/multi-policy interaction). The `id`/`iri` are still present for anything that needs to look the schema up.

### DIFF
--- a/frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block-addon/request-document-block-addon.component.ts
+++ b/frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block-addon/request-document-block-addon.component.ts
@@ -17,6 +17,7 @@ import { DialogService } from 'primeng/dynamicdialog';
 import { AbstractUIBlockComponent } from '../models/abstract-ui-block.component';
 import { RequestDocumentBlockDialog } from '../request-document-block/dialog/request-document-block-dialog.component';
 import { SchemaRulesService } from 'src/app/services/schema-rules.service';
+import { SchemaService } from 'src/app/services/schema.service';
 import { prepareVcData } from 'src/app/modules/common/models/prepare-vc-data';
 import { PolicyTestAutomationService } from '../../policy-test-automation/policy-test-automation.service';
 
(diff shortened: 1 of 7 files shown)


frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block/request-document-block.component.ts
```diff
@@ -209,6 +211,24 @@ export class RequestDocumentBlockComponent
...
+        // before the render flow; a full schema is used as-is.
+        const schemaRef = data?.schema;
+        if (schemaRef && !schemaRef.document && schemaRef.id) {
+            this.loading = true;
+            this.schemaService.resolveSchemaById(schemaRef.id).subscribe({
+                next: (full) => {
+                    data.schema = full;
+                    this._applyBlockData(data);
+                },
+                error: () => this._applyBlockData(data)
```
Pyatakov: On resolution failure this renders with the bare reference schema. `new Schema(reference)` has no `document`, so `parseDocument()` is skipped (interfaces/src/models/schema.ts:234) and `fields` stays empty so the user gets a silently blank form. Route the failure through the existing error handler instead:
```suggestion
                error: (e) => this._onError(e)
```

frontend/src/app/modules/policy-engine/policy-viewer/blocks/request-document-block-addon/request-document-block-addon.component.ts
```diff
@@ -94,12 +95,30 @@ export class RequestDocumentBlockAddonComponent
...
+        // Resolve a schema reference (id, no document) to the full schema before setData.
+        const schemaRef = data?.schema;
+        if (schemaRef && !schemaRef.document && schemaRef.id) {
+            this.loading = true;
+            this.schemaService.resolveSchemaById(schemaRef.id).subscribe({
+                next: (full) => {
+                    data.schema = full;
+                    super._onSuccess(data);
+                },
+                error: () => super._onSuccess(data)
```
Pyatakov: Same issue as the non-addon block: on error this falls back to the reference schema (no `document`) and renders an empty form. Use `error: (e) => this._onError(e)` so the failure goes through the normal error/loading path.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6414, opened 2026-07-23
fix: mirror node rate-limit resilience - pace and back off topic listener polling, retry 429 in worker-service

vshvets-bc:
## Problem

The Hedera mirror node rate limit is **per source IP** and is shared by every process behind that IP — `worker-service`, `topic-listener-service`, an indexer, other tenants. Guardian currently assumes it has that budget to itself in two places, and neither degrades gracefully when it doesn't.

We hit this in a production deployment with 147 registered listeners: a **50-minute window of continuous HTTP 429s across 115 distinct topics**, during which the listener ingested nothing and kept retrying every throttled topic on every cycle. 1051 failures produced 6,306 log lines, because each one printed a full stack.

## 1. `topic-listener-service`: no pacing, no backoff

`ListenerService.scheduler()` polls every registered listener back to back with **no delay at all**, then sleeps 10s and repeats. `Listener.search()` catches, logs, and moves on without advancing `searchIndex`, so a throttled topic is retried immediately on the next cycle, at full rate.

- **Pace the calls.** A configurable delay between mirror-node requests (`LISTENER_CALL_DELAY_MS`, default 25ms ≈ 40 req/s) instead of none.
- **Back off on failure.** Honour `Retry-After` on 429 (seconds or HTTP date); otherwise exponential backoff 5s→5min with **downward-only** jitter, so the configured caps stay hard limits.
- **Back off when idle.** A topic that keeps returning nothing is polled progressively less often, 10s→60s. In a real deployment most listeners sit on topics that rarely or never produce a message — in ours, 110 of 147 had never received one — and they were being polled as often as active ones. The cap keeps a topic that *does* become active responsive.
- **Don't spend budget on listeners that aren't due.** `search()` now returns whether it actually called upstream; the scheduler only applies its pacing delay when a call happened, so a listener in backoff costs no cycle time.
- **Request timeout 2min → 15s** (`LISTENER_REQUEST_TIMEOUT_MS`). The loop is sequential, so one hung request stalled every other listener for up to two minutes.
- **One log line per failure** with the retry decision, instead of a full stack.

The loop stays sequential on purpose — parallelising would burst against the very limit this is protecting.

## 2. `topic-listener-service`: chunked messages could never complete

`Listener.addMessages()` returns before advancing `_searchIndex` for continuation chunks:

```ts
if (chunkId) {
    for (const m of this._messages) {
        if (m.chunkId === chunkId) {
            m.addChunk(message);
            return;          // <-- _searchIndex never advances past chunk 1
        }
    }
}
```

So the index sticks at the sequence number of the group's **first** chunk. Once `confirmData()` drops the completed message, the next poll (`sequencenumber=gt:<first chunk>`) re-fetches chunks 2..N. `Message.addChunk()` has no dedup, so `messages.length` overshoots `chunkTotal`, and `compressMessages()` only sets `COMPRESSED` on **exact** equality — the group flips to permanent `COMPRESSING`, `next()` never returns it, and the listener re-fetches and re-appends the same chunks every cycle while growing in memory.

Fixed by advancing the index for continuation chunks (safe: the constructor resumes from `min(searchIndex, sendIndex)`, so an incomplete in-memory group is re-fetched whole after a restart), ignoring re-fetched chunks, ordering chunks by sequence number before reassembly (concatenation depends on order), and letting the completeness check tolerate an over-full group.

## 3. `worker-service`: 429 was fatal

`axiosGetWithRetry()` retries 502/503/504 and network timeouts, but 429 fell straight through as fatal — a short throttling window failed the whole task. 429 joins the transient set, and when the response carries `Retry-After` the retry waits exactly that (capped at 60s) instead of the helper's own backoff. Every other 4xx still fails fast, so 404 behaviour is unchanged.

This covers both call sites: the mirror node fetch in `hedera-sdk-helper.ts` and the IPFS gateway fetch in `ipfs-client-class.ts`.

## Configuration

All new knobs have defaults matching current behaviour where behaviour is unchanged, and are env-overridable: `LISTENER_CALL_DELAY_MS`, `LISTENER_REQUEST_TIMEOUT_MS`, `LISTENER_ERROR_BACKOFF_MIN_MS`, `LISTENER_ERROR_BACKOFF_MAX_MS`, `LISTENER_IDLE_BACKOFF_STEP_MS`, `LISTENER_IDLE_BACKOFF_MAX_MS`.

## Tests

- `topic-listener-service/tests/listener-backoff.test.mjs` (new, 11 tests): idle and error backoff growth and caps, `Retry-After` parsing (seconds / HTTP date / unusable), skip-while-backing-off, chunk index advance, re-fetched chunk dedup, out-of-order reassembly. Suite: **38 passing**.
- `worker-service/tests/axios-get-with-retry.test.mjs` (new, 4 tests): driven against a local HTTP server — 429 retried to success, `Retry-After: 1` actually held the retry ~1s, 404 not retried, attempt cap respected. Suite: **283 passing** (`tests/*.test.mjs`).
- `tslint` clean on both services; `tsc` clean.

Note: `worker-service/tests/ipfs-client.test.mjs` fails to load on `develop` before this change — it imports `../dist/api/ipfs-client.js`, but the module is `ipfs-client-class.ts`. Unrelated to this PR, but it does mean `tests/**/*.test.mjs` cannot currently run to completion.

### DIFF
--- a/topic-listener-service/src/api/listener-service.ts
+++ b/topic-listener-service/src/api/listener-service.ts
@@ -11,6 +11,13 @@ export class ListenerService extends NatsService {
     public replySubject = 'listeners-queue-reply-' + GenerateUUIDv4();
 
     private readonly delay: number = 10 * 1000;
+    /**
+     * Delay between mirror node calls. The public mirror node allows ~50 req/s per
+     * source IP and that budget is shared with every other process behind the same
+     * IP (worker-service, indexer, ...), so the listener paces itself well below it.
+     */
+    private readonly callDelay: number =
+        parseInt(process.env.LISTENER_CALL_DELAY_MS, 10) || Math.ceil(1000 / 40);
     private readonly map: Map<string, Listener>;
 
     constructor(
@@ -111,7 +118,11 @@ export class ListenerService extends NatsService {
     public async scheduler(): Promise<void> {
         while (true) {
             for (const listener of this.map.values()) {
-                await listener.search();
+                const polled = await listener.search();
+                if (polled) {
+                    //only pace real mirror node calls - a listener in backoff costs nothing
+                    await new Promise(resolve => setTimeout(resolve, this.callDelay));
+                }
             }
             await new Promise(resolve => setTimeout(resolve, this.delay));
         }
--- a/topic-listener-service/src/api/listener-service.ts
+++ b/topic-listener-service/src/api/listener-service.ts
@@ -2,6 +2,7 @@ import { DatabaseServer, MessageResponse, NatsService, PinoLogger } from '@guard
 import { GenerateUUIDv4, IListenerOptions, ListenerEvents } from '@guardian/interfaces';
 import { TopicListener as ListenerCollection } from '../entity/index.js';
 import { Listener } from './listener.js';
+import { envNumber } from '../helpers/env.js';
 
 /**
  * Worker class
@@ -17,7 +18,7 @@ export class ListenerService extends NatsService {
      * IP (worker-service, indexer, ...), so the listener paces itself well below it.
      */
     private readonly callDelay: number =
-        parseInt(process.env.LISTENER_CALL_DELAY_MS, 10) || Math.ceil(1000 / 40);
+        envNumber('LISTENER_CALL_DELAY_MS', Math.ceil(1000 / 40));
     private readonly map: Map<string, Listener>;
 
     constructor(
--- a/topic-listener-service/src/api/listener.ts
+++ b/topic-listener-service/src/api/listener.ts
@@ -16,11 +16,31 @@ export class Listener {
 
     public static readonly REST_API_MAX_LIMIT: number = 100;
 
+    /**
+     * The mirror node rate limit is per source IP and is shared with every other
+     * process behind it, so a throttled listener has to yield instead of retrying
+     * at full rate.
+     */
+    public static readonly REQUEST_TIMEOUT_MS: number =
+        parseInt(process.env.LISTENER_REQUEST_TIMEOUT_MS, 10) || 15 * 1000;
+    public static readonly ERROR_BACKOFF_MIN_MS: number =
+        parseInt(process.env.LISTENER_ERROR_BACKOFF_MIN_MS, 10) || 5 * 1000;
+    public static readonly ERROR_BACKOFF_MAX_MS: number =
+        parseInt(process.env.LISTENER_ERROR_BACKOFF_MAX_MS, 10) || 5 * 60 * 1000;
+    public static readonly IDLE_BACKOFF_STEP_MS: number =
+        parseInt(process.env.LISTENER_IDLE_BACKOFF_STEP_MS, 10) || 10 * 1000;
+    public static readonly IDLE_BACKOFF_MAX_MS: number =
+        parseInt(process.env.LISTENER_IDLE_BACKOFF_MAX_MS, 10) || 60 * 1000;
+
     private _messages: Message[];
 
     private _searchIndex: number;
     private _sendIndex: number;
     private _subscription: Subscription;
+
+    private _errorCount: number;
+    private _idleCount: number;
+    private _nextPollAt: number;
 
     constructor(
         channel: ListenerService,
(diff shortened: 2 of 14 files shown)


topic-listener-service/src/api/listener.ts
Pyatakov: Remove the `console.log`; `search()` now logs the failure with topic + retry context.

topic-listener-service/src/api/listener-service.ts
```diff
@@ -11,6 +11,13 @@ export class ListenerService extends NatsService {
     public replySubject = 'listeners-queue-reply-' + GenerateUUIDv4();
 
     private readonly delay: number = 10 * 1000;
+    /**
+     * Delay between mirror node calls. The public mirror node allows ~50 req/s per
+     * source IP and that budget is shared with every other process behind the same
+     * IP (worker-service, indexer, ...), so the listener paces itself well below it.
+     */
+    private readonly callDelay: number =
+        parseInt(process.env.LISTENER_CALL_DELAY_MS, 10) || Math.ceil(1000 / 40);
```
Pyatakov: Env can't be set to `0`. `parseInt(process.env.X, 10) || default` treats `0` as falsy, so an operator who sets `LISTENER_CALL_DELAY_MS=0` to disable pacing silently gets 25ms. Applies to all six knobs – `listener-service.ts:19-20` and `listener.ts:24-33`. If `0` should be a valid value (I believe so), replace `|| default` with an explicit NaN check, e.g.
```
const n = parseInt(process.env.LISTENER_CALL_DELAY_MS, 10);
const callDelay = Number.isFinite(n) ? n : Math.ceil(1000 / 40);
```

topic-listener-service/src/api/listener.ts
```diff
@@ -98,6 +218,11 @@ export class Listener {
             for (const m of this._messages) {
                 if (m.chunkId === chunkId) {
                     m.addChunk(message);
+                    //continuation chunks must advance the index too, otherwise the next poll
+                    //re-fetches them forever and the group never completes.
+                    //Safe on restart: the constructor resumes from min(searchIndex, sendIndex),
+                    //so an incomplete in-memory group is re-fetched whole.
+                    this._searchIndex = message.sequence_number;
```
Pyatakov, line 224: This advance makes `_searchIndex` (last chunk's seq) diverge from `_sendIndex`, which is confirmed at the **first** chunk's seq – `sendData()` publishes `message.sequenceNumber`, and `Message.index` is set to `first.sequenceNumber`. So after a completed & confirmed 2-chunk message (seq 11 -> 12): `sendIndex=11`, `searchIndex=12`.

While running this is fine, but the constructor resumes from `min(searchIndex, sendIndex)` = **11**. On restart the next poll is `gt:11`, which re-fetches only chunk 2 -> a lone group stuck in `COMPRESSING` at `_messages[0]` -> `next()` returns `null` and every later message on that topic is blocked permanently.

So the "Safe on restart … re-fetched whole" comment holds for *incomplete* groups but not for a *completed & confirmed* one, where `sendIndex` lands mid-group and `gt:` skips the head chunk. This is newly exposed because chunk groups couldn't complete before this PR.

**Suggested fix:** in `addMessages`, treat a continuation chunk (`chunk_info.number > 1`) that matches no in-progress group as an already-consumed tail – advance `_searchIndex` and skip it instead of starting a new group (in normal forward paging chunk 1 always arrives first, so this only drops the orphan tail this scenario produces). Worth a test that constructs a listener with `searchIndex > sendIndex` and asserts it doesn't wedge.

Pyatakov (changes requested):
This PR adds six variables but no config file is updated, so they're undiscoverable. Please add them (commented, with the code defaults) to every config file under `topic-listener-service/configs`:
```
#LISTENER_CALL_DELAY_MS="25"          # delay between mirror-node calls (~40 req/s)
#LISTENER_REQUEST_TIMEOUT_MS="15000"  # per-request timeout
#LISTENER_ERROR_BACKOFF_MIN_MS="5000"
#LISTENER_ERROR_BACKOFF_MAX_MS="300000"
#LISTENER_IDLE_BACKOFF_STEP_MS="10000"
#LISTENER_IDLE_BACKOFF_MAX_MS="60000"
```

Additionally, the worker-service 429 change relies on `REST_API_RETRY_COUNT` / `REST_API_RETRY_DELAY`, which are also absent in templates at `worker-service/configs`. Pre-existing, but the same discoverability gap, so it'd be worth documenting them in the same pass.

Pyatakov (approved):

PR #6421, opened 2026-07-24
fix(api-gateway): cached @Response() endpoints hang on cache hit; preserve response headers

vshvets-bc:
## Problem

A `@UseCache()` route whose handler manages its own response via `@Response() res` (Fastify manual-response mode) **hangs on every cache hit**.

The api-gateway runs on Fastify, and these handlers send the body themselves via `res.send(...)`. In that mode Nest does not auto-send the interceptor's emitted value — it waits for `res.send()`.

On a cache **hit** the real handler never runs, so nobody calls `res.send()`. The `CacheInterceptor` only sends on the `isFastify` branch (`responseContext.send(...)`); without the flag it returns `of(value)`, which Fastify's `@Res` mode drops → the reply is never written → the request hangs until the client times out.

- First load (cache **miss**) → handler runs → `res.send()` → works.
- Reload within TTL (cache **hit**) → interceptor returns a value nobody sends → **hang**.

## Affected handlers

Six handlers use `@UseCache()` + `@Res` without `isFastify:true`, so each hangs on a cache hit:

| file | handler |
|---|---|
| `schema.ts` | `getSchemasPage`, `getSchemasPageByTopicId`, `getSystemSchema` |
| `contract.ts` | `getContracts` |
| `tool.ts` | `getTools` |
| `themes.ts` | `exportTheme` |

## Fix

**1.** Each handler is fixed to the existing working pattern (`getModuleSchemas`): `@Req() req` + `req.locals = <body>` + `@UseCache({ isFastify: true })`. `req.locals` is required — a `@Res` handler emits `undefined` to the interceptor, so without it the miss caches an empty body and the hit serves nothing.

**2.** The interceptor now **preserves response headers across a cache hit**. A hit replays only the cached body, so headers the client relies on were lost:
- **`X-Total-Count`** — paginated list endpoints (`observe: 'response'`) silently fell back to the current page length, undercounting pagination on cached responses.
- **`Content-Disposition` / `Content-Type`** — cached file downloads (`exportTemplate` is already `@UseCache({ isFastify: true })`) arrived with **no filename and wrong MIME** on cache hits.

The interceptor captures an allowlist (`X-Total-Count`, `Content-Disposition`, `Content-Type`) on the miss and re-applies them via `reply.header()` before `reply.send()` on the hit. Only allowlisted headers survive a hit (auth/set-cookie/etc. must not be replayed across users).

## Tests

- `cache-fastify-response.invariant.test.js` — fails if any `@UseCache` handler uses `@Response()`/`@Res` without `isFastify: true`; guards the six and any future regression.
- `cache-interceptor-headers.test.js` — `X-Total-Count` and download-header (`Content-Disposition`/`Content-Type`) capture-on-miss / replay-on-hit round-trips.

`tsc` build clean.

## Note

Cached file downloads are stored base64 in the cache store; fine for the small templates here, but caching large downloads through this JSON cache would be a memory-growth concern (pre-existing, out of scope).

### DIFF
--- a/api-gateway/tests/cache-interceptor-headers.test.js
+++ b/api-gateway/tests/cache-interceptor-headers.test.js
@@ -81,7 +81,7 @@ describe('CacheInterceptor header preservation (fastify)', () => {
         const sent = [];
         const req = { headers: {}, url: '/dl', locals: Buffer.from('zipbytes') };
         const res = {
-            getHeader: (k) => hdr[k],
+            getHeader: (k) => hdr[k.toLowerCase()],
             header: (k, v) => { applied[k] = v; },
             send: (x) => { sent.push(x); return 'SENT'; },
         };
--- a/api-gateway/tests/cache-interceptor-headers.test.js
+++ b/api-gateway/tests/cache-interceptor-headers.test.js
@@ -1,0 +1,103 @@
+import 'reflect-metadata';
+import assert from 'node:assert/strict';
+import { firstValueFrom, of } from 'rxjs';
+
+import { CacheInterceptor } from '../dist/helpers/interceptors/cache.js';
+
+function makeStore(initial = {}) {
+    const data = { ...initial };
+    const sets = [];
+    return {
+        data,
+        sets,
+        service: {
+            get: async (k) => (k in data ? data[k] : null),
+            set: async (k, v) => { sets.push({ k, v }); data[k] = v; },
+            addTags: async () => {},
+        },
+    };
+}
+
+function makeContext(request, response) {
+    const handler = function handlerFn() {};
+    return {
+        switchToHttp: () => ({ getRequest: () => request, getResponse: () => response }),
+        getHandler: () => handler,
+        getClass: () => function classFn() {},
+    };
+}
+
+function fastify(ctx) {
+    Reflect.defineMetadata('fastify', true, ctx.getHandler());
+}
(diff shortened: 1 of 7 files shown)


api-gateway/tests/cache-interceptor-headers.test.js
```diff
@@ -0,0 +1,103 @@
...
+        const store = makeStore();
+        const hdr = {
+            'Content-Disposition': 'attachment; filename=template.xlsx',
+            'Content-Type': 'application/zip',
+        };
+        const applied = {};
+        const sent = [];
+        const req = { headers: {}, url: '/dl', locals: Buffer.from('zipbytes') };
+        const res = {
+            getHeader: (k) => hdr[k],
```
Pyatakov: The mock `getHeader: (k) => hdr[k]` is a case-sensitive object lookup, so it doesn't reflect Fastify's case-insensitive `getHeader`. It passes only because the mock keys match the allowlist casing. If a handler ever set a header with different casing, production would still work but this test wouldn't be exercising that path. 
Suggestion:
```suggestion
            getHeader: (k) => hdr[k.toLowerCase()],
```
with lowercase keys to mirror real Fastify.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6447, opened 2026-07-27
fix: step block unavailable flash

nazlo-climission:
**Description**:

While a policy workflow moves from one step to the next, the API briefly reports the current block as gone (`Block Unavailable`) before a follow-up websocket update delivers the next block. The frontend applied that empty response immediately, so a normal transition was rendered as a real "nothing to show" state and the message *"This step isn't available to you right now"* flashed up before disappearing a moment later. This PR waits out that gap, so only a state that outlives a transition reaches the user.

* Delay applying an empty block result by 600 ms so a gap between two steps is never rendered
* Keep the spinner up during that gap instead of showing the unavailable message
* Cancel a pending empty commit as soon as a newer block response arrives
* Cancel the in-flight block request when a newer reload starts, so a stale response cannot overwrite a newer one
* Treat `422 Block Unavailable` as "not your turn" alongside `503`
* Clear the pending timer on destroy and on retry
* Add unit tests covering the transition gap, the genuine unavailable state, and the error mapping

**Notes for reviewer**:

**Root cause.** `policy-service` raises `Block Unavailable` as `503`, but `api-gateway` remaps block-data errors to `422` - its own swagger documents this: *"the API responds with `422 Unprocessable Entity` and `message: "Block Unavailable"`"*. `_onError` routed the empty result straight into `setData(null)`, which applied it synchronously. Nothing distinguished a 300 ms gap between two steps from a genuine role/state gate.

Worth a look as a second, independent bug: `_onError` only checked `503`, so a genuine "not your turn" arriving as `422` fell through to `hasError` and rendered the red *"Couldn't load this step"* error card instead of the intended message. Happy to split this out of the PR if preferred.

**Behaviour, measured in the DOM** by simulating the documented contract (`422` at t=0, next block at t=300 ms):

| | t=0 | t=60ms | t=200ms | t=300ms |
|---|---|---|---|---|
| **Before** | message painted, opacity 1, 203px | opacity 1 | opacity 1 | removed |
| **After** | not in DOM | not in DOM | not in DOM | not in DOM |

Counter-check - a genuine unavailable state (nothing arrives afterwards) still displays correctly: `unavailable` goes `false → false → true` across the delay boundary and renders the expected message.

**Manual repro**, with the frontend running and a policy open on a step block (DevTools console):

```js
const cmp = ng.getComponent(document.querySelector('step-block'));
const real = { blocks: cmp.blocks, index: cmp.index, readonly: cmp.readonly };

cmp._onError({ status: 422, error: { message: 'Block Unavailable' } });  // transition starts
setTimeout(() => console.log('message showing?', cmp.unavailable), 200); // expect false
setTimeout(() => cmp.setData(real), 300);                                // next block arrives
```

Two gotchas when checking by hand: reload between runs (the component keeps its state, so a second run starts already-empty and looks like a failure), and keep the tab focused - backgrounded tabs clamp `setTimeout` to ~1 s, which stretches the 300 ms gap past the 600 ms window and fakes a regression.

**Tests are mutation-checked**, since passing tests prove nothing on their own. Reverting the fix to an immediate commit turns the suite red:

```
with fix:       13 specs, 0 failures
fix reverted:   13 specs, 7 failures
```

This caught a weak assertion along the way: a spec that checked only the end state passed even with the bug present, so it now samples throughout the gap.

**Not verified.** The transition is simulated at the component boundary - I did not drive a real role handoff (submitting as one role so the workflow advances to a step owned by another), because the dry-run policy available to me had not advanced that far. The DOM measurements above were taken against a running build of this component rather than this repo's frontend, which I did not start. Unit tests were executed under Node with Angular's decorators stubbed, as Chrome was unavailable in my environment; they exercise the real component logic but not template rendering, so a `yarn test` run under karma is worth doing before merge.

**Checklist**
- [x] Documented (Code comments, README, etc.) - inline comments explaining why the empty result is delayed and why both `503` and `422` are handled
- [x] Tested (unit, integration, etc.) - `step-block.component.spec.ts`, mutation-checked

### DIFF
--- a/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.ts
+++ b/frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.ts
@@ -16,6 +16,16 @@ import { HttpErrorResponse } from '@angular/common/http';
 })
 export class StepBlockComponent implements OnInit {
     private socket: Subscription | null;
+    // In-flight block request. A step transition emits a burst of websocket
+    // updates, each triggering a reload; we keep only the latest so a stale
+    // intermediate response can't overwrite the UI and flash the "unavailable"
+    // message before the real block arrives.
+    private dataSub: Subscription | null = null;
+    // How long an "empty" result must persist before it is applied. A step
+    // transition briefly reports no block; anything shorter than this is a gap
+    // between two steps, not a real "not your turn" state.
+    private static readonly EMPTY_COMMIT_DELAY_MS = 600;
+    private emptyTimer: any;
 
     get loading(): boolean {
         // Only spin until the first response has been processed. Previously this
@@ -79,6 +89,8 @@ export class StepBlockComponent implements OnInit {
         if (this.socket) {
             this.socket.unsubscribe();
         }
+        this.dataSub?.unsubscribe();
+        clearTimeout(this.emptyTimer);
     }
 
     onUpdate(blocks: string[]): void {
@@ -91,13 +103,16 @@ export class StepBlockComponent implements OnInit {
         if (this.static) {
             this._onSuccess(this.static);
         } else {
-            this.policyEngineService
+            // Cancel any request still in flight so only the latest reload applies.
+            this.dataSub?.unsubscribe();
+            this.dataSub = this.policyEngineService
                 .getBlockData(this.id, this.policyId, this.savepointIds)
                 .subscribe(this._onSuccess.bind(this), this._onError.bind(this));
         }
     }
 
     retry() {
+        clearTimeout(this.emptyTimer);
         this.loaded = false;
         this.hasError = false;
         this.loadData();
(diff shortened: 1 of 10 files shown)


Pyatakov (changes requested):
@nazlo-climission It would be great if `request-document-block-dialog.component.html` and `request-document-block-dialog.component.scss` can be excluded from the PR, since these changes are unrelated to the issue this PR address.

Pyatakov (dismissed):

frontend/src/app/modules/policy-engine/policy-viewer/blocks/step-block/step-block.component.ts
```diff
@@ -87,55 +129,72 @@ export class StepBlockComponent implements OnInit {
...
+        // status is a genuine failure: stop the spinner but show a distinct error
+        // state so a real outage isn't disguised as a normal workflow message.
+        const blockUnavailable = e.status === 503 ||
+            (e.status === 422 && e.error?.message === 'Block Unavailable');
+        if (blockUnavailable) {
             this._onSuccess(null);
         } else {
             this.hasError = true;
             this.loaded = true;
         }
```
Pyatakov, line 176: `_pending` is not cleared when entering the error state – the spinner and the error card render at the same time.
```suggestion
        } else {
            this._pending = false;
            this.hasError = true;
            this.loaded = true;
        }
```
Please also add a spec: component in `pending` state, then `_onError({ status: 500, … })`, asserting `errored === true` and `loading === false`.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6482, opened 2026-07-28
fix: 6423 surface unreachable IPFS data on remote policy/tool import

vshvets-bc:
## Issues

Addresses #6423 (unhelpful error when a remote import fails on missing IPFS data). Related to #6424 (IPFS pinning durability) - this PR handles the *reporting* side, not pinning.

## Problem

Importing a remote policy or tool by Hedera message id fails with no meaningful feedback when the referenced IPFS data is unpinned or offline. `MessageServer.getMessage()` catches every failure and returns `null`, so callers cannot distinguish "message not found on Hedera" from "message found but its IPFS payload is unreachable":

- Policy import dereferences the `null` and throws a generic error.
- Tool import reports a generic "Invalid Message".
- On the frontend, a failed push-import task is swallowed on the full-page `/task/:id` route (it redirects without showing anything).

## Changes

- **`common`**: new `MessageIpfsError` (HTTP code `422`). `getMessage()` now wraps the IPFS load step and throws it instead of collapsing to `null` when the message header was retrieved but its documents could not be loaded. A genuinely missing message still returns `null`, so the existing contract is preserved.
- **`guardian-service`**: forward `error.code` across NATS in the policy and tool import handlers (sync + async), so the `422` reaches the client instead of a generic `500`. (Matches the many handlers that already do `new MessageError(error, error?.code)`.)
- **`frontend`**: `AsyncProgressComponent.setError()` surfaces the failure via `InformService.processAsyncError` (which shows the task code + message) before redirecting, so push-task failures are no longer silently swallowed. This covers all push-task errors, not only import.

## Result

- Unreachable IPFS -> `422 IPFS_UNAVAILABLE: IPFS data for message <id> could not be retrieved. It may be unpinned or the source node is offline.`
- Message genuinely absent -> unchanged not-found behaviour.
- The message is shown at the preview step (where the IPFS fetch first happens) and on import execution if it fails there.

## Tests

- `common`: `MessageIpfsError` class; `getMessage()` IPFS-failure handling (throws on IPFS load failure, returns null on non-IPFS failure / empty id, skips the IPFS step when `loadIPFS=false`, returns the loaded message on success).

### DIFF
--- a/frontend/src/app/modules/common/async-progress/async-progress.component.ts
+++ b/frontend/src/app/modules/common/async-progress/async-progress.component.ts
@@ -383,7 +383,7 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
         }
 
         // Show the failure before the redirect branches below navigate away.
-        this.informService.processAsyncError(error);
+        this.toastService.processAsyncError(error);
 
         if (this.last) {
             this.redirect(this.last);
--- a/frontend/src/app/modules/common/async-progress/async-progress.component.ts
+++ b/frontend/src/app/modules/common/async-progress/async-progress.component.ts
@@ -382,6 +382,9 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
             return;
         }
 
+        // Show the failure before the redirect branches below navigate away.
+        this.informService.processAsyncError(error);
+
         if (this.last) {
             this.redirect(this.last);
             return;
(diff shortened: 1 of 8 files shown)


frontend/src/app/modules/common/async-progress/async-progress.component.ts
```diff
@@ -382,6 +382,9 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
             return;
         }
 
+        // Show the failure before the redirect branches below navigate away.
+        this.informService.processAsyncError(error);
```
Pyatakov: `this.informService` does not exist, needs to be replaced with:
```suggestion
        this.toastService.processAsyncError(error);
```

Pyatakov (changes requested):

Pyatakov (commented):
Approved

PR #6498, opened 2026-07-30
feat: Address External Integration Challenges with Policy Filter #5797

gayanath8:
### Issue Labels
[https://github.com/hashgraph/guardian/issues/5797](https://github.com/hashgraph/guardian/issues/5797)

### Requirements

- Investigate and design a solution that allows external systems to discover and execute actions defined within policy grid containers (e.g., approve, reject) without needing to decode and replicate the full policy logic.
- External consumers should be able to:
- Retrieve the list of available actions for a given filtered dataset/grid.
- Execute those actions via API on specific records.
- The solution should not require external integrators to have deep knowledge of the internal policy block structure.
- Note: This item requires further discussion to determine the optimal approach, as it involves exposing policy workflow logic through APIs without breaking the abstraction.

### Summary of Changes

 Four new REST endpoints were added to the existing Policy API controller under /policies/:policyId/grids, exposing Guardian's policy block engine to external API callers. Access is role-scoped at the policy engine level each endpoint returns only the grids and actions the authenticated user's role is permitted to see.
 
Full API reference: [Policy Grid Actions docs](docs/guardian/standard-registry/policies/policy-grid-actions/README.md)

**1. List Grids (GET /api/v1/policies/:policyId/grids)**

- Authorization: Bearer <accessToken> required missing or invalid returns 401
- Account must hold POLICIES_POLICY_EXECUTE or POLICIES_POLICY_MANAGE otherwise 403
- policyId must be a running policy instance not found returns 404
- Results are filtered by the caller's policy role inaccessible grids are silently excluded
- Roles with no accessible grids receive 200 with an empty array

**Request:**
```
GET /api/v1/policies/6a3cdae37da45039a883bb36/grids
Authorization: Bearer <accessToken>
```

**Response `200 OK`:**
```json
[
  {
    "gridId": "registry_MRs",
    "title": "Monitoring Reports",
    "filterSchema": [],
    "columnSchema": [
      { "name": "", "title": "Action1", "type": "block", "bindActions": ["bb_issueToken"] },
      { "name": "", "title": "Action2", "type": "block", "bindActions": ["final_mint_button"] }
    ]
  }
]
```

- Returns all grids the caller's role can access in the policy
- Each grid descriptor includes its stable gridId, column schema, and filter schema
- Grids the caller's role cannot reach are silently excluded no error is returned
- Use the gridId values from this response in all subsequent calls
- Read-only grids (no action blocks) are included they just return [] from /actions

image
image

**2. List Actions for a Grid (GET /api/v1/policies/:policyId/grids/:gridId/actions)**

- Authorization: Bearer <accessToken> required missing or invalid returns 401
- Account must hold POLICIES_POLICY_EXECUTE or POLICIES_POLICY_MANAGE otherwise 403
- gridId must exist in the policy not found returns 404
- Caller's role must have access to the grid inaccessible grid returns 503
- Actions requiring roles the caller does not hold are silently excluded from the response
- Read-only grids (no action blocks bound to columns) return 200 with an empty array

**Request:**
```
GET /api/v1/policies/6a3cdae37da45039a883bb36/grids/registry_MRs/actions
Authorization: Bearer <accessToken>
```

**Response `200 OK`:**
```json
[
  {
    "actionId": "btn_reject",
    "label": "Reject",
    "requiredRoles": ["OWNER"],
    "appliesTo": "row",
    "inputSchema": { "type": "object", "properties": {} }
  },
  {
    "actionId": "final_mint_button",
    "label": "Submit Final Amount",
    "requiredRoles": ["ANY_ROLE"],
    "appliesTo": "row",
    "inputSchema": {
      "type": "object",
      "required": ["document"],
      "properties": {
        "document": {
          "type": "object",
          "description": "Credential subject conforming to schema #c9efd555-0f02-4301-9b0b-792ec7edda90&1.0.0"
        }
      }
    }
  },
  {
    "actionId": "assign_vvb",
    "label": "VVB",
    "requiredRoles": ["PROJECT_DEVELOPER"],
    "appliesTo": "row",
    "inputSchema": {
      "type": "object",
      "required": ["value"],
      "properties": {
        "value": {
          "description": "Value to set on field \"assignedTo\"",
          "enum": ["did:hedera:testnet:xxxxxx...VVB1"],
          "options": [{ "name": "Xeno VVB", "value": "did:hedera:testnet:xxxxxx...VVB1" }]
        }
      }
    }
  }
]
```

- Returns all row-level actions bound to columns in the specified grid
- Actions the caller's role is not permitted to execute are silently excluded
- Covers all supported action block types: selector options, button blocks, dropdown blocks, and request-VC-document blocks
- download-type action blocks are excluded they are not workflow actions
- `inputSchema` reflects what the Execute Action body must contain for that specific action, selector/button actions have an empty schema (`{}` body), dropdown actions require a `value` (with the currently valid choices listed under `enum`/`options`), and request-VC-document actions require a `document` object matching the named schema
- Read-only grids return 200 with an empty array, not an error
- Use the actionId values from this response in the execute call

image

**3. List Records in a Grid (GET /api/v1/policies/:policyId/grids/:gridId/records)**

- Authorization: Bearer <accessToken> required missing or invalid returns 401
- Account must hold POLICIES_POLICY_EXECUTE or POLICIES_POLICY_MANAGE otherwise 403
- gridId must exist in the policy not found returns 404
- Caller's role must have access to the grid inaccessible grid returns 503
- pageSize values above 200 are clamped to 200; invalid values default to 25
- Records are scoped by the grid's documentsSourceAddon the caller only sees documents their role's addon exposes

**Request:**
```
GET /api/v1/policies/6a3cdae37da45039a883bb36/grids/registry_MRs/records?page=1&pageSize=20
Authorization: Bearer <accessToken>
```

**Response `200 OK`:**
```json
{
  "data": [
    {
      "_id": "6627f3b2e4b0c1a2b3c4d5e6",
      "owner": "did:hedera:testnet:z...",
      "option": { "status": "Submitted" },
      "_actions": ["btn_reject", "final_mint_button", "assign_vvb"]
    }
  ],
  "page": 1,
  "pageSize": 20,
  "totalCount": 1
}
```

- Returns the paginated document set for the grid, filtered by the grid's own source addon scope
- Each record is annotated with _actions the action IDs the caller may execute on that row
- Each record includes a history array showing past document states when a historyAddon is present on the grid
- Use _id (not id) as the recordId in the execute call id is the VC @id (urn:uuid:...) and will return 404
- pageSize is capped at 200; values above that are clamped silently

image

**4. Execute an Action on a Record (POST /api/v1/policies/:policyId/grids/:gridId/records/:recordId/actions/:actionId)**

- Authorization: Bearer <accessToken> required missing or invalid returns 401
- Account must hold POLICIES_POLICY_EXECUTE or POLICIES_POLICY_MANAGE otherwise 403
- gridId must exist in the policy not found returns 404
- actionId must be bound to a column in the resolved grid not found returns 404
- recordId must match a document visible to the caller through the grid's source addon scope not found returns 404
- Caller's role must satisfy the action block's permissions inaccessible action returns 503
- Body must be valid JSON, and its required shape depends on the action's `inputSchema` (from List Actions) ,it is not always `{}`
  - Selector / button actions (e.g. Approve, Reject) — `{}` is sufficient
  - Dropdown actions (e.g. assigning a VVB) — require `{ "value": "<one of inputSchema.enum>" }`; missing `value` returns `400`
  - Request-VC-document actions (e.g. "Submit Final Amount") — require `{ "document": { ...fields... } }`; missing `document` returns `400`
  
  
**Request (selector/button):**
```
POST /api/v1/policies/6a3cdae37da45039a883bb36/grids/registry_MRs/records/6627f3b2e4b0c1a2b3c4d5e6/actions/btn_reject
Authorization: Bearer <accessToken>
Content-Type: application/json

{}
```

**Request (dropdown):**
```
POST /api/v1/policies/6a3ce9e57da45039a883c076/grids/projectDocuments_pd/records/6627f3b2e4b0c1a2b3c4d5e6/actions/assign_vvb
Authorization: Bearer <accessToken>
Content-Type: application/json

{ "value": "did:hedera:testnet:xxxxxx...VVB1" }
```

**Request (request-VC-document):**
```
POST /api/v1/policies/6a3cdae37da45039a883bb36/grids/registry_MRs/records/6627f3b2e4b0c1a2b3c4d5e6/actions/final_mint_button
Authorization: Bearer <accessToken>
Content-Type: application/json

{ "document": { "finalMintAmount": 1500 } }
```

**Response `200 OK`:**

- Executes a single row-level action on the specified record ,POST is required because this mutates server state
- The server resolves `gridId` → grid block, `actionId` → action block + option value, `recordId` → document internally; no block UUIDs are ever exposed
- The caller must be able to see the record through the grid's source addon scope the server verifies this before executing
- For **selector** actions: the bound field (e.g. `option.status`) is updated directly in the database; a history entry is recorded
- For **button** actions: the block's event chain is triggered by the button's tag; a history entry is recorded with the current document state
- For **dropdown** actions: the submitted `value` is written to the block's bound field (e.g. `assignedTo`), then the block's `DropdownEvent` chain is triggered; a history entry is recorded
- For **request-VC-document** actions: the submitted `document` fields are used to mint a brand-new signed VC referencing the record via `ref` **this is not idempotent**, each call mints a new VC
- Response body is always `{}` re-fetch `/records` to observe the field change (or the new VC, for request-VC-document actions)

image
image
image
image

### Unit Tests

`policy-service/tests/unit-tests/blocks/grid-action-resolver.test.mjs` covers `computeGridId`, `buildGridDescriptor`, `resolveGridBlock`, `resolveAction`, `getGrids`, `getActions`, and `getRecords` across all four supported action-block types (selector, button, dropdown, request-VC-document), plus permission/availability/error-code edge cases.

image
image
image


api-gateway/src/api/service/policy.ts
```diff
@@ -8329,4 +8329,383 @@ export class PolicyApi {
...
+    async getGridRecords(
+        @AuthUser() user: IAuthUser,
+        @Param('policyId') policyId: string,
+        @Param('gridId') gridId: string,
+        @Query() query: any,
+    ): Promise<any> {
+        try {
+            const params: any = {};
+            if (query.page !== undefined) params.page = parseInt(query.page, 10);
+            if (query.pageSize !== undefined) params.pageSize = Math.min(parseInt(query.pageSize, 10), 200);
```
Pyatakov: ### `page/pageSize` NaN not handled
`parseInt` on non-numeric input yields `NaN` (and `Math.min(NaN, 200)` stays `NaN`), and negative/zero values aren't floored, so `page`/`pageSize` can be passed through as `NaN`. The docs promise a default of 25 and 1-based paging; clamp explicitly, e.g. `Number.isFinite(n) ? Math.min(Math.max(n, 1), 200) : DEFAULT`.

policy-service/src/policy-engine/external/grid-action-resolver.ts
```diff
@@ -0,0 +1,472 @@
...
+            actionBlock as IPolicyInterfaceBlock, user
+        );
+        if (actionError) {
+            throw Object.assign(new Error(actionError.message), { code: actionError.code });
+        }
+
+        // 4. Look up the document directly by ID — access is already gated by
+        //    isAvailableGetData (grid) and isAvailableSetData (action block) above.
+        const ref = (gridBlock as any);
+        const document = await ref.databaseServer.getVcDocument({
```
Pyatakov, line 446: ### Record-scope authorization gap
The record is looked up by `id` + `policyId` only, so any document in the policy can be targeted – `isAvailableSetData`/`isAvailableGetData` only check `block.isAvailable(user)`, not the grid's `documentsSourceAddon` scope. A caller can execute an action on records the grid would never show them. Resolve the record through the grid's source-addon query (as `getRecords` does) and reject `recordId`s outside that filtered set before executing.

policy-service/src/policy-engine/external/grid-action-resolver.ts
```diff
@@ -0,0 +1,472 @@
...
+            throw Object.assign(new Error(error.message), { code: error.code });
+        }
+
+        const result = await PolicyComponentsUtils.blockGetData(gridBlock as IPolicyInterfaceBlock, user, params);
+        const data = result.body;
+
+        if (Array.isArray(data?.data)) {
+            const actions = await GridActionResolver.getActions(policyId, gridId, user);
+            const actionIds = actions.map((a) => a.actionId);
+            data.data = data.data.map((doc: any) => ({ ...doc, _actions: actionIds }));
```
Pyatakov, line 402: ### `_actions` is grid-level, not row-level
`_actions` is computed once at grid level and copied onto every record, but the API docs describe it as "the action IDs the caller may execute on **that row**." If action availability can vary per record, this over-reports. Either evaluate actions per row, or correct the docs to state `_actions` is grid-level.

policy-service/src/policy-engine/block-tree-generator.ts
```diff
@@ -382,6 +383,63 @@ export class BlockTreeGenerator extends NatsService {
...
+            } catch (error) {
+                return new MessageError(error.message, error.code || 500);
+            }
+        });
+
+        this.getPolicyMessages(PolicyEvents.EXECUTE_GRID_ACTION, policyId, async (msg: any) => {
+            try {
+                const { user, gridId, recordId, actionId, body, syncEvents, history } = msg;
+                const userFull = await this.getUser(policyInstance, user);
+                const resolution = GridActionResolver.resolveAction(policyId, gridId, actionId);
```
Pyatakov, line 424: ### Redundant `resolveAction` + `undefined`-block closure
`resolveAction` runs here and again inside `executeAction` – duplicate work. When `actionId` is unknown this yields `resolution = null`, and the `RecordActionStep` closure captures `resolution?.actionBlock` as `undefined`; it's only saved by `executeAction` throwing 404 first. Resolve once and pass the result into `executeAction`, or guard the null case explicitly.

policy-service/src/policy-engine/block-tree-generator.ts
```diff
@@ -382,6 +383,63 @@ export class BlockTreeGenerator extends NatsService {
...
+                const result = await GridActionResolver.getRecords(policyId, gridId, userFull, params || {});
+                return new MessageResponse(result);
+            } catch (error) {
+                return new MessageError(error.message, error.code || 500);
+            }
+        });
+
+        this.getPolicyMessages(PolicyEvents.EXECUTE_GRID_ACTION, policyId, async (msg: any) => {
+            try {
+                const { user, gridId, recordId, actionId, body, syncEvents, history } = msg;
```
Pyatakov: ### `history` is a dead read
`history` is destructured from `msg` but the upstream `EXECUTE_GRID_ACTION` sender never includes it, so it's always `undefined`. Drop the read or wire `history` through the gateway/guardian-service intentionally.

Pyatakov (changes requested):

api-gateway/src/api/service/policy.ts
gayanath8: Fixed the issue now non-finite value never reaches the clamp

policy-service/src/policy-engine/external/grid-action-resolver.ts
gayanath8: executeAction now verifies the record is actually in the caller's scoped grid results (via blockGetData, paged in bounded POLICY_DATA_MAX_PAGE_SIZE chunks) before executing, instead of trusting id+policyId alone. Distinguishes "confirmed not found" (404) from "couldn't finish scanning" (500) if a grid ever exceeds the safety ceiling.

policy-service/src/policy-engine/external/grid-action-resolver.ts
gayanath8: corrected get-records.md and the Swagger schema to describe _actions accurately as a grid-level catalog,

policy-service/src/policy-engine/block-tree-generator.ts
gayanath8: resolves once and returns 404 immediately if unresolvable, before ever building RecordActionStep, executeAction takes the pre-resolved result as an optional param and only re-resolves if called without one.

policy-service/src/policy-engine/block-tree-generator.ts
gayanath8: dropped from the handler's destructure

Pyatakov (approved):

Issue #6546, opened 2026-08-05
Array field path resolution and dual-source conditions in documentValidatorBlock


dariy-rised:
### Problem description
The documentValidatorBlock conditions system has two limitations:
  1. Top-level conditions can only compare a document field against a static literal. No field-to-field comparison within the same document.
  2. Field path resolution does not work over array fields. Paths through an array return the array object itself rather than mapping the remainder of the path over each element, making array-typed schema fields unusable in condition paths.

  ### Requirements
  - Top-level conditions must support a Value Type selector.
  - Field path resolution used by all condition types must correctly traverse array fields.

  ### Definition of done
  - [ ] Top-level condition UI shows a Value Type field (Value / Document Field).
  - [ ] resolveFieldPath work with arrays at every intermediate segment.
  - [ ] checkDocumentField (top-level) and resolveDocumentValue / resolveSourceValue (cross-doc) both use resolveFieldPath.
  - [ ] Existing policy configs without valueSource set continue to use the legacy code path unchanged.

 ###  Acceptance criteria
  - A condition with field = "items.value" where items is an array correctly evaluates against every element in that array.
  - A top-level condition field A equal field B passes when both fields hold the same value and fails when they differ.
  - Policies created before this change validate identically to before.

PR #6548, opened 2026-08-05
feat: 6546 Array field path resolution and dual-source conditions in documentValidatorBlock

dariy-rised:
**Description**:
- Add array path resolution, coerces both condition sides, and resolves document-field references in error messages.

**Related issue(s)**:
Resolves [#6546](https://github.com/hashgraph/guardian/issues/6546)


policy-service/tests/unit-tests/helpers/check-document-field.test.mjs
Pyatakov, line 1: The duplicate test file. `policy-service/tests/unit-tests/blocks/policy-utils-field-condition.test.mjs` alongside the new additions to `helpers/check-document-field.test.mjs` – both files now declare `describe('PolicyUtils.resolveFieldPath')`, `describe('PolicyUtils.evaluateFieldCondition')` and `describe('PolicyUtils.firstFailingPair')`, one with `chai` and one with `node:assert/strict`, with heavily overlapping cases. They don't contradict each other, but mocha will run both. Delete the `blocks/` file, folding in the few cases it uniquely covers (`10 equal '10'` coercion, `[1,2] in [2,1]` membership, the legacy strict-type pin).

policy-service/src/policy-engine/helpers/utils.ts
```diff
@@ -439,46 +439,210 @@ export class PolicyUtils {
...
+        if (leftIsArray && rightIsArray) {
+            if (left.length !== right.length) { return false; }
+            return left.every((l: any, i: number) =>
+                PolicyUtils.evaluateFieldCondition(l, type, right[i])
+            );
+        }
+        if (leftIsArray) {
+            return left.every((l: any) => PolicyUtils.evaluateFieldCondition(l, type, right));
+        }
+        return PolicyUtils.compareScalarPair(left, type, right);
```
Pyatakov, line 558: Scalar left vs non-empty array right now falls through to `compareScalarPair` for every non-`in` operator, so the result comes from JS array-to-primitive coercion rather than a defined rule:
```
5 gt [3]        -> true    // [3] becomes "3"
5 gt [3,4]      -> false   // "3,4" becomes NaN
5 gt [9]        -> false
5 equal [5]     -> false   // strict !==
5 not_equal [5] -> true
```
`gt` sees through a one-element array, `equal` does not, and any array of length 2 or more silently fails through `NaN`. The test suite pins only `1 equal [1] === false` and labels it a type mismatch, so the relational asymmetry is untested and undocumented.

Suggest making it explicit: either return `false` for scalar vs array on non-`in` operators, or restore a defined for-all rule over the right side.

policy-service/src/policy-engine/blocks/document-validator-block.ts
```diff
@@ -304,19 +245,18 @@ export class DocumentValidatorBlock {
...
             const counted = new Set<string>();
             for (let ci = 0; ci < conditions.length; ci++) {
                 const condition = conditions[ci];
-                const coerceDeep = (v: any) => Array.isArray(v) ? v.map((e: any) => this.coerceValue(e)) : this.coerceValue(v);
-                const left  = coerceDeep(this.resolveConditionSide(condition.field, condition.fieldSource, condition.type, document, [sourceDoc]));
-                const right = coerceDeep(this.resolveConditionSide(condition.value, condition.valueSource, condition.type, document, [sourceDoc]));
-                if (!this.evaluateCrossCondition(left, condition.type, right)) {
+                const left  = this.resolveConditionSide(condition.field, condition.fieldSource, condition.type, document, [sourceDoc]);
+                const right = this.resolveConditionSide(condition.value, condition.valueSource, condition.type, document, [sourceDoc]);
+                if (!PolicyUtils.evaluateFieldCondition(left, condition.type, right)) {
```
Pyatakov, line 201: Routing cross-source conditions through `PolicyUtils.evaluateFieldCondition` changes `in` semantics for policies already in production. Compared with the deleted `evaluateCrossCondition`:

```
tags=['a','b'] in 'a'            old: true   new: false
tags=['a','b'] in ['a','b','c']  old: false  new: true
unknown / missing type           old: true   new: false
```

The old evaluator asked "does the field array contain the value" (`left.includes(right)`); the new one applies for-all over the CSV list. The first case flips a passing validation to failing on deploy. Same-doc conditions are protected by the `valueSource` gate, but this path has none, and no test exercises `runSourceValidation` at runtime.

Suggest either gating the cross path the same way, or keeping the change and adding runtime tests plus a release note.

policy-service/src/policy-engine/blocks/document-validator-block.ts
```diff
@@ -150,14 +110,14 @@ export class DocumentValidatorBlock {
     }
 
     private resolveDocumentValue(path: string, document: IPolicyDocument): any {
-        return PolicyUtils.getObjectValue(document, path);
+        return PolicyUtils.resolveFieldPath(document, path);
```
Pyatakov, line 51: `resolveDocumentValue` now uses `resolveFieldPath`, which also affects `buildSourceFilter` (lines 178 and 225). A `typeValue: 'variable'` filter whose path crosses an array used to resolve to `undefined` and now resolves to an array, so the generated query changes from `{$eq: undefined} to {$eq: [1,2]}` – different documents are selected for existing policies. `coerceValue` also only coerces one level, so nested arrays reach the query uncoerced. No test covers the filter-building path.

Pyatakov (changes requested):

policy-service/src/policy-engine/helpers/utils.ts
dariy-rised: fixed `left-scalar vs right-array`
kept `left-array vs right-scalar`
added comments and tests for this case

policy-service/src/policy-engine/blocks/document-validator-block.ts
dariy-rised: I added tests.

Pyatakov (approved):

PR #6553, opened 2026-08-06
fix: enhance task dispatch handling and payload limits in queue service

nazlo-climission:
**Description**:

This PR hardens `queue-service` task dispatch so a single oversized or undeliverable task can no longer degrade or stall the whole queue, porting and adapting the fix from `Managed-Guardian-Service#1552` (including its two review follow-up commits) to this repo's simpler polling-based dispatch loop.

* Add `TASK_MAX_PAYLOAD` (default 256MB) and reject task payloads above it at creation (`TaskEntity`), instead of accepting a task that can never be delivered
* Record `dataSize` on every task at creation to scale the dispatch timeout and aid diagnostics without re-serializing the payload
* Add `dispatchAttempt` (`TaskEntity` / `ITask`) to count failed hand-offs to a worker separately from `attempt` (worker-reported business failures)
* Dead-letter a task (`isError: true` + `TASK_COMPLETE` publish) once `WORKER_DISPATCH_MAX_ATTEMPTS` (default 3) dispatch attempts fail, instead of retrying forever
* Order the dispatch query FIFO by `createDate` and exclude `done`/`isError` tasks, so an undeliverable task no longer pins the head of the queue
* Claim a task before dispatching it and dispatch without awaiting inside the worker loop, so one slow/undeliverable payload can't block dispatch to every other worker
* Scale the worker ack timeout with payload size (`WORKER_SEND_TASK_TIMEOUT_PER_MB` / `_MAX`) instead of a flat timeout sized for inline payloads only
* Release (not dead-letter) a claim when a worker reports `{ result: false }` (busy), and re-read the task by id before writing any dispatch-state change, so a stale in-memory snapshot can't revert a task that completed while the send was still in flight
* Reorder the `idx_dispatch_queue` index fields per Equality-Sort-Range so the dispatch query's sort is served by the index instead of an in-memory sort stage
* Wrap `refreshAndReassignTasks` in `try/finally` so an unhandled error can no longer permanently stick the reassignment loop
* Add/extend unit test coverage for all of the above

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

- Source PR (`Managed-Guardian-Service#1552`) was written against a heavily modified fork of `queue-service` (semaphore-based dispatch, `WORKER_READY` events, tenant context). This repo's `queue-service` uses a simpler interval-polling loop with no timeout on worker sends at all, so the fix was re-implemented against that architecture rather than cherry-picked — the goal (bounded dispatch retries, no head-of-line blocking, scaled ack timeout) is the same.
- `helm/queue-service/values-dev.yaml` from the source PR was not ported — this repo has no Helm charts (that's ops-repo specific).
- `tsc` build is clean across `interfaces`, `common`, and `queue-service`; full `queue-service` test suite passes (49/49), including new suites `task-entity-max-payload.test.mjs` and `queue-service-dispatch.test.mjs`.

**Checklist**

### DIFF
--- a/common/src/mq/nats-service.ts
+++ b/common/src/mq/nats-service.ts
@@ -255,7 +255,11 @@ export abstract class NatsService {
         const timeoutPromise = new Promise<T>((_, reject) => {
             setTimeout(() => {
                 this.responseCallbacksMap.delete(messageId);
-                reject(new Error(`Timeout exceed (${subject})`));
+                // Tagged so callers can tell "no ack within the deadline" apart from a transport
+                // error: a timeout does not prove the message was never delivered.
+                const error: any = new Error(`Timeout exceed (${subject})`);
+                error.code = 'DISPATCH_TIMEOUT';
+                reject(error);
             }, timeout);
         });
 
--- a/common/src/mq/nats-service.ts
+++ b/common/src/mq/nats-service.ts
@@ -378,8 +378,15 @@ export abstract class NatsService {
      * @param subject
      * @param cb
      * @param noRespond
+     * @param beforeDecode Optional synchronous hook run right after auth, before `codec.decode`.
+     * `codec.decode` can HTTP-fetch an out-of-band `directLink` payload, which for a large task
+     * takes far longer than authenticating the message - a caller that needs to claim some piece
+     * of state per-message (e.g. a busy flag) has to do it here, not in `cb`, or a second message
+     * arriving mid-decode observes the stale pre-claim state. Return a value to short-circuit:
+     * skip `cb`/decode entirely and respond with it (or, if `noRespond`, just skip). Return
+     * `undefined` to proceed normally.
      */
-    public getMessages<T, A>(subject: string, cb: Function, noRespond = false): Subscription {
+    public getMessages<T, A>(subject: string, cb: Function, noRespond = false, beforeDecode?: () => unknown): Subscription {
         this.addAdditionalAvailableEvents([subject]);
         return this.connection.subscribe(subject, {
             queue: this.messageQueueName,
@@ -417,6 +424,17 @@ export abstract class NatsService {
                     } catch (err) {
                         throw err;
                     }
+
+                    if (beforeDecode) {
+                        const early = beforeDecode();
+                        if (early !== undefined) {
+                            if (!noRespond) {
+                                msg.respond(await this.codec.encode(early), { headers: head });
+                            }
+                            return;
+                        }
+                    }
+
                     if (!noRespond) {
                         msg.respond(await this.codec.encode(await cb(await this.codec.decode(msg.data), msg.headers)), { headers: head });
                     } else {
--- a/common/src/mq/nats-service.ts
+++ b/common/src/mq/nats-service.ts
@@ -255,10 +255,9 @@ export abstract class NatsService {
         const timeoutPromise = new Promise<T>((_, reject) => {
             setTimeout(() => {
                 this.responseCallbacksMap.delete(messageId);
-                // Tagged so callers can tell "no ack within the deadline" apart from a transport
-                // error: a timeout does not prove the message was never delivered.
+                // Reuse REQUEST_TIMEOUT - same "no ack" code requestOrThrow uses below.
                 const error: any = new Error(`Timeout exceed (${subject})`);
-                error.code = 'DISPATCH_TIMEOUT';
+                error.code = 'REQUEST_TIMEOUT';
                 reject(error);
             }, timeout);
         });
@@ -385,17 +384,27 @@ export abstract class NatsService {
      * arriving mid-decode observes the stale pre-claim state. Return a value to short-circuit:
      * skip `cb`/decode entirely and respond with it (or, if `noRespond`, just skip). Return
      * `undefined` to proceed normally.
+     * @param onError Runs if auth/decode/cb throws before a response was sent, so a stranded
+     * `beforeDecode` claim can be released. `claimed` is true only for this message's own claim.
      */
-    public getMessages<T, A>(subject: string, cb: Function, noRespond = false, beforeDecode?: () => unknown): Subscription {
+    public getMessages<T, A>(
+        subject: string,
+        cb: Function,
+        noRespond = false,
+        beforeDecode?: () => unknown,
+        onError?: (error: unknown, claimed: boolean) => void
+    ): Subscription {
         this.addAdditionalAvailableEvents([subject]);
         return this.connection.subscribe(subject, {
             queue: this.messageQueueName,
             callback: async (error, msg) => {
+                let head: ReturnType<typeof headers> | undefined;
+                let claimed = false;
                 try {
                     const messageId = msg.headers?.get('messageId');
                     const serviceToken = msg.headers?.get('serviceToken');
                     // const isRaw = msg.headers.get('rawMessage');
-                    const head = headers();
+                    head = headers();
                     if (messageId) {
                         head.append('messageId', messageId);
                     }
@@ -433,6 +442,7 @@ export abstract class NatsService {
                             }
                             return;
                         }
+                        claimed = true;
                     }
 
                     if (!noRespond) {
@@ -442,6 +452,29 @@ export abstract class NatsService {
                     }
                 } catch (error) {
                     console.error(error);
+
+                    // Release a stranded beforeDecode claim and reply with an error.
+                    if (onError) {
+                        try {
+                            onError(error, claimed);
+                        } catch (hookError) {
+                            console.error(hookError);
+                        }
+                    }
+
+                    if (!noRespond && head) {
+                        try {
+                            await msg.respond(await this.codec.encode({
+                                body: null,
+                                error: error instanceof Error ? error.message : String(error),
+                                code: 500,
+                                name: 'Error',
+                                message: error instanceof Error ? error.message : String(error)
+                            }), { headers: head });
+                        } catch (respondError) {
+                            console.error(respondError);
+                        }
+                    }
                 }
             }
         });
--- a/queue-service/configs/.env.queue
+++ b/queue-service/configs/.env.queue
@@ -11,6 +11,18 @@ DB_HOST="localhost"
 MQ_MAX_PAYLOAD="1048576"
 #MQ_MESSAGE_CHUNK=5000000
 
+# Task payload limits and dispatch retry budget.
+# TASK_MAX_PAYLOAD - hard ceiling (bytes); a larger task is rejected at creation. NOT the same as
+#   MQ_MAX_PAYLOAD: payloads above MQ_MAX_PAYLOAD are legitimate and travel out-of-band via
+#   LargePayloadContainer. Keep in sync with the worker memory limit.
+# WORKER_SEND_TASK_TIMEOUT_PER_MB / _MAX - ack timeout scaling for out-of-band payloads, which the
+#   worker only acks after fetching and parsing in full.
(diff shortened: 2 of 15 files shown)


queue-service/src/queue-service/queue-service.ts
```diff
@@ -213,24 +241,173 @@ export class QueueService extends NatsService {
...
+
+                // Claim the task before attempting to send it, and dispatch without awaiting
+                // here: refreshAndReassignTasks is single-flight (guarded by `this.trigger`), so a
+                // single slow or undeliverable payload would otherwise hold up every other worker
+                // in this batch - and, on the next tick, the whole queue.
+                task.processedTime = new Date();
+                task.sent = true;
+                await dataBaseServer.save(TaskEntity, task);
+
+                this.dispatchClaimedTask(worker, task, payload, timeout).catch((error) => {
```
Pyatakov: Dispatching without awaiting removes the only thing that limited a worker to one in-flight send. A worker sets `isInUse` at handler entry (`worker-service/src/api/worker.ts:186`), which is *after* `ZipCodec.decode` has HTTP-fetched a `directLink` payload, and `GET_FREE_WORKERS` is a separate subscription, so during an 80MB delivery the worker keeps reporting itself free on every cycle (~300ms: `REFRESH_INTERVAL=50` plus the 300ms window in `getFreeWorkers`). Each cycle claims another task and ships it to the same worker; those tasks are then invisible to genuinely free workers.

Worse, `getMessages` uses the callback form of `subscribe` (`common/src/mq/nats-service.ts:378`), which nats.js does not await, so a small task decoded in the meantime wins the `isInUse` race and the 80MB task is rejected with `{result:false}` after a full download. That goes down the busy path at line 304, which releases the claim *without* incrementing `dispatchAttempt`. The task returns to head-of-FIFO and is downloaded again, forever – the "retries forever" failure this PR fixes, in a new form.

Suggest a `Set<string>` of worker subjects with a dispatch in flight: add before `dispatchClaimedTask`, remove in a `finally`, and skip those workers in the loop.

queue-service/src/queue-service/queue-service.ts
```diff
@@ -213,24 +241,173 @@ export class QueueService extends NatsService {
...
+                return;
+            }
+            if (r && r.result === false) {
+                // `{ result: false }` means the worker's SEND_TASK_TO_WORKER handler found
+                // `isInUse` true - it was busy, not a failed hand-off. This happens routinely:
+                // a worker reported free by getFreeWorkers() (up to 300ms + one refresh tick
+                // earlier) can already be occupied by the time the send actually reaches it.
+                // Routing this through releaseOrParkTask would burn the dead-letter budget on
+                // healthy tasks under normal load, so just release the claim.
+                task.processedTime = null;
```
Pyatakov: This path writes the caller-held snapshot instead of re-reading, unlike `releaseOrParkTask` and `resetDispatchAttempt`. The comment implies a brief window, but `{result:false}` only arrives after the worker's decode – for an out-of-band payload that's the whole download, i.e. minutes. Not a demonstrated data-loss bug today (the task is claimed, so nothing races it), but worth routing through a shared `releaseClaim(taskId)` helper for consistency.

queue-service/src/queue-service/queue-service.ts
```diff
@@ -213,24 +241,173 @@ export class QueueService extends NatsService {
...
+    /**
+     * Roll a failed hand-off back into the queue, or dead-letter the task once it has exhausted
+     * its dispatch budget.
+     *
+     * `attempt` is only ever incremented in the WorkerEvents.TASK_COMPLETE handler, i.e. only on
+     * errors *reported by a worker*. A task that never reaches a worker at all - a dispatch
+     * timeout or transport error - would otherwise retry forever without ever being marked
+     * failed. `dispatchAttempt` gives that case its own bounded budget.
+     *
+     * NOTE: a send timeout does not prove the task was not delivered, so re-queueing can in
```
Pyatakov: "predates this change" isn't accurate. `sendMessage` had no timeout, and `processedTime` was set only on success, so the only sub-hour re-dispatch was `clearLongPendingTasks` at `processTimeout` = 1h. With a 5s default ack timeout, any stalled ack now re-queues the task while the first worker may still be running it – a duplicated Hedera mint/transfer.

Consider not re-queueing on *timeout* specifically (dead-letter, or leave it to `clearLongPendingTasks`) while still re-queueing on transport errors, which are provably undelivered.

queue-service/src/queue-service/queue-service.ts
```diff
@@ -11,6 +11,25 @@ export class QueueService extends NatsService {
     private readonly refreshInterval = parseInt(process.env.REFRESH_INTERVAL, 10) || 1 * 1000; // 1s
     private readonly processTimeout = parseInt(process.env.PROCESS_TIMEOUT, 10) || 1 * 60 * 60000; // 1 hour
 
+    private readonly workerSendTaskTimeout = parseInt(process.env.WORKER_SEND_TASK_TIMEOUT, 10) || 5 * 1000; // 5s
```
Pyatakov, line 14: `WORKER_SEND_TASK_TIMEOUT` is new. The env block in `configs/.env.queue` documents `_PER_MB`, `_MAX` and `WORKER_DISPATCH_MAX_ATTEMPTS` but omits this one, which is the variable that changes behaviour for every task. Please add it.

queue-service/configs/.env.queue
```diff
@@ -11,6 +11,18 @@ DB_HOST="localhost"
...
 
+# Task payload limits and dispatch retry budget.
+# TASK_MAX_PAYLOAD - hard ceiling (bytes); a larger task is rejected at creation. NOT the same as
+#   MQ_MAX_PAYLOAD: payloads above MQ_MAX_PAYLOAD are legitimate and travel out-of-band via
+#   LargePayloadContainer. Keep in sync with the worker memory limit.
+# WORKER_SEND_TASK_TIMEOUT_PER_MB / _MAX - ack timeout scaling for out-of-band payloads, which the
+#   worker only acks after fetching and parsing in full.
+# WORKER_DISPATCH_MAX_ATTEMPTS - failed hand-offs before the task is dead-lettered.
+#TASK_MAX_PAYLOAD="268435456"
+#WORKER_SEND_TASK_TIMEOUT_PER_MB="2000"
```
Pyatakov, line 24: Missing `#WORKER_SEND_TASK_TIMEOUT="5000"` — see the comment on `queue-service.ts:14` (identical blocks in `.env.queue.develop` / `.env.queue.template`).

Pyatakov (changes requested):

common/src/mq/nats-service.ts
```diff
@@ -255,7 +255,11 @@ export abstract class NatsService {
         const timeoutPromise = new Promise<T>((_, reject) => {
             setTimeout(() => {
                 this.responseCallbacksMap.delete(messageId);
-                reject(new Error(`Timeout exceed (${subject})`));
+                // Tagged so callers can tell "no ack within the deadline" apart from a transport
+                // error: a timeout does not prove the message was never delivered.
+                const error: any = new Error(`Timeout exceed (${subject})`);
+                error.code = 'DISPATCH_TIMEOUT';
```
Pyatakov: ### Minor
`requestOrThrow` in this same file already uses `REQUEST_TIMEOUT` / `NO_RESPONDERS` for the same distinction. A third code for "timed out" is worth avoiding – consider reusing `REQUEST_TIMEOUT`.

queue-service/src/queue-service/queue-service.ts
```diff
@@ -201,36 +237,230 @@ export class QueueService extends NatsService {
...
+                // earlier) can already be occupied by the time the send actually reaches it.
+                // Routing this through releaseOrParkTask would burn the dead-letter budget on
+                // healthy tasks under normal load, so just release the claim.
+                await this.releaseClaim(task.taskId);
+                return;
+            }
+            console.log('task sent error');
+            await this.releaseOrParkTask(task, `Task was rejected by the worker: ${JSON.stringify(r)}`);
+        } catch (error) {
+            if (error?.code === 'DISPATCH_TIMEOUT') {
```
Pyatakov: ### Blocker
This relies on `clearLongPendingTasks` reclaiming the task, but its predicate can't do that. `common/src/helpers/db-helper.ts:544` matches on `processedTime - createDate > processTimeout` – how long the task waited *before* being claimed, not how long it has been claimed. Both operands are frozen at claim time (`createDate` is never rewritten; this branch returns without touching `processedTime`), so the delta is a constant that never grows.

Result depends on when the claim happened, and neither case is good:
- Claimed within `processTimeout` of creation (normal): never matches. The task sits `sent: true` / `processedTime` set / `done: false` / `isError: false` – excluded from the dispatch query, ignored by `clearLongPendingTasks`, and `clearOldTasks` needs `done: true`. Stuck permanently; the caller hangs until its own deadline.
- Claimed later than that (backed-up queue): matches on every claim thereafter, so it's released every 30s, re-dispatched, times out, and repeats indefinitely.

In both cases `dispatchAttempt` is never incremented here, so an ack timeout can never reach the dead-letter budget, and an ack timeout is the dominant symptom of the undeliverable task this PR targets. The "retries forever" stall is traded for a silent permanent stall.

Suggest keeping the claim (correct – no immediate re-dispatch, so no duplicate execution) but still incrementing `dispatchAttempt` and dead-lettering once exhausted. That bounds the failure while preserving at-most-once. If reclaim-after-timeout is also wanted, the predicate needs to compare `processedTime` against *now*, not `createDate`.

Pyatakov:
@nazlo-climission any update here?

nazlo-climission:
> @nazlo-climission any update here?

yep, pls check [this commit](https://github.com/hashgraph/guardian/pull/6553/changes/45a88f40f9fe745188f4eb4cca56c62d8e5ab033)

Pyatakov:
@nazlo-climission, the last commit you're referring to addresses the previous set of comments, but there are still two unresolved. Could you please review them?

worker-service/src/api/worker.ts
```diff
@@ -224,31 +225,39 @@ export class Worker extends NatsService {
...
+        // SEND_TASK_TO_WORKER or the direct-dispatch path below) can win the race and get
+        // accepted first, and the original task is then busy-rejected after downloading in full.
+        // `beforeDecode` runs synchronously right after auth and before decode starts, so
+        // whichever message's callback resumes first claims `isInUse` atomically - the claim no
+        // longer races the decode.
+        const claimIfFree = (): MessageResponse<{ result: boolean }> | undefined => {
+            if (this.isInUse) {
+                return new MessageResponse({ result: false });
+            }
+            this.isInUse = true;
```
Pyatakov, line 233: ### Blocker
**`isInUse` is claimed before decode and never released if decode throws**

New regression, introduced by the latest commit. `getMessages` awaits `codec.decode` inside its `try` (`nats-service.ts:439`) and the `catch` only logs (`:444`), so `runTask` is the only path back to `isInUse = false` (`:225`, or `clearState` `:304`) and it never runs. The worker stays busy for the rest of its lifetime.

To reproduce, run the PR-head callback and `claimIfFree` verbatim against a rejecting decode, comparing old head vs new head with the same input:

```
old: isInUse after failed decode = false  (worker still accepts tasks: true)
new: isInUse after failed decode = true   (worker still accepts tasks: false)
```

No reply is sent either, so the task is stranded alongside the worker. `ZipCodec.decode` throws on any `directLink` fetch failure or `JSON.parse` OOM, reachable in exactly the large-payload case this PR targets.

nazlo-climission:
> unresolved

please check latest commits, everything should be fixed

Pyatakov (approved):

PR #6624, opened 2026-08-11
feat: make tests idempotent; fixing tests; introducing eslinter

Neurone:
## Summary

Makes the Cypress e2e suite idempotent and safe to re-run in any order/against a reused environment, fixes a handful of tests and one real backend bug found along the way, and adds an ESLint setup for `e2e-tests`. Most of the changed-line count comes from the linter's `--fix` pass (quote-style, `no-var`, etc.) rather than from behavior changes — see the dedicated section below.

## Idempotent tests

Several specs previously depended on state that only existed once: a fixed name/tag, a value in `list[0]`, a binary fixture exported months ago, or a Hedera message ID that testnet/IPFS had since aged out. Any of that made a second run (or a run against a non-fresh DB) fail with "already exists", a stale array index, or `IPFS_UNAVAILABLE`. Fixed the whole suite for this class of issue, using a few repeated patterns:

- **Unique per-run identifiers.** Names, tags and UUIDs that used to be hardcoded constants (e.g. `policyTag: "Tag_11223344"`) are now suffixed/generated with `randomInt`/`crypto.randomUUID()` per run, so re-running a spec never collides with what a previous run left behind (`nestedSchemasDeletion.cy.js`, `007_modules/*`, `012_analytics/compareModules.cy.js`, tags specs, …).
- **Async task polling.** Added `support/CustomHelpers/tasks.js` (`waitForTask`): several endpoints (e.g. schema/tag deletion) only return a `taskId` and say nothing about the outcome. Tests used to assert right after the response and read pre-task state. `waitForTask` polls `tasks/{taskId}` until it errors or completes.
- **Live IPFS/Hedera seeding instead of stale message IDs.** Added `support/CustomHelpers/ipfsSeeding.js` (`publishPolicyFixture`, `publishSchema`, `importPolicyFromMessage`): publishes a policy/schema fixture live and hands back the fresh Hedera message ID, with retry/backoff on `IPFS_UNAVAILABLE` while the gateway catches up. Replaces hardcoded message IDs in test config, which rot as pinned IPFS content gets evicted or testnet gets reset.
- **On-demand fixtures instead of positional/leftover state.** New `support/api/modules.js` and `support/api/accounts.js`, plus new `commands.js` helpers (`getOrCreateRetireContractId`, `getOrCreateModuleId`, …): specs build or resolve the entity they need (e.g. `resolveDraftModule`) instead of assuming `at(0)`/`at(1)` of a list or relying on an entity created by an earlier pool/spec. This is also what decouples the `007_modules` suite from run order.
- **Dynamic export instead of binary fixtures.** Removed stale binary fixtures (`exportedModule.module`, `importDraftModule.module`, `importPublichedModule.module`, `module_*.module`) in favor of exporting the module/policy at test time and feeding that back into import/compare assertions.
- **Standalone-runnable specs.** `commands.js` "get or create" helpers let specs that used to reuse whatever a previous pool left behind run on their own or against a fresh DB, without changing what they assert.
- **Default password change.** The shared demo/test password changed from `test` to `TestPass1` (auth-service fixtures, frontend demo login, `guardian-cli`, `demia` scripts, `cypress.env.json`) since `test` no longer satisfies password-strength validation and was silently breaking login/register flows in tests and demo data.
- **README rewrite** documenting the idempotency guarantee, the one-time `preparing`-tag prerequisite to seed accounts/balances, and reorganized Docker/local run instructions.

## Bug found while seeding fixtures live

Exercising the "import from a freshly published Hedera message" path (via `ipfsSeeding.js`) surfaced that module import-from-message endpoints collapsed every failure to a generic 500. Fixed in `guardian-service/module.service.ts` / `api-gateway/module.ts`: errors now forward the real `error.code` via `loadErrorCode`, so a missing Hedera message correctly returns 404 (`MESSAGE_NOT_FOUND`) and a missing IPFS payload returns 422, instead of both surfacing as 500. Swagger docs updated to match.

Also uncommented `MAX_TRANSACTION_FEE` in `configs/.env.quickstart.guardian.system`, which was left disabled.

## Linter

Added ESLint (flat config, `eslint.config.mjs`) with `eslint-plugin-cypress` for `e2e-tests`, and ran `--fix` across the suite. This accounts for the bulk of the diff in `cypress/e2e/**` and `cypress/support/**` (mostly single-quote strings, `no-var`, object-shorthand) — no behavioral changes there.

## Checklist


guardian-service/src/api/module.service.ts
Pyatakov, line 1: I'd recommend splitting this into two PRs. Keep the changes to this file and `api-gateway/src/api/service/module.ts` in a separate PR for better traceability, since these changes aren't directly related to the e2e test issues this PR addresses.

e2e-tests/cypress/support/CustomHelpers/ipfsSeeding.js
```diff
@@ -0,0 +1,183 @@
...
+            expect(response.status, `import of ${fixtureName}`).to.eq(STATUS_CODE.SUCCESS);
+            const policyId = JSON.parse(new TextDecoder().decode(response.body)).at(0).id;
+            return cy.request({
+                method: METHOD.PUT,
+                url: `${API.ApiServer}${API.Policies}${policyId}/${API.Publish}`,
+                body: { policyVersion },
+                headers: { authorization },
+                timeout: PUBLISH_TIMEOUT,
+            }).then((publishResponse) => {
+                expect(publishResponse.status, `publish of ${fixtureName}`).to.eq(STATUS_CODE.OK);
```
Pyatakov, line 43: Policy publish answers `200` with `isValid: false` and stays in `DRAFT` (`policy-engine.service.ts:1658-1674`), so a failed validation passes this check and then fails later as `expected undefined to match /^\d+\.\d+$/`. Assert `body.isValid`.

e2e-tests/cypress/e2e/api-tests/000_accounts_tests/postRegister.cy.js
```diff
@@ -128,26 +134,27 @@ context("Register", { tags: ['accounts', 'firstPool', 'all'] }, () => {
...
+            password_confirmation: credentials.altPassword,
             role: 'USER',
         }).then((response) => {
             expect(response.status).to.eq(STATUS_CODE.UNPROCESSABLE);
             expect(response.body.message).to.deep.eq(['Passwords must match']);
         });
     });
 
-    // password validation sets to easy + 4 length for CICD, so we cannot get weak password
+    // password validation sets to easy + 8 length for CICD, so we cannot get weak password
```
Pyatakov: This assertion expects the MEDIUM/8 message, but every workflow running the API suite starts auth-service with `MIN_PASSWORD_LENGTH: 4` and `PASSWORD_COMPLEXITY: easy`:
- `api-after-commit.yml:147`
- `api-manual.yml:155`
- `api-schedule-all.yml:148`
- `api-schedule-vm0033.yml:148`
- `ui-manual.yml:152`

This causes the test to fail. 

The comment above also says "easy + 8" while the workflows set 4. Matching `/^Password must be at least \d+ characters long/` keeps the test agnostic to the password policy.

e2e-tests/cypress/support/commands.js
```diff
@@ -147,6 +147,134 @@ Cypress.Commands.add('getPolicyLabels', (token) => {
...
+                },
+                timeout: 180000,
+            }).then(() => cy.request({
+                // Schema listings are cached by request URL for 10 minutes, so a plain listing
+                // straight after the creation keeps serving the pre-creation (empty) view.
+                // A unique query parameter (ignored by the API) gives an uncached read.
+                method: METHOD.GET,
+                url: API.ApiServer + API.Schemas + policy.topicId,
+                headers: { authorization },
+                qs: { cacheBust: `${randomInt(999999)}_${Date.now()}` },
```
Pyatakov: `cacheBust` workaround can be safely dropped here and in 5(?) other places, since the root cause was fixed in #6634

Pyatakov (changes requested):

e2e-tests/cypress/support/commands.js
Neurone: Removed 4 of the 6: schema listing-after-create and tool listing are now covered by #6634's tag-prefix cache invalidation, including the tools one — it only ever failed because tags used to carry the query string.

Two had to stay, for different reasons:

- `getAccounts.cy.js`: `api-gateway/src/api/service/account.ts` has **no** cache invalidation at all while `GET /accounts` is `@UseCache()` for 10 minutes, so registering a user genuinely leaves the listing stale. Updated the comment to explain why and linked a follow-up issue for giving accounts routes the same tag-prefix invalidation schemas/tools already have.
- `nestedSchemasDeletion.cy.js`, but only for the two tests that delete a **single** schema by id (`DELETE /schemas/{id}`), not the topic-wide delete test. Tracing this one found a real, separate bug: that endpoint is task-based on both sides — `guardian-service`'s handler kicks off the actual deletion in its own fire-and-forget `RunFunctionAsync` and acks immediately (`schema.service.ts:1968`), and `api-gateway` invalidates the schema cache right after that quick ack (`schema.ts:1586`) — i.e. *before* the real deletion has necessarily finished. A read in that window gets cached with the pre-deletion list for the full 10-minute TTL, and nothing invalidates it again afterward, so `waitForTask` correctly resolving doesn't help — the cache can already be stuck on stale data by then. Filed as a separate issue (linked in the code comment); until it's fixed, an uncached read is the only way to get the true post-single-delete state without an intermittent, hard-to-reproduce CI failure.

e2e-tests/cypress/e2e/api-tests/000_accounts_tests/postRegister.cy.js
Neurone: Went with deriving the exact expected message from the configured policy rather than a generic pattern — a regex like `/^Password must be at least \d+ characters long/` would pass even if the length in the message were wrong. Added `passwordPolicy.js`, a small mirror of `auth-service/src/constants/password.ts`, and switched `postRegister.cy.js`, `postUpdatePassword.cy.js`, and the UI `homePage.js` page object to assert the exact message it produces. Also wired `MIN_PASSWORD_LENGTH`/`PASSWORD_COMPLEXITY` through to Cypress as `CYPRESS_minPasswordLength`/`CYPRESS_passwordComplexity` in all 5 workflows (they were previously only visible to the Guardian containers, not the test runner), so CI's `easy`/4 config and local's default `medium`/8 both produce a correct assertion instead of a hardcoded string that only happened to match locally.

e2e-tests/cypress/support/CustomHelpers/ipfsSeeding.js
Neurone: Added the `isValid` assertion to `publishPolicyFixture` in `ipfsSeeding.js`, surfacing `errors` in the failure message so a validation failure doesn't just show up later as a confusing `expected undefined to match /^\d+\.\d+$/`. Schema publish doesn't return `isValid` in its response contract (it returns the schema listing), so `publishSchema` now asserts `status === 'PUBLISHED'` on the matching schema instead, mirroring the existing pattern in `putSchemaPublish.cy.js`. Also fixed a stale comment in that function claiming the topic-filtered schema listing is cached while the per-topic route isn't — after #6634 both are invalidated on schema mutations.

e2e-tests/cypress/support/commands.js
Neurone: Also opened related issues for the remaining issues: https://github.com/hashgraph/guardian/issues/6660 https://github.com/hashgraph/guardian/issues/6661

e2e-tests/cypress/e2e/api-tests/000_accounts_tests/postRegister.cy.js
Neurone: I also modified the CI files to pass the password configs to the script, although they still continue using easy/4

guardian-service/src/api/module.service.ts
Neurone: Split here: https://github.com/hashgraph/guardian/pull/6663

Pyatakov (approved):

PR #6640, opened 2026-08-12
fix: 6622 Condition fields order

dariy-rised:
**Description**:
* Fixed condition fields order
* Fixed unreachable conditions in Else block (if required)

**Related issue(s)**:

Fixes [#6622](https://github.com/hashgraph/guardian/issues/6622)

### DIFF
--- a/common/src/hedera-modules/vcjs/vcjs.ts
+++ b/common/src/hedera-modules/vcjs/vcjs.ts
@@ -5,7 +5,7 @@ import { Ed25519Signature2018 } from '@digitalbazaar/ed25519-signature-2018';
 import { Ed25519VerificationKey2018 } from '@digitalbazaar/ed25519-verification-key-2018';
 import { PrivateKey } from '@hiero-ledger/sdk';
 import { SchemaValidationResult } from './schema-validation-result.js';
-import { GenerateUUIDv4, ICredentialSubject, ISchemaArrayDependency, IVC, Schema, SchemaHelper, SignatureType } from '@guardian/interfaces';
+import { GenerateUUIDv4, ICredentialSubject, ISchemaArrayDependency, IVC, Schema, SchemaField, SchemaHelper, SignatureType } from '@guardian/interfaces';
 import { VcDocument } from './vc-document.js';
 import { VpDocument } from './vp-document.js';
 import { VcSubject } from './vc-subject.js';
@@ -439,7 +439,7 @@ export class VCJS {
 
         for (const condEntry of schema.allOf) {
             if (!condEntry?.if) { continue; }
-            for (const branch of [condEntry.then, SchemaHelper.unwrapConditionElse(condEntry.else)]) {
+            for (const branch of [condEntry.then, condEntry.else]) {
                 collectPath(branch, [], rootProperties);
             }
         }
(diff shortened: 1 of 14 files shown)


common/src/hedera-modules/vcjs/vcjs.ts
```diff
@@ -702,15 +703,38 @@ export class VCJS {
...
+     * The schema document declares each branch but does not mark its fields required:
+     * JSON Schema applies `else` whenever `if` fails, including when the field the `if`
+     * reads was never asked, which would demand fields the form never showed. Those rules
+     * are enforced here instead, in the same shape ajv reports.
+     * @param subject credential subject
+     * @param schemaObject parsed schema
+     */
+    private static conditionErrors(subject: any, schemaObject: Schema): ArrayGroupValidationError[] {
+        return SchemaHelper
+            .validateConditionFields(schemaObject?.conditions || [], subject)
```
Pyatakov: `schemaObject.conditions` holds only the root schema's conditions – a sub-schema's conditions live on the ref field (`field.conditions`), so this call never sees them. Combined with `buildSub` no longer emitting `required`, a required field of a condition branch inside a sub-schema is now enforced by nothing: with `#Sub` = `if A == 1 then B (required)` referenced through `$defs`, `{x: 'v', child: {A: '1'}}` was rejected by ajv on develop (`#Sub/allOf/0/then/required`) and passes verification here. The form still marks the field required, so only API/XLSX/imported submissions slip through – which is exactly the path that matters for issued VCs.

Recurse into sub-schema conditions before merging the results, e.g. walk `schemaObject.fields` for `isRef` fields and run `validateConditionFields(field.conditions, subject[field.name])` (repeating per entry for `isArray`), prefixing the produced `instancePath` with the field path.

Pyatakov (commented):
**Must fix before merge**
* `common/src/hedera-modules/vcjs/vcjs.ts:728` – required fields of a sub-schema's condition branch are no longer enforced anywhere

**Should fix**
* `interfaces/src/helpers/schema-helper.ts:455` – unwrap shim targets an else shape no released version ever produced

interfaces/src/helpers/schema-helper.ts
```diff
@@ -438,6 +443,302 @@ export class SchemaHelper {
...
+    /**
+     * Unwrap the reachability gate around a condition's `else` branch.
+     *
+     * Schemas saved while the gate was emitted wrap the `else` of a chained condition in
+     * `{ if: <field present>, then: <else fields>, else: <forbid> }`. The gate is no longer
+     * produced, but readers that treat `else` as a field container must still unwrap it or
+     * the `else` fields of an already-saved schema are dropped.
+     * @param node `else` node of an `allOf` entry
+     */
+    public static unwrapConditionElse(node: any): any {
```
Pyatakov: The gated `else` shape this unwraps was only ever emitted by 7ae12730e – this branch's own earlier commit, which 0b2f19206 replaced – and never by develop or any release (`git log --all -S"then: elseObj"` finds nothing else). So no stored schema has it, and the shim buys no compatibility while costing ~50 lines replicated in three packages plus four call sites, and it silently rewrites a hand-written presence-only else-if: `else: {if: {required: ['note']}, then: {silver}, else: {bronze}}` parses here as `elseFields: ['silver']` with `bronze` dropped.

Suggest dropping `unwrapConditionElse`/`isPresenceOnlyPredicate` and their call sites here, in `indexer-interfaces`, `indexer-service` and `guardian-service`, unless a stored schema with the gate can actually be pointed at.

Pyatakov:
Re-reviewed at `d42c9bb9e`. Both findings are resolved, verified by running the code.

Tests: `interfaces` 2538 passing (the 7 removed unwrap tests account for the drop); frontend karma 209 passing.

LGTM.

PR #6665, opened 2026-08-14
feat: 6641 make schema import errors and warnings visible

### LINKED ISSUE #6641: Make Schema related Errors and Warnings Prominently Visible on the Import Schema Dialog
## Job Story
When importing a schema and the import schema dialog is open, I want errors and warning to be prominently displayed and not have to scroll to the bottom of a tall list just to see them, so I that don't miss the errors and proceed with an import and can fix them and try again. 

## Acceptance Criteria
- [ ] pin the number of warnings / errors in the footer where it will always be visible, click to navigate to the errors section
- [ ] if a user decides to import a schema with errors or warning, confirm the user's choices with a dialog

## Artifacts & Considerations
https://www.loom.com/share/7f2c2ee9949143d8a85683b93a2505c9

ihar-tsykala:
**Description**:

When a schema file is imported, the preview dialog lists every schema it found and puts the warnings and errors at the  very end of that list. With a large workbook the problems are several screens below the fold, so the user reaches the Import button without ever having seen them and imports a file that is already known to be broken.

The preview now carries a summary line directly above its footer, which stays in view while the list scrolls. It counts the problems the file has — for example "3 errors, 5 warnings" — and points to the detailed list at the end of the schemas. When the file carries hard errors, pressing Import no longer starts the import: a confirmation asks whether to import anyway, and the import runs only if it is accepted. A file with warnings only, or with no problems at all, imports exactly as before.

The Import button itself was deliberately left enabled. Importing a schema with errors is supported on purpose — the broken schemas are saved and marked Incomplete so they can be fixed in place — so the change makes the problems
impossible to miss rather than blocking the action. The detailed list of warnings and errors was not moved, and the shared dialog styling was not changed.

Resolves #6641

**Checklist**

### DIFF
--- a/frontend/src/app/modules/schema-engine/schema-view-dialog/schema-view-dialog.component.html
+++ b/frontend/src/app/modules/schema-engine/schema-view-dialog/schema-view-dialog.component.html
@@ -108,6 +108,14 @@
     </div>
   }
 </div>
+@if (errors.length > 0) {
+  <div class="issues-summary" [class.has-errors]="errorCount > 0">
+    <div class="issues-summary-counts">{{ issuesSummary }}</div>
+    <div class="issues-summary-hint">
+      See the details in the Warning &amp; Errors list at the end of the schemas above.
+    </div>
+  </div>
+}
 <div class="dialog-footer">
   <div class="action-buttons">
     <div class="dialog-button cancel-button">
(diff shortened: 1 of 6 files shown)


ihar-tsykala:
image

image

PR #6671, opened 2026-08-14
fix: 6579 stop recording before completing the save dialog

### LINKED ISSUE #6579: Recording Continues while Completing Save Dialog
## Bug Summary

When I click stop to save a recording, the recording continues while I fill out the dialog box which can add a lot of empty time to a recording.

## Steps to Reporduce

1.  Start a recording
2.  Complete Steps
3.  Click Stop
4.  Fill out dialog box
5.  Save
    

## Expected Result

I expect when I click stop, that the recording will stop at that time and when I save the recording the .record file will not save the time while I'm filling out the form.

## Actual Result

The recording doesn't actually stop until I save the recording files. If I leave the dialog box open for 5 minutes while filling in the title and description of the test the record file will include 5 empty minutes of time which bloats the record file.

## Artifacts

_Add supporting artifacts_

ihar-tsykala:
**Description**:

When a user stops a policy recording, the recording currently continues while the save dialog is open. The time spent entering the recording name and description is therefore included as an empty tail in the downloaded record.

Add a reversible backend pause boundary before metadata entry, then keep the existing stop-and-download operation for finalization. Cancel resumes the same recording, while Save and Stop & Discard retain their existing actions. The
existing archive format, metadata contract, filename handling, and replay behavior remain unchanged.

Fixes #6579

**Checklist**


policy-service/src/policy-engine/record/recording.ts
```diff
@@ -98,14 +99,17 @@ export class Recording {
...
             ? RecordStatus.Recording
             : RecordStatus.New;
+        this._pausedAt = null;
     }
 
     /**
      * Check if recording is active
      * @private
      */
     private isActive(): boolean {
```
Pyatakov, line 128: Pause makes `isActive()` false, and resume is only ever sent from the save dialog's Cancel branch. If the dialog is abandoned (reload, navigate away, tab close) the recording stays paused for the container's lifetime while `getStatus()` still reports `Recording`, so the UI shows an active recording and every subsequent action is dropped silently. Either expose a `paused` flag in the status with a Resume control, or narrow `pause()` to only freeze the timestamp `stop()` uses.

policy-service/src/policy-engine/record/recording.ts
```diff
@@ -182,6 +186,33 @@ export class Recording {
...
+        if (this.mode === 'auto') {
+            return false;
+        }
+        if (
+            this._status !== RecordStatus.Recording ||
+            this._pausedAt === null
+        ) {
+            return false;
+        }
+        this._pausedAt = null;
```
Pyatakov, line 234: `resume()` clears `_pausedAt` without compensating the paused interval. Replay derives waits from `next.time - action.time` (`policy-service/src/policy-engine/record/running.ts:755`), so on the Cancel path the dialog time ends up as a stall in the replayed test – the same artifact the PR removes from the STOP entry.

**Suggested:**
A `_pausedOffset` subtracted in `appendRecord`.

frontend/src/app/modules/policy-engine/record/record-controller/record-controller.component.ts
```diff
@@ -156,8 +161,24 @@ export class RecordControllerComponent implements OnInit {
...
+            } else {
+                this._stopPending = false;
+            }
+        });
+    }
+
+    private pauseAndOpenSaveRecordDialog(includePolicyTestMetadata: boolean): void {
+        this.recordService.pauseRecording(this.policyId).subscribe((result) => {
+            if (!result) {
+                this._stopPending = false;
```
Pyatakov: A falsy pause result (policy restarted, recording controller gone) returns with no dialog, no error, no state change: Stop becomes a silent no-op with no way to clear the stale state. Treat it like the error branch or re-run `loadStatus()`.

frontend/src/app/modules/policy-engine/record/record-controller/record-controller.component.ts
```diff
@@ -156,8 +161,24 @@ export class RecordControllerComponent implements OnInit {
...
-                this.openSaveRecordDialog(false);
+                this.pauseAndOpenSaveRecordDialog(false);
+            } else {
+                this._stopPending = false;
+            }
+        });
+    }
+
+    private pauseAndOpenSaveRecordDialog(includePolicyTestMetadata: boolean): void {
+        this.recordService.pauseRecording(this.policyId).subscribe((result) => {
```
Pyatakov: Deprecated `subscribe(next, error)` overload.

Pyatakov (changes requested):

policy-service/src/policy-engine/record/recording.ts
ihar-tsykala: I fixed the issue. Could you check. Thanks!

policy-service/src/policy-engine/record/recording.ts
ihar-tsykala: I fixed the issue. Could you check. Thanks!

frontend/src/app/modules/policy-engine/record/record-controller/record-controller.component.ts
ihar-tsykala: I fixed the issue. Could you check. Thanks!

frontend/src/app/modules/policy-engine/record/record-controller/record-controller.component.ts
ihar-tsykala: I fixed the issue. Could you check. Thanks!

Pyatakov (approved):

PR #6694, opened 2026-08-18
fix(guardian-service): strip server-managed fields from an imported schema template

vshvets-bc:
A schema template zip is user-supplied. `createSchemaTemplate()` sanitises the payload:

```ts
delete payload._id;
delete payload.id;
delete payload.status;
delete payload.owner;
delete payload.creator;
delete payload.messageId;
delete payload.version;
delete payload.previousVersion;
```

but leaves `topicId`, `contentFileId`, `configFileId` and `_configFileId`.

`SchemaTemplateImportExport.generateZipFile` already omits all of them, so nothing legitimate depends on them surviving an import — a hand-crafted zip is the only way they arrive. Two are dangerous rather than merely untidy.

### Arbitrary GridFS deletion

`contentFileId` / `configFileId` are GridFS handles, and `SchemaTemplate.deleteFiles()` passes them straight to `DataBaseHelper.gridFS.delete()`:

```ts
@AfterDelete()
deleteFiles() {
    if (this.configFileId) { DataBaseHelper.gridFS.delete(this.configFileId)...
    if (this.contentFileId) { DataBaseHelper.gridFS.delete(this.contentFileId)...
```

A user who can create and delete a schema template can therefore delete an arbitrary GridFS file — a policy package, a schema document, another template's payload — by naming its id in the imported zip and then deleting the template.

### Publishing into a topic you do not own

`topicId` short-circuits `createTemplateTopic()`, which returns early when the template already carries one. No topic is created and no `CreateSchemaTemplate` message is sent. The imported template binds silently to an existing topic, its schemas are imported into that topic, and a later publish loads that topic's stored submit key and writes the importer's `PublishSchemaTemplate` message into a topic belonging to someone else.

### Fix

Extend the strip list with the four server-managed fields.

### Tests

Adds a spec to `schema-template-handlers.test.mjs` asserting the forged values are dropped, that legitimate fields survive, and that the server-assigned `owner` / `creator` / `status` still win. It fails on the current code.

Full `guardian-service` suite on this branch: `1807 passing`.

### Checklist

- [x] Includes tests to exercise the new behaviour
- [x] Code is documented
- [x] Local run of tests succeeds
- [x] Commit message includes context
- [ ] Related issue — none found; a search of open issues turned up no existing report

### DIFF
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -147,6 +147,14 @@ async function createSchemaTemplate(
     delete (payload as any).contentFileId;
     delete (payload as any).configFileId;
     delete (payload as any)._configFileId;
+    /*
+     * uuid is server-managed by the same definition as the rest of this list.
+     * generateZipFile strips it on export, so it never appears in a legitimate
+     * archive, but setDefaults() only generates one when it is missing - so a
+     * hand-crafted zip can pin a chosen uuid, which then propagates into the topic's
+     * targetUUID and the snapshot's templateUUID as a misleading identifier.
+     */
+    delete payload.uuid;
 
     payload.creator = owner.creator;
     payload.owner = owner.owner;
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -147,13 +147,8 @@ async function createSchemaTemplate(
     delete (payload as any).contentFileId;
     delete (payload as any).configFileId;
     delete (payload as any)._configFileId;
-    /*
-     * uuid is server-managed by the same definition as the rest of this list.
-     * generateZipFile strips it on export, so it never appears in a legitimate
-     * archive, but setDefaults() only generates one when it is missing - so a
-     * hand-crafted zip can pin a chosen uuid, which then propagates into the topic's
-     * targetUUID and the snapshot's templateUUID as a misleading identifier.
-     */
+    // setDefaults() only generates a uuid when missing, so a forged one would be
+    // kept and propagate into targetUUID / templateUUID.
     delete payload.uuid;
 
     payload.creator = owner.creator;
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -128,22 +128,12 @@ async function createSchemaTemplate(
     }
 
     /*
-     * Strip every server-managed field from an imported payload.
-     *
-     * SchemaTemplateImportExport.generateZipFile already omits these, so nothing
-     * legitimate depends on them surviving an import - a hand-crafted zip is the
-     * only way they arrive. Two of them are dangerous rather than untidy:
-     *
-     *  - contentFileId / configFileId are GridFS handles, and
-     *    SchemaTemplate.deleteFiles() passes them straight to
-     *    DataBaseHelper.gridFS.delete(). A user able to create and delete a
-     *    template could therefore delete an arbitrary GridFS file - a policy
-     *    package, a schema document, another template's payload.
-     *  - topicId short-circuits createTemplateTopic(), which returns early when it
-     *    is already set. The imported template then binds silently to an existing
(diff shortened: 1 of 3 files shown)


guardian-service/tests/unit/schema-template-handlers.test.mjs
```diff
@@ -265,6 +265,56 @@ describe('schema template CRUD and query handlers', () => {
...
+        assert.equal(saved.contentFileId, undefined, 'a forged GridFS handle must not survive import');
+        assert.equal(saved.configFileId, undefined, 'a forged GridFS handle must not survive import');
+        assert.equal(saved._configFileId, undefined, 'a forged GridFS handle must not survive import');
+
+        assert.equal(saved.messageId, undefined);
+        assert.equal(saved.version, undefined);
+        assert.equal(saved.previousVersion, undefined);
+
+        assert.equal(saved.owner, owner.owner);
+        assert.equal(saved.creator, owner.creator);
```
Pyatakov, line 320: This compares `undefined` to `undefined`; the fixture has no `creator`, so the reassignment itself is uncovered.

guardian-service/src/api/schema-template.service.ts
```diff
@@ -135,6 +143,10 @@ async function createSchemaTemplate(
     delete payload.messageId;
     delete payload.version;
     delete payload.previousVersion;
+    delete payload.topicId;
+    delete (payload as any).contentFileId;
+    delete (payload as any).configFileId;
+    delete (payload as any)._configFileId;
```
Pyatakov, line 149: the sibling sanitizer in `policy-import.ts:571` strips `configFileId`/`contentFileId` but not `_configFileId`. The lists are now out of step.

guardian-service/src/api/schema-template.service.ts
```diff
@@ -135,6 +143,10 @@ async function createSchemaTemplate(
     delete payload.messageId;
     delete payload.version;
     delete payload.previousVersion;
+    delete payload.topicId;
```
Pyatakov, line 146: Add `delete payload.uuid;` to the list.

**Why:** `SchemaTemplateImportExport.generateZipFile` already strips `uuid` when exporting (so it never appears in a legitimate export). But `SchemaTemplate.setDefaults()` only generates one if missing, so a hand-crafted zip can pin a specific uuid. That uuid then propagates into the topic's `targetUUID` and the snapshot's `templateUUID`, creating misleading identifiers. Not exploitable today, but it's server-managed by the PR's own definition, so it belongs in the sanitizer for consistency and to prevent future code from relying on it as unique.

Also add it to the sibling sanitizer in `policy-import.ts:571` where it parses a message-sourced zip, that one strips `configFileId`/`contentFileId` but should strip `uuid` too for the same reason.

Pyatakov (changes requested):

vshvets-bc:
Thanks — all three addressed in `bea3a9948`.

**`uuid`** — agreed, and the reasoning holds: it is server-managed by this PR's own definition, `generateZipFile` strips it on export, and `setDefaults()` only generates one when missing, so a hand-crafted zip can pin it and have it propagate into `targetUUID` / `templateUUID`. Deleted with the rest.

**Sibling sanitizer** — `policy-import.ts` now strips `_configFileId` and `uuid` as well, so the two lists are back in step. That one parses a message-sourced zip, which is equally hand-craftable, so it wanted the same treatment rather than just the two fields.

**The vacuous assertion** — correct, and thank you for catching it. The `owner` fixture had no `creator`, so `assert.equal(saved.creator, owner.creator)` compared `undefined` to `undefined` and the reassignment was never exercised. The fixture now carries a distinct `creator`, and the forged payload sets `did:victim` for both `owner` and `creator`, so the assertion proves the overwrite.

I verified the strengthening rather than assuming it: removing either the `creator` reassignment or the new `uuid` delete now fails the test. Before the change, removing the reassignment did not.

guardian-service unit suite: 1790 passing.

Pyatakov (approved):

PR #6696, opened 2026-08-18
fix(guardian-service): stop schema-template reads writing, scope usedByPolicyNames

vshvets-bc:
Two independent defects in `guardian-service/src/api/schema-template.service.ts`.

## 1. Read paths persist normalization ids

`normalizeSchemaTemplateConfig` → `ensureTemplateSchemaReferences` assigns any missing `templateSchemaId` / `templateFieldId` and then persists them:

```ts
changed = SchemaHelper.ensureTemplateFieldIds(schema.document) || changed;
if (changed) {
    await DatabaseServer.updateSchema(schema.id, schema);
}
```

Those helpers run from the **read** paths too — `GET_SCHEMA_TEMPLATE`, `SCHEMA_TEMPLATE_EXPORT_FILE`, and `previewSchemaTemplateUpdate`. A published template is readable by every user holding `TEMPLATES_TEMPLATE_READ`, so:

- a non-owner's GET mutates the owner's schema documents;
- two concurrent readers race to store *different* random ids for the same field, and whichever loses has already handed its ids to a client.

Both helpers now take `persist: boolean = true`; the three read paths pass `false`. The ids are still filled in memory, so responses are unchanged — they are only written when the caller is actually performing a write.

## 2. `usedByPolicyNames` is returned to every reader

`GET_SCHEMA_TEMPLATES` shows published templates to everyone with `TEMPLATES_TEMPLATE_READ`, and `addSchemaCounts` fills `usedByPolicyNames` from the **template owner's** policies — drafts included:

```ts
const allPolicies = template.owner ? await getCachedPolicies(template.owner) : [];
…
item.usedByPolicyNames = usedByPolicies.map(p => p.name || p.id)…
```

So any user could read another Standard Registry's private draft-policy names off a shared listing. `addSchemaCounts` now takes the caller and returns the names only to the template owner. `usedByPoliciesCount` is unchanged for everyone — it says how widely a template is used without naming anything.

## Tests

4 new cases in `guardian-service/tests/unit/schema-template-handlers.test.mjs`; **3 fail against `develop`**:

- `GET_SCHEMA_TEMPLATE` does not persist the ids it fills in (and still fills them in memory, so the response is unchanged)
- `SCHEMA_TEMPLATE_EXPORT_FILE` does not persist them either
- the names are withheld from a non-owner, while the count is kept
- the names are still returned to the owner — the control case, showing the fix does not just drop the field

`guardian-service` unit suite: 1793 passing.

### DIFF
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -434,11 +434,9 @@ async function getPoliciesUsingSchemaTemplate(
 }
 
 /**
- * `owner` gates usedByPolicyNames. The listing shows published templates to every
- * user with TEMPLATES_TEMPLATE_READ, and these names come from the *template
- * owner's* policies, drafts included - so returning them to everyone disclosed
- * another Standard Registry's private draft-policy names. The count is kept for
- * everyone: it says how widely a template is used without naming anything.
+ * `owner` gates usedByPolicyNames: the names come from the template owner's
+ * policies, drafts included. The count stays public - it says how widely a
+ * template is used without naming anything.
  */
 async function addSchemaCounts(templates: SchemaTemplate[], owner?: IOwner): Promise<any[]> {
     const ownerPoliciesCache = new Map<string, any[]>();
@@ -1518,16 +1516,10 @@ async function updateAppliedSchemaTemplate(
 }
 
 /**
- * `persist` exists because the read paths must not write.
- *
- * GET, export and update-preview all normalize the template config, and
- * normalization assigns any missing templateSchemaId / templateFieldId. Writing
- * from there meant a non-owner's GET on a published template - readable by every
- * user with TEMPLATES_TEMPLATE_READ - mutated the owner's schema documents, and
- * two concurrent readers raced to store different random ids for the same field.
- *
- * The ids are still filled in memory, so the response is identical either way;
- * they are only written when the caller is actually performing a write.
+ * `persist` exists because the read paths must not write. Normalization assigns
+ * missing templateSchemaId / templateFieldId, so a non-owner's GET mutated the
+ * owner's schemas and concurrent readers raced to store different ids. The ids are
+ * still filled in memory, so the response is identical either way.
  */
 async function ensureTemplateSchemaReferences(
     schema: Schema,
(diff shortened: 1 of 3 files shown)


guardian-service/src/api/schema-template.service.ts
```diff
@@ -417,7 +417,12 @@ async function getPoliciesUsingSchemaTemplate(
     });
 }
 
-async function addSchemaCounts(templates: SchemaTemplate[]): Promise<any[]> {
+/**
+ * `owner` gates usedByPolicyNames: the names come from the template owner's
+ * policies, drafts included. The count stays public - it says how widely a
+ * template is used without naming anything.
+ */
+async function addSchemaCounts(templates: SchemaTemplate[], owner?: IOwner): Promise<any[]> {
```
Pyatakov: Change signature to `owner: IOwner` (remove the `?`). This makes it a type error to forget it, rather than a silent data loss

guardian-service/src/api/schema-template.service.ts
```diff
@@ -617,7 +624,10 @@ function normalizeTemplateConfigKeys(
     return normalized;
 }
 
-async function normalizeSchemaTemplateConfig(template: SchemaTemplate): Promise<void> {
+async function normalizeSchemaTemplateConfig(
+    template: SchemaTemplate,
+    persist: boolean = true
```
Pyatakov: Change all three (this, lines 1060 and 1497) signatures to default `false`: `persist: boolean = false` and update the five write call sites to explicitly pass `persist: true`. This makes the safe behaviour the default and prevents mistakes

guardian-service/src/api/schema-template.service.ts
Pyatakov, line 378: In `guardian-service/src/helpers/import-helpers/policy/policy-import.ts` around line 596, after setting `schema.templateId = template.id`, add:
  ```ts
  schema.templateSchemaId = schema.templateSchemaId || GenerateUUIDv4();
  SchemaHelper.ensureTemplateFieldIds(schema.document);
  ```
This mirrors the highlighted lines and ensures new rows have the ids from the start.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6697, opened 2026-08-18
fix: keep required custom fields, strip $defs markers, stabilize policy hash

vshvets-bc:
Three small, independent schema-template defects. Each has its own test, and reverting each fix individually fails only its own tests.

## 1. A preserved custom field loses its `required` flag

`preparePolicySchemaUpdate` replaces the target document with a fresh clone of the template — so the `required` list becomes *the template's* — and then calls `mergeCustomFieldsIntoDocument` to put the caller's own fields back. That re-inserted them into `properties` only:

```ts
parent.properties[fieldName] = cloneJson(property);
```

`required` lives on the parent, not on the property, so a **required** custom field survived a template update as **optional**, and VCs missing it started validating. The flag is now carried across with the field, onto the field's own parent rather than the root (nested fields land in the right list).

## 2. Detach leaves `templateFieldId` markers inside `$defs`

`SchemaHelper.removeTemplateFieldIds` walks via `walkDocumentProperties`, which only follows `properties`. Embedded sub-schema definitions under `$defs` kept their markers after a detach — inconsistent with the cleaned sub-schema rows, and leaking into subsequently published documents.

Only *removal* walks `$defs`. `ensureTemplateFieldIds` is deliberately left alone, so this cannot start minting ids in places that never had them.

## 3. The template binding destabilizes the policy hash

`cleanBeforeHash` already drops `components.schemaTemplateSnapshot`, but it left:

- `policy.schemaTemplate` — carries `snapshotId`, `schemaMap` ObjectIds, `appliedAt` / `updatedAt`
- each schema's `templateId` and `templateSchemaId` — assigned per environment by `ensureTemplateSchemaReferences`

All of those differ between environments, so exporting and re-importing an *identical* template-bound policy produced a different hash every time — hash-based same-policy detection reported a difference for every template-bound policy.

## Notes

`mergeCustomFieldsIntoDocument` is exported for its test, matching how `normalizeFieldForDiff` and `buildFieldChangeDetails` are already exported from that module.

## Tests

- `interfaces/tests/schema-helper-template-field-ids.test.mjs` (4) — including a case that asserts the fixture actually carries markers, so the others cannot pass vacuously
- `common/tests/unit-tests/policy-hash-schema-template.test.mjs` (3) — hash unchanged by a binding, stable across environments with different snapshot ids, and still distinguishing policies that genuinely differ
- `guardian-service/tests/unit/schema-template-update-helpers.test.mjs` (4) — required restored, optional left optional, nested tracked on its own parent, no duplicate entry

Suites: interfaces 2543 passing, common 3342 passing, guardian-service unit 1793 passing.

### DIFF
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -1006,35 +1006,35 @@ function findSchemaProperty(document: any, path: string[]): any {
     return current;
 }
 
-function ensureSchemaPropertyParent(document: any, path: string[]): any {
+/**
+ * Walk to the object that owns `path`'s last segment.
+ *
+ * `create` distinguishes the two callers: the target document is being built, so it
+ * wants the `properties` containers filled in on the way down; the source document is
+ * only being read, and creating nodes there would mutate the very thing being copied
+ * from.
+ */
+function schemaPropertyParent(document: any, path: string[], create: boolean): any {
     let current = document;
     for (const key of path.slice(0, -1)) {
         const parent = current?.type === 'array' ? current.items : current;
-        parent.properties = parent.properties || {};
-        current = parent.properties[key];
+        if (create) {
+            parent.properties = parent.properties || {};
+        }
+        current = parent?.properties?.[key];
         if (!current) {
             return null;
         }
     }
     const target = current?.type === 'array' ? current.items : current;
-    target.properties = target.properties || {};
+    if (create && target) {
+        target.properties = target.properties || {};
+    }
     return target;
 }
 
-/**
- * Locate the object that owns `path`'s last segment, without creating anything.
- * Used to read the source document's `required` list for a preserved field.
- */
-function findSchemaPropertyParent(document: any, path: string[]): any {
-    let current = document;
-    for (const key of path.slice(0, -1)) {
-        const target = current?.type === 'array' ? current.items : current;
-        current = target?.properties?.[key];
-        if (!current) {
-            return null;
-        }
-    }
-    return current?.type === 'array' ? current.items : current;
+function ensureSchemaPropertyParent(document: any, path: string[]): any {
+    return schemaPropertyParent(document, path, true);
 }
 
 export function mergeCustomFieldsIntoDocument(
@@ -1064,7 +1064,15 @@ export function mergeCustomFieldsIntoDocument(
          * as optional, and VCs missing it started validating. `required` lives on the
          * parent, not on the property, so it has to be copied separately.
          */
-        const sourceParent = findSchemaPropertyParent(sourceDocument, path);
+        /*
+         * Read the source document's own `required` list rather than the parsed
+         * field.required flag: parseField sets `required || !!conditionRequired`, so a
+         * field required only by a condition branch reports true there while the
+         * document's required array correctly omits it. Copying that flag across would
+         * promote a branch-scoped requirement into an unconditional one - the thing the
+         * $comment marking in this same change exists to avoid.
+         */
+        const sourceParent = schemaPropertyParent(sourceDocument, path, false);
         if (Array.isArray(sourceParent?.required) && sourceParent.required.includes(fieldName)) {
             parent.required = Array.isArray(parent.required) ? parent.required : [];
             if (!parent.required.includes(fieldName)) {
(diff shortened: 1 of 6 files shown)


guardian-service/src/api/schema-template.service.ts
```diff
@@ -1038,6 +1054,23 @@ function mergeCustomFieldsIntoDocument(
...
+         * as optional, and VCs missing it started validating. `required` lives on the
+         * parent, not on the property, so it has to be copied separately.
+         */
+        const sourceParent = findSchemaPropertyParent(sourceDocument, path);
+        if (Array.isArray(sourceParent?.required) && sourceParent.required.includes(fieldName)) {
+            parent.required = Array.isArray(parent.required) ? parent.required : [];
+            if (!parent.required.includes(fieldName)) {
+                parent.required.push(fieldName);
+            }
+        }
```
Pyatakov: This new code walks the `sourceDocument` tree to re-find the owning object's `required` list.
But `field` – is already a parsed `SchemaField` object that carries `.required` as a boolean, set during parsing from that same document's structure.

**What needs to change:** replace the `sourceParent` lookup with a direct check:

```ts
if (field.required) {
    parent.required = Array.isArray(parent.required) ? parent.required : [];
    if (!parent.required.includes(fieldName)) {
        parent.required.push(fieldName);
    }
}
```

Then delete `findSchemaPropertyParent` entirely, it's a duplicate of `ensureSchemaPropertyParent` that only exists to serve this check, and the check doesn't need it.

This is a simplification, not a correctness issue. The current code works because both paths (walking the tree vs. reading the parsed flag) extract the same information. The refactor just removes the redundant tree walk.

Pyatakov (changes requested):

vshvets-bc:
Thanks — half applied, half not, and I want to show why for the second half.

**The duplication is fair, and is gone.** `findSchemaPropertyParent` and `ensureSchemaPropertyParent` walked the same path and differed only in whether they created the `properties` containers on the way down. They are now one helper with a `create` flag, and `ensureSchemaPropertyParent` remains a thin alias so the existing caller is untouched. That was the right call.

**Reading `field.required` instead is not equivalent, so I have not applied it.** `SchemaHelper.parseField` does:

```ts
field.required = required || !!conditionRequired;
```

so a field required only by a *condition branch* reports `required === true` on the parsed object, while the document's own `required` array correctly omits it. I checked rather than reasoned about it — parsing a schema whose custom field carries `conditionRequired: true` in its `$comment`:

```
field=branchOnly  parsed.required=true   inDocumentRequired=false
```

Copying that flag across would push `branchOnly` into the parent's unconditional `required` list. That is precisely what the `$comment` marking in this same PR exists to prevent: JSON Schema cannot tell "the field was answered differently" from "the field was never asked", so a branch-scoped requirement promoted to an unconditional one makes documents unsatisfiable whenever the branch is closed.

I have added two cases pinning the distinction — a branch-required custom field stays out of the parent's `required`, a genuinely required one still crosses. **Applying the suggested form fails 3 of the 12 cases in that file**, which is the clearest argument I can offer that the two paths are not extracting the same information.

Happy to be shown otherwise if you think the condition case cannot arise here. guardian-service unit suite: 1795 passing.

guardian-service/src/api/schema-template.service.ts
```diff
@@ -1064,7 +1064,15 @@ export function mergeCustomFieldsIntoDocument(
          * as optional, and VCs missing it started validating. `required` lives on the
          * parent, not on the property, so it has to be copied separately.
          */
-        const sourceParent = findSchemaPropertyParent(sourceDocument, path);
+        /*
+         * Read the source document's own `required` list rather than the parsed
```
Pyatakov: Please keep your comments brief, 2-3 lines usually is enough

guardian-service/src/api/schema-template.service.ts
vshvets-bc: Kept the source-document lookup, and here is why the two paths differ.

`SchemaHelper.parseField` sets `field.required = required || !!conditionRequired` (interfaces/src/helpers/schema-helper.ts:321). So for a field that is required only inside a condition branch, the parsed flag reads true while the document's own `required` array correctly omits it. Reading `field.required` here would copy that into the target as an unconditional requirement - which is what the `$comment` marking in this same PR exists to prevent.

Shortened the comment to 6 lines per your other note, and `findSchemaPropertyParent` is already gone: it is now `schemaPropertyParent(document, path, create)` with `ensureSchemaPropertyParent` as a two-line wrapper.

Happy to switch if I have misread `parseField`.

Pyatakov (approved):

PR #6700, opened 2026-08-18
fix(guardian-service): refuse to strand published schemas, free the superseded package

vshvets-bc:
Two file/lifecycle defects around schema templates.

## 1. Deleting a `PUBLISH_ERROR` template strands published schemas

`ensureEditable` blocks only a `PUBLISHED` template, so a `PUBLISH_ERROR` one is deletable. But `publishSchemaTemplate` flips every template schema to `PUBLISHED` *before* the step that can fail:

```ts
for (const schema of schemas as Schema[]) {
    schema.status = SchemaStatus.PUBLISHED;
}
await DatabaseServer.updateSchemas(schemas as Schema[]);
```

…and the delete handler's cleanup removes only `DRAFT` and `ERROR` rows. So schemas an earlier attempt had already published survived the template and were left pointing at a `templateId` that no longer exists.

Deletion is now refused while they exist, naming them — mirroring the used-by-policies guard directly above it. Nothing that reached a topic is destroyed silently, and the owner is told what to resolve. A template whose schemas are all `DRAFT` or `ERROR` still deletes exactly as before.

## 2. Re-publishing orphans the previous package

```ts
template.contentFileId = await DatabaseServer.saveFile(GenerateUUIDv4(), Buffer.from(buffer));
```

`configFileId` has the `_configFileId` "previous handle" mechanism on the entity for exactly this, cleaned up by its `@AfterUpdate` hook. `contentFileId` has none, so every publish attempt simply overwrote the handle and left the file it replaced in GridFS forever.

The superseded file is deleted after the new one is safely stored, and best-effort — losing the old file is not a reason to fail a publish.

## Tests

4 cases in `guardian-service/tests/unit/schema-template-handlers.test.mjs`, **all failing against `develop`**:

- delete is refused when a schema is `PUBLISHED`, the message names it, and the template survives the refusal
- delete still succeeds and still runs the ordinary `DRAFT`/`ERROR` cleanup when nothing is published
- the superseded content package is deleted on re-publish
- nothing is deleted on a first publish

guardian-service unit suite: 1793 passing.

### DIFF
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -536,13 +536,9 @@ async function publishSchemaTemplate(
 
         notifier.startStep(STEP_SAVE_FILE);
         /*
-         * Drop the file the previous attempt left behind.
-         *
-         * configFileId has the _configFileId "previous handle" mechanism on the
-         * entity, cleaned up by its @AfterUpdate hook; contentFileId has no
-         * equivalent, so every publish attempt simply overwrote the handle and
-         * stranded the file it replaced. Deleted after the new one is safely stored,
-         * and best-effort: losing the old file is not a reason to fail a publish.
+         * contentFileId has no _configFileId-style previous-handle mechanism, so each
+         * publish overwrote the handle and stranded the old file. Deleted after the new
+         * one is stored, best-effort: losing it must not fail a publish.
          */
         const supersededContentFileId = template.contentFileId;
         template.contentFileId = await DatabaseServer.saveFile(GenerateUUIDv4(), Buffer.from(buffer));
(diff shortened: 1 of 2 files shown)


guardian-service/src/api/schema-template.service.ts
```diff
@@ -2134,6 +2148,26 @@ export async function schemaTemplatesAPI(logger: PinoLogger): Promise<void> {
...
+                    if (publishedSchemas.length) {
+                        const names = publishedSchemas
+                            .map((schema) => schema.name || schema.iri || schema.id)
+                            .filter((name) => !!name);
+                        const shown = names.slice(0, 5).join(', ');
+                        const hidden = names.length > 5 ? ` and ${names.length - 5} more` : '';
+                        throw new Error(
+                            `Schema template has published schemas and cannot be deleted` +
+                            `${shown ? `: ${shown}${hidden}` : ''}. ` +
+                            `They were published by an earlier attempt; remove them first.`
```
Pyatakov: Change error message, replace:
```
They were published by an earlier attempt; remove them first.
```
with:
```
They were published by an earlier attempt; retry the publish to finish it.
```
The current message tells the owner to do something that's impossible – there's no way to delete a published schema (the delete path rejects anything that isn't DRAFT or ERROR). Retrying the publish is the action that actually works: it's allowed on a PUBLISH_ERROR template, it will finish the original attempt, and it brings the template into sync with its already-published schemas.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6701, opened 2026-08-18
fix(guardian-service): make schema-template apply and update recoverable

vshvets-bc:
Three related defects in `applySchemaTemplate` / `updateAppliedSchemaTemplate`.

## 1. APPLY has no rollback

The schema copies are persisted one at a time:

```ts
for (const schema of templateSchemas as Schema[]) {
    const copied = await createSchemaAndArtifacts(SchemaCategory.POLICY, copy, owner, NewNotifier.empty());
    copiedSchemas.push(copied);
}
```

A throw part-way through leaves schemas 1..N-1 sitting in the policy topic carrying `templateId` and `templateSchemaId`, with no snapshot and no binding. Nothing cleans them up — and because the binding is written **last**, `hasSchemaTemplateBinding()` still reports false, so a retry sails past the guard and copies the whole set again. **Every failed attempt adds another duplicate.**

The only existing compensation covered a failing `updatePolicy`, and only by removing the snapshot.

Undo is exact here: these are rows the call created and their ids are known.

## 2. UPDATE mutates before it commits

Policy schemas are rewritten **in place** and persisted before the new snapshot is saved and the binding swapped. A mid-loop failure leaves some schemas on the new template version while `policy.schemaTemplate` still points at the old snapshot and state hash — and the catch removes only the new snapshot, never the schema edits. Preview then diffs against a snapshot that no longer describes reality.

In-place edits cannot be undone by deletion, so each document is captured before it is touched and restored on failure; copies created during an update are deleted like APPLY's. The old snapshot is now removed only *after* the new binding is committed, so a failure above still leaves the policy describable by the snapshot its binding points at.

## 3. Neither has a concurrency guard

Two overlapping runs for one policy both pass the "already applied" check, both do the work, and both call `updatePolicy`: last writer wins, the loser's snapshot is orphaned and its copies are left as duplicates. The binding cannot be the guard, because it is written last — that window is precisely what is being raced.

`withPolicyTemplateLock` is modelled on `withMigrationLock` (`common/src/helpers/migration-lock.ts`), which already runs in production: a Mongo compare-and-set with a lease, so a holder that dies frees the policy after the TTL rather than wedging it forever.

Unlike the migration lock it does **not** wait. These run behind a request or a push task, and queueing a second APPLY would only mean it starts, discovers the binding the first one just wrote, and fails anyway — after a long silence. Failing immediately says something true and actionable instead. If there is no connection to coordinate through it runs unlocked, which is the pre-existing behaviour rather than a new failure mode.

All compensation is best-effort: a compensating write can itself fail, and the original error is the one worth surfacing.

## Tests

- `guardian-service/tests/unit/policy-template-lock.test.mjs` — 6 cases: acquire and release, refusal of a concurrent run on the same policy, no blocking across different policies, release when the operation throws, takeover of an expired lease, and the no-connection fallback.
- `guardian-service/tests/unit/schema-template-handlers.test.mjs` — 7 cases: APPLY deletes its copies when a later copy / the snapshot / the binding write fails and removes nothing on success; UPDATE restores the edited document, removes the new snapshot rather than the old one on failure, and removes the old one only once the binding is committed.

Reverting each rollback fails its own cases. The existing APPLY suite stubs the lock module so esmock does not re-import `@guardian/common` through it (that alone cost ~4s per case).

guardian-service unit suite: 1802 passing.

### DIFF
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -1462,7 +1462,9 @@ async function updateAppliedSchemaTemplate(
      */
     const originalSchemas = new Map<string, Schema>();
     const createdSchemas: Schema[] = [];
+    const pendingRemovals: Schema[] = [];
     let nextSnapshot: Awaited<ReturnType<typeof saveApplySnapshot>> | null = null;
+    let result: Policy;
     const rollback = async (): Promise<void> => {
         if (nextSnapshot) {
             try {
@@ -1541,7 +1543,9 @@ async function updateAppliedSchemaTemplate(
         );
         const action = conflict ? resolutions.get(conflict.id) : null;
         if (action === SchemaTemplateUpdateResolutionAction.REMOVE_FROM_POLICY) {
-            await removePolicySchema(policySchema, owner);
+            // deleted only past the commit: rollback() restores originalSchemas and
+            // drops createdSchemas, and a hard-deleted row is in neither
+            pendingRemovals.push(policySchema);
         } else {
             // another in-place edit; capture before detaching the markers
             if (!originalSchemas.has(policySchema.id)) {
@@ -1582,18 +1586,24 @@ async function updateAppliedSchemaTemplate(
         updatedAt: appliedAt,
         schemaMap
     };
-    const result = await DatabaseServer.updatePolicy(context.policy);
+    result = await DatabaseServer.updatePolicy(context.policy);
     // the old snapshot only goes once the new binding is committed, so a failure
     // above still leaves the policy describable by its old snapshot
     nextSnapshot = null;
-    await DatabaseServer.removeSchemaTemplateSnapshot(previousSnapshot);
-    return result;
     } catch (error) {
         // undo the copies and restore the edited rows, then surface the original
         // failure. Previously only the new snapshot was removed.
         await rollback();
         throw error;
     }
+
+    // Past the commit point. rollback() must not run from here: the binding already
+    // names the new snapshot, so restoring the old schema rows would contradict it.
+    for (const policySchema of pendingRemovals) {
+        await removePolicySchema(policySchema, owner);
+    }
+    await DatabaseServer.removeSchemaTemplateSnapshot(previousSnapshot);
+    return result;
 }
 
 /**
--- a/guardian-service/src/api/schema-template.service.ts
+++ b/guardian-service/src/api/schema-template.service.ts
@@ -50,6 +50,7 @@ import {
     TopicType
 } from '@guardian/interfaces';
 import { ApiResponse } from './helpers/api-response.js';
+import { withPolicyTemplateLock } from '../helpers/policy-template-lock.js';
 import { createSchemaAndArtifacts, deleteSchema, SchemaImportExportHelper, updateSchemaDefs } from '../helpers/import-helpers/index.js';
 
 async function createTemplateTopic(
@@ -1436,6 +1437,7 @@ async function updateAppliedSchemaTemplate(
     templateId: string,
     policyId: string,
     owner: IOwner,
+    logger: PinoLogger,
     options?: ISchemaTemplateUpdateOptions
 ): Promise<any> {
     const context = await loadSchemaTemplateUpdateContext(templateId, policyId, owner, true);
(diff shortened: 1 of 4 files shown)


guardian-service/src/api/schema-template.service.ts
```diff
@@ -1463,12 +1509,16 @@ async function updateAppliedSchemaTemplate(
...
     };
-    try {
-        const result = await DatabaseServer.updatePolicy(context.policy);
-        await DatabaseServer.removeSchemaTemplateSnapshot(previousSnapshot);
-        return result;
+    const result = await DatabaseServer.updatePolicy(context.policy);
+    // the old snapshot only goes once the new binding is committed, so a failure
+    // above still leaves the policy describable by its old snapshot
+    nextSnapshot = null;
+    await DatabaseServer.removeSchemaTemplateSnapshot(previousSnapshot);
```
Pyatakov: `updatePolicy` on 1512 commits the new binding, but line 1516 is still inside the same `try`, so a throw from `removeSchemaTemplateSnapshot(previousSnapshot)` reaches `rollback()`. That reverts every edited schema to its captured document and deletes every schema the call created, while `policy.schemaTemplate` already points at the new snapshot, new state hash, and a `schemaMap` whose ids no longer exist. `nextSnapshot = null` spares the new snapshot but not the schema restore, which is the destructive half.

**Suggested**
Move the old-snapshot cleanup past the commit point. Wrap it in a conditional that only runs if the new snapshot was successfully committed (after `updatePolicy` succeeds):
A `finally` block with a guard – only delete `previousSnapshot` if `nextSnapshot === null`, meaning the binding got written and the new snapshot is safe to keep.

guardian-service/src/api/schema-template.service.ts
```diff
@@ -1429,6 +1470,10 @@ async function updateAppliedSchemaTemplate(
         if (action === SchemaTemplateUpdateResolutionAction.REMOVE_FROM_POLICY) {
```
Pyatakov, line 1545: `REMOVE_FROM_POLICY` hard-deletes a policy schema (`removePolicySchema` -> `deleteSchema` -> `deleteSchemas`) before the snapshot and binding are written, and `rollback()` restores only `originalSchemas` / deletes `createdSchemas`. A schema removed here is in neither, so a later `updatePolicy` failure loses it permanently and leaves the binding pointing at a snapshot that still lists it – the exact failure this PR sets out to close. Collect into `pendingRemovals` and drain after the commit.

**Suggestion**
Defer the hard delete. Instead of calling `removePolicySchema` immediately, collect those schemas in a `pendingRemovals` list during the loop. After `updatePolicy` succeeds on line 1512, then drain `pendingRemovals` with the actual deletes. That way if `updatePolicy` fails, nothing was deleted and rollback can restore the edited ones.

Pyatakov (changes requested):

guardian-service/src/api/schema-template.service.ts
```diff
@@ -1463,14 +1513,24 @@ async function updateAppliedSchemaTemplate(
...
+        await rollback();
         throw error;
     }
+
+    // Past the commit point. rollback() must not run from here: the binding already
+    // names the new snapshot, so restoring the old schema rows would contradict it.
+    for (const policySchema of pendingRemovals) {
+        await removePolicySchema(policySchema, owner);
+    }
+    await DatabaseServer.removeSchemaTemplateSnapshot(previousSnapshot);
```
Pyatakov: The post-commit cleanup is unguarded, so a committed update still reports as failed. Both the `pendingRemovals` drain and the `removeSchemaTemplateSnapshot` call now sit outside the try with no error handling of their own. A throw from either propagates to the API handler, which logs it and returns `MessageError`, so the user is told the update failed when in fact `updatePolicy` already committed and the binding is live.
**Suggestion:**
Wrap the `pendingRemovals` drain loop and the `removeSchemaTemplateSnapshot` call in try/catch (each in its own) that logs on error.

guardian-service/src/api/schema-template.service.ts
```diff
@@ -1597,12 +1597,33 @@ async function updateAppliedSchemaTemplate(
...
-    // names the new snapshot, so restoring the old schema rows would contradict it.
+    /*
+     * Past the commit point. rollback() must not run from here: the binding already
+     * names the new snapshot, so restoring the old schema rows would contradict it.
+     *
+     * Neither cleanup may throw either. updatePolicy has committed and the binding is
+     * live, so letting a failure propagate would reach the API handler and report the
+     * update as failed when it succeeded. Both are logged and swallowed; what they
+     * leave behind is a stale row, not a broken binding.
+     */
```
Pyatakov: No need for an own comment, the code is self-explanatory

vshvets-bc: Dropped in 9f51602d9 — the code stands on its own.

Pyatakov (approved):

PR #6702, opened 2026-08-18
fix: an organization policy assignment must belong to the policy's owner

vshvets-bc:
Two halves of one privilege escalation.

## 1. `ASSIGN_POLICY_TO_ORG` never looks at the policy

The gateway route calls straight through to NATS:

```ts
const owner = new EntityOwner(user);
return toOrgResponse(await (new Users()).assignPolicyToOrg(id, body.policyId, owner));
```

`auth-service` owns the `PolicyOrgAssignment` row but **not** the `Policy` collection, so its handler can only verify that the caller owns the organization and that `policyId` is non-empty. Nothing above it looks at the policy either.

A Standard Registry can therefore assign **any** policy id — including another SR's — to their own organization.

## 2. `policy-service` honours that assignment across owners

```ts
if (await DatabaseServer.getAssignedEntity(AssignedEntityType.Policy, this.policyId, policyUser.did)) {
    return true;
}
try {
    return await isPolicyAssignedToUserOrg(policyUser, this.policyId);
```

There is no policy-owner comparison, so the assignee's organization members pass the relayed block-action gate for a policy belonging to someone else. `guardian-service`'s `accessPolicyCode` rejects `user.owner !== policy.owner` *before* it consults an assignment — the two gates were asymmetric.

## Fix

**Gateway.** It is the one layer that can reach both services, so it resolves the policy and refuses anything the caller does not own — 404 when the policy does not exist, 403 when it belongs to another owner. `InternalException` re-throws `HttpException` unchanged, so both statuses survive to the client.

**policy-service.** The gate now makes the same owner comparison `accessPolicyCode` already makes, so a row written before the gateway check existed is no longer honoured. `PolicyUser` did not carry the user's owning registry, so it gains `parent`, derived exactly as `EntityOwner` derives `owner` (`user.parent` for a USER, absent for a bare DID).

Only a **known** mismatch denies. A virtual (dry-run) user or one built from a bare DID has no parent, and treating unknown as mismatched would break dry-run runs that never had an owner to compare. A direct `AssignedEntity` grant is unaffected — the guard applies to the organization fallback.

## Tests

`api-gateway/tests/service/organization-policy-assign.test.mjs` — 4 cases: an owned policy is assigned, a foreign policy is 403 **and nothing is written**, a missing policy is 404, and the lookup uses the id being assigned.

`policy-service/tests/unit-tests/policy-engine/actions-service-access-policy.test.mjs` — 5 cases, including that on a mismatch the organization fallback is **not even consulted**, so a stale row cannot grant access; plus a direct entity assignment and blanket `ACCESS_POLICY_ALL` staying unaffected.

`policy-user.test.mjs` — 3 cases for `parent`.

All fail against `develop`. api-gateway 3038 passing, policy-service 3613 passing.

### DIFF
--- a/policy-service/src/policy-engine/actions-service.ts
+++ b/policy-service/src/policy-engine/actions-service.ts
@@ -864,6 +864,22 @@ export class PolicyActionsService {
             if (await DatabaseServer.getAssignedEntity(AssignedEntityType.Policy, this.policyId, policyUser.did)) {
                 return true;
             }
+            /*
+             * An organization assignment only grants access within the policy's own
+             * owner.
+             *
+             * guardian-service's accessPolicyCode rejects `user.owner !== policy.owner`
+             * before it consults an assignment; this gate did not, so the two were
+             * asymmetric - a PolicyOrgAssignment row naming a policy that belongs to a
+             * different Standard Registry let that organization's members through here.
+             *
+             * Only a *known* mismatch denies. A virtual (dry-run) user or one built
+             * from a bare DID has no parent, and treating unknown as mismatched would
+             * break dry-run runs that never had an owner to compare.
+             */
+            if (policyUser.parent && this.policyOwner && policyUser.parent !== this.policyOwner) {
+                return false;
+            }
             try {
                 return await isPolicyAssignedToUserOrg(policyUser, this.policyId);
             } catch {
--- a/policy-service/src/policy-engine/actions-service.ts
+++ b/policy-service/src/policy-engine/actions-service.ts
@@ -865,17 +865,9 @@ export class PolicyActionsService {
                 return true;
             }
             /*
-             * An organization assignment only grants access within the policy's own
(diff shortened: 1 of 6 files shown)


policy-service/src/policy-engine/actions-service.ts
```diff
@@ -864,6 +864,22 @@ export class PolicyActionsService {
             if (await DatabaseServer.getAssignedEntity(AssignedEntityType.Policy, this.policyId, policyUser.did)) {
                 return true;
             }
+            /*
```
Pyatakov, line 867: Please update both comments (here and at policy-service/src/policy-engine/policy-user.ts:85) to be concise and describe the current behavior rather than previous implementation.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6703, opened 2026-08-18
fix(frontend): scope the app theme to the signed-in user

vshvets-bc:
## Problem

`AppThemeService` stores the theme under a single global key:

```ts
const APP_THEME_STORAGE_KEY = 'GUARDIAN_APP_THEME';
…
localStorage.setItem(APP_THEME_STORAGE_KEY, this.currentTheme);
```

A browser is shared. So the preference one account chose is applied to the **next account that signs in on the same machine**, and it survives signing out entirely — nothing clears or rescopes it.

## Fix

The storage key is suffixed with the user id, and `AuthStateService` drives it: adopt the signing-in user's preference, drop back to the default when the session ends. The theme follows the account rather than the machine.

The pre-upgrade global key is still read as a fallback when an account has no stored choice of its own, so an existing preference is not silently lost on upgrade — and a user's own choice takes precedence over it once made.

One subtlety worth flagging for review: the `AuthStateService` constructor calls `updateState(false, true)` as a bootstrap "we don't know yet" state. That is **not** a sign-out and must not reset the theme — doing so would discard it on every page refresh. That is exactly what the existing `noClearLocalStorage` flag distinguishes, so the reset is gated on it.

`AuthService` gains `getUserId()`, which decodes `userId` from the access token. It is used **only** to scope preferences stored in this browser to the account that chose them, never as an access decision: the token is not verified there, and a forged one would only let its bearer read their own local settings. That is stated in the doc comment so it does not get repurposed later.

## Tests

9 cases across two new specs, **6 of which fail against the current single-key behaviour**:

`app-theme.service.spec.ts`
- the choice is stored under the signed-in user
- one user's theme is not carried over to the next
- a user's own theme is restored when they sign back in
- sign-out drops back to the default **without discarding** the stored preference
- a preference under the pre-upgrade global key is still adopted
- a user's own choice wins over the global key

`auth-state.service.spec.ts`
- the bootstrap state does not reset the theme
- signing in adopts that user's theme
- the session ending resets it

Run with `ng test --include='**/app-theme.service.spec.ts' --include='**/auth-state.service.spec.ts'`: 9 SUCCESS.

### DIFF
--- a/frontend/src/app/services/app-theme.service.ts
+++ b/frontend/src/app/services/app-theme.service.ts
@@ -8,7 +8,19 @@ export interface AppThemeOption {
     icon: string;
 }
 
+/**
+ * Per-user, because a browser is shared.
+ *
+ * A single global key meant the theme one account chose was applied to the next
+ * account that signed in on the same machine, and survived signing out. The
+ * legacy key is still read once as a fallback so an existing preference is not
+ * silently lost on upgrade.
+ */
 const APP_THEME_STORAGE_KEY = 'GUARDIAN_APP_THEME';
+
+function storageKeyForUser(userId: string | null): string {
+    return userId ? `${APP_THEME_STORAGE_KEY}:${userId}` : APP_THEME_STORAGE_KEY;
+}
 const LIGHT_CLASS = 'guardian-theme-light';
 const DARK_CLASS = 'guardian-theme-dark';
 
@@ -23,6 +35,7 @@ export class AppThemeService implements OnDestroy {
     ];
 
     private currentTheme: AppTheme = 'light';
+    private userId: string | null = null;
     private readonly darkModeQuery: MediaQueryList = window.matchMedia('(prefers-color-scheme: dark)');
     private readonly onSystemChange = (): void => {
         if (this.currentTheme === 'system') {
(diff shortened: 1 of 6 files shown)


frontend/src/app/services/auth.service.ts
```diff
@@ -108,6 +108,33 @@ export class AuthService {
...
+        }
+        try {
+            const payload = token.split('.')[1];
+            if (!payload) {
+                return null;
+            }
+            // base64url -> base64
+            const normalized = payload.replace(/-/g, '+').replace(/_/g, '/');
+            const decoded = JSON.parse(atob(normalized));
+            return decoded?.userId ? String(decoded.userId) : null;
```
Pyatakov: The access token has no `userId` claim, so `getUserId()` returns `null` for every real session.

`generateAccessToken` (auth-service/src/utils/user-access-token.ts:74) signs `{ username, did, role }` (+ `expireAt`), and every token the frontend stores comes from that one issuer. With `null`, `storageKeyForUser` falls back to the bare `GUARDIAN_APP_THEME` key, i.e. exactly the shared-machine behaviour the PR is fixing. Both new specs pass only because they supply the id themselves (one calls `applyForUser('user-a')`, the other stubs `getUserId`).

Fix: read `decoded?.username`, handle non-ASCII before `JSON.parse` (`atob` mangles it), and add a spec built from the real payload shape.

frontend/src/app/services/app-theme.service.ts
```diff
@@ -46,11 +55,37 @@ export class AppThemeService implements OnDestroy {
 
     public setTheme(theme: AppTheme): void {
         this.currentTheme = this.findTheme(theme).value;
-        localStorage.setItem(APP_THEME_STORAGE_KEY, this.currentTheme);
+        localStorage.setItem(storageKeyForUser(this.userId), this.currentTheme);
```
Pyatakov: A `setTheme` with a null `userId` writes to the bare legacy key, which is also the fallback `readStoredTheme` uses at line 89. So it doesn't merely fail to scope, it overwrites the value every not-yet-chosen user reads. Guard the write on `this.userId`.

frontend/src/app/services/app-theme.service.ts
```diff
@@ -46,11 +55,37 @@ export class AppThemeService implements OnDestroy {
...
+     * preference - the next sign-in reads its own.
+     */
+    public reset(): void {
+        this.userId = null;
+        this.currentTheme = this.themes[0].value;
         this.applyResolvedTheme();
     }
 
     private readStoredTheme(): AppTheme {
+        const stored = localStorage.getItem(storageKeyForUser(this.userId));
```
Pyatakov, line 97: The constructor resolves the theme with `userId` still null, while `applyForUser` only arrives from the header's `sessions()` callback after an HTTP round-trip. Once a preference lives under the per-user key, a dark-mode user gets a flash of light on every refresh – new behaviour. Persist a "last signed-in id" and seed `this.userId` in the constructor.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6710, opened 2026-08-19
fix(frontend): log a failed dry-run comment count instead of swallowing it

vshvets-bc:
**Scope cut after review** — the backend half is gone. What remains is one frontend error arm.

## What the review established

The eleven handlers this PR originally touched are the discussions/comments endpoints, and their only frontend caller is the comments panel, which is gated off in dry-run (`vc-fullscreen-dialog.component.ts:220` sets `this.comments = comments !== false && !dryRun`). None of them is reachable that way. The stated repro ("Download → 422 Document not found") cannot happen either: `onDownloadJsonFile` (`:355`) builds a Blob from JSON already in the dialog and never calls the server.

`findPolicyVcDocument` and its eleven call sites have been dropped, along with their test. `policy-engine.service.ts` is back to `develop`.

## What is left

`getPolicyCommentsCount` in `documents-source-block.component.ts:311` fires on dialog close **for every row, dry-run included**, with no error handler. In dry-run the count is not resolvable and the failure was swallowed.

It now logs. The row keeps its previous count either way — this does not make a stale counter fresh, it stops the failure disappearing silently. The comment says exactly that and nothing more.


Pyatakov (changes requested):
@vshvets-bc **The premise of the PR is wrong.**

All 11 changed handlers are exactly the discussions/comments endpoints served by `api-gateway/src/api/service/policy-comments.ts` (users, relationships, schemas, discussions, comments, comment-count, discussion-key, ipfs add/get). Nothing else sends those NATS events.
The comments panel is the only frontend caller, and it is hosted in exactly one place – `vc-fullscreen-dialog.component.html:198` under `@if (comments)`, where `frontend/src/app/modules/schema-engine/vc-fullscreen-dialog/vc-fullscreen-dialog.component.ts:220` sets `this.comments = comments !== false && !dryRun`. Dry-run documents reach that dialog with `dryRun: !!row.dryRunId`, so the panel never renders and none of those endpoints is called.
The stated repro ("click Download → 422 Document not found") doesn't exist at all: `[onDownloadJsonFile]` at `frontend/src/app/modules/schema-engine/vc-fullscreen-dialog/vc-fullscreen-dialog.component.ts:355` makes a Blob from the JSON already in the dialog and never hits the server.
One nuance in the PR's favour: `frontend/src/app/modules/policy-engine/policy-viewer/blocks/documents-source-block/documents-source-block.component.ts:311` calls `getPolicyCommentsCount` on dialog close **for every row**, dry-run included, with no error handler. So one of the eleven call sites is genuinely reachable in dry-run, and its failure is silently swallowed, not user-visible. That's the entire real defect.

Server side there is no published-only guard (`accessPolicy(..., 'execute')` only checks ownership/assignment), and `PolicyCommentsUtils.isDryRun` exists to route Hedera messages as virtual, so the backend would support dry-run comments. But nothing today exercises it.

frontend/src/app/modules/policy-engine/policy-viewer/blocks/documents-source-block/documents-source-block.component.ts
```diff
@@ -310,9 +310,15 @@ export class DocumentsSourceBlockComponent implements OnInit {
...
-                    .subscribe((count) => {
-                        row.comments = count?.count ?? row.comments;
-                    });
+                    .subscribe(
+                        (count) => {
+                            row.comments = count?.count ?? row.comments;
+                        },
+                        // fires for every row on dialog close, dry-run included, where
+                        // the count is not resolvable. Without an arm the failure was
+                        // swallowed and the stale count silently kept.
```
Pyatakov: After the change, if a failure occurs, the counter remains stale – it reflects the count as of the previous successful update. So the only difference now is that the failure is not swallowed but is logged. Currently, the comment is causing confusion.

guardian-service/src/policy-engine/policy-engine.service.ts
```diff
@@ -147,6 +147,19 @@ export class PolicyEngineChannel extends NatsService {
     }
 }
 
+/**
+ * Resolve a policy document from the collection that policy actually writes to.
+ *
+ * getVCById reads VcDocument; a dry-run policy's documents live in DryRun, keyed by
+ * dryRunId. Looking a dry-run document up the first way finds nothing.
+ */
+export async function findPolicyVcDocument(policy: any, documentId: string): Promise<any> {
```
Pyatakov: Please, drop it entirely as it is not needed

frontend/src/app/modules/policy-engine/policy-viewer/blocks/documents-source-block/documents-source-block.component.ts
vshvets-bc: Agreed — the count stays exactly as it was; the only change is that the failure is logged instead of disappearing. Comment now says that and nothing more (fba1dfe4c).

guardian-service/src/policy-engine/policy-engine.service.ts
vshvets-bc: Dropped. `findPolicyVcDocument`, all eleven call sites and its test are gone; `policy-engine.service.ts` is back to `develop` (d495f4fd8). The PR is frontend-only now — retitled and the description rewritten to match.

Pyatakov (approved):

PR #6716, opened 2026-08-19
fix: stop dropping schema import errors

### LINKED ISSUE #6715: Schema import errors never reach the user, and one unknown field type collapses all of them
Two independent problems stop the schema-import error surface from working.

## 1. An unknown field type crashes the error collector (common)

`common/src/xlsx/models/xlsx-result.ts`, in `updateSchemas()`:

```ts
const link = this._linkCache.get(field.type);
if (!link) {
    this.addError({ type: 'error', text: `Unknown field type.`, ... }, field);
}          // <-- no continue

let subSchemaCache: ISchemaCache;
if (link.worksheet) {          // <-- TypeError when link is undefined
```

The `addError` is followed by a dereference of the same `undefined` link. The resulting `TypeError: Cannot read properties of undefined (reading 'worksheet')` is caught by the outer `try/catch` around `updateSchemas`, which replaces **every error collected so far** with one generic `Failed to update schemas.` and abandons the rest of the run.

So a partial import with a single bad cell loses all of its precise, per-field diagnostics — the opposite of what the error surface is for.

Related: line ~227 skips only `#GeoJSON`, but `guardian-service/src/helpers/import-helpers/schema/schema-import.ts` pre-seeds **both** `#GeoJSON` and `#SentinelHUB` as validated ref types. A `#SentinelHUB` field therefore has no link entry, is reported as an unknown type (false positive), and then triggers the crash above.

## 2. Import errors never reach the UI (frontend)

The backend builds `this.errors` and returns them as `ImportSchemaResult.errors`, but `frontend/src/app/modules/common/async-progress/async-progress.component.ts` ignores `result.errors` for `IMPORT_SCHEMA_FILE` / `IMPORT_SCHEMA_MESSAGE` and just navigates to `schemas`.

A JSON or xlsx bundle that fails def validation completes **silently**; the only residual signal is the per-schema `status = ERROR` badge in the list.

## Reproduce

1. Import an xlsx whose schema has an unrecognised (or `#SentinelHUB`) field type — actual: one generic "Failed to update schemas.", every other error lost.
2. Import a bundle containing a def-validation error — actual: navigates to Schemas with nothing shown.

## Expected

One bad cell does not discard the other errors, `#SentinelHUB` is not reported as unknown, and the aggregate errors are shown when the import task completes.

vshvets-bc:
Fixes #6715

## Problem

**common** — `updateSchemas()` pushes an "Unknown field type" error for an unresolvable ref and then dereferences the same `undefined` link (`link.worksheet`). The `TypeError` is caught by the surrounding `catch`, which replaces every error collected so far with one generic `Failed to update schemas.` and abandons the run. One bad cell discards all the precise diagnostics.

`#SentinelHUB` is also pre-seeded as a validated ref type by the schema import (alongside `#GeoJSON`), so it has no link entry — it was reported as an unknown type and then triggered that crash.

**frontend** — the import returns `ImportSchemaResult.errors`, but the `IMPORT_SCHEMA_FILE` / `IMPORT_SCHEMA_MESSAGE` resolver navigates away without reading it, so a bundle that fails def validation completes silently.

## Fix

- add the missing `continue` after the unknown-type `addError`, and skip `#SentinelHUB` alongside `#GeoJSON`
- surface `result.errors` at task completion as one sticky toast naming each schema that failed

Reported by toast rather than through `SchemaViewDialog`: that dialog counts `type === 'error' | 'warning'` and renders `.text`, which does not match `ImportSchemaError`'s `{type,uuid,name,error}` and would render blank rows.

## Tests

New `common/tests/unit-tests/xlsx-result/xlsx-result-unknown-type.test.mjs`:

- the unknown-type error survives and is **not** replaced by the generic message
- errors from schemas processed after the bad one are still collected
- `#SentinelHUB` is treated as known, like `#GeoJSON`

All three fail on `develop`; the first reports `TypeError: Cannot read properties of undefined (reading 'worksheet')`, which is the bug. `17 passing` for the xlsx-result suite on this branch.

The fixtures stub everything `updateSchemas()` touches, so a throw is the defect under test rather than an artefact of the fixture.

### DIFF
--- a/common/src/xlsx/models/xlsx-result.ts
+++ b/common/src/xlsx/models/xlsx-result.ts
@@ -224,7 +224,9 @@ export class XlsxResult {
                 const schemaCache = this._schemaCache.find(c => c.iri === schema.iri);
                 for (const field of schema.fields) {
                     if (field.isRef) {
-                        if (field.type === '#GeoJSON') {
+                        // both are pre-seeded as validated ref types by the schema import,
+                        // so neither has a link entry
+                        if (field.type === '#GeoJSON' || field.type === '#SentinelHUB') {
                             continue;
                         }
                         const link = this._linkCache.get(field.type);
@@ -236,6 +238,9 @@ export class XlsxResult {
                                 worksheet: schemaCache?.worksheet,
                                 row: field.order
                             }, field);
+                            // without this, link.worksheet below throws and the outer
+                            // catch replaces every collected error with one generic message
+                            continue;
                         }
 
                         let subSchemaCache: ISchemaCache;
--- a/common/src/xlsx/models/xlsx-result.ts
+++ b/common/src/xlsx/models/xlsx-result.ts
@@ -238,8 +238,8 @@ export class XlsxResult {
                                 worksheet: schemaCache?.worksheet,
                                 row: field.order
                             }, field);
-                            // without this, link.worksheet below throws and the outer
-                            // catch replaces every collected error with one generic message
+                            // clear the dangling ref so it is not persisted into the schema
+                            field.type = null;
                             continue;
                         }
 
--- a/common/tests/unit-tests/xlsx-result/xlsx-result-unknown-type.test.mjs
+++ b/common/tests/unit-tests/xlsx-result/xlsx-result-unknown-type.test.mjs
@@ -1,0 +1,73 @@
+import { assert } from 'chai';
+import { XlsxResult } from '../../../dist/xlsx/models/xlsx-result.js';
+import { Workbook } from '../../../dist/xlsx/models/workbook.js';
+
+function ws(name) {
+    return new Workbook().createWorksheet(name);
+}
+
+/*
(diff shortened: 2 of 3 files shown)


common/src/xlsx/models/xlsx-result.ts
```diff
@@ -236,6 +238,9 @@ export class XlsxResult {
                                 worksheet: schemaCache?.worksheet,
                                 row: field.order
                             }, field);
+                            // without this, link.worksheet below throws and the outer
+                            // catch replaces every collected error with one generic message
+                            continue;
```
Pyatakov, line 243:
```suggestion
                            field.type = null;
                            continue;
```
This matches what the sibling error case does at line 269, clearing the dangling reference so it doesn't get persisted into the schema. Without it, the field keeps its original unknown type and survives into the imported schema.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6718, opened 2026-08-19
fix(frontend): guard lists that are read before they load

### LINKED ISSUE #6717: Comparison pages crash on an empty diff and show nothing on failure; Wiper Requests dialog crashes on open
Four components read a list before it exists.

## 1. Comparison pages crash on an empty/degenerate diff

`compare-policy`, `compare-schema` and `compare-module` assign `this.X = <node>?.report` and then immediately loop `this.X.length` with no guard:

- `frontend/src/app/modules/analytics/compare-policy/compare-policy.component.ts:77,81` — `this.blocks = blocks?.report;` then `for (let i = 0; i < this.blocks.length; i++)`
- `.../compare-schema/compare-schema.component.ts:51,57` — same via `fields?.report`
- `.../compare-module/compare-module.component.ts:112,115` — same via `blocks?.report`

The `?.report` shows the node can be absent, but the `.length` that follows does not allow for it. A degenerate diff throws `Cannot read properties of undefined (reading 'length')`, and because the parent has already cleared its spinner the user is left on a visibly broken page.

The newer siblings — `compare-tool`, `compare-document`, `compare-record` — all harden this exact spot with `Array.isArray(...)` (e.g. compare-tool `:247`). The guard was never back-ported to the three older ones.

## 2. A failed comparison shows a blank page with no message

`frontend/src/app/modules/analytics/compare/compare.component.ts` — all eleven load error callbacks do only:

```ts
}, ({ message }) => {
    this.loading = false;
    console.error(message);
});
```

`this.error` is never set, so the template's `@if (error)` banner is reachable only from the local "Invalid params" pre-check. On a timeout, a 500 or an oversized diff the spinner stops and the user sees an empty page with no explanation.

## 3. Wiper Requests dialog crashes on open

`frontend/src/app/modules/contract-engine/dialogs/wipe-requests-dialog/wipe-requests-dialog.component.ts:13-17` declares:

```ts
requests: { id: string; user: string; token?: string; }[];   // undefined until the load resolves
```

while `wipe-requests-dialog.component.html:4` reads `@if (requests.length)` on the first render — so `undefined.length` throws in change detection before `getWipeRequests` resolves.

All four sibling dialogs (retire-requests, retire-pools, user-retire-pools, user-retire-requests) initialise to `[]`; this one is the only exception.

Worse, the load error callback (`:56`) only sets `loading = false` and never assigns `requests`, so if the request list API fails the `@if` keeps throwing and **neither the list nor the empty state ever renders** — a permanently broken dialog body.

## Reproduce

1. Compare two items whose diff is empty/degenerate → crash.
2. Trigger a compare API failure or timeout → blank page, no message.
3. Open the Wiper Requests dialog on a wipe contract → crash; if the list API errors, the body never recovers.

vshvets-bc:
Fixes #6717

## Fixes

**compare-policy / compare-schema / compare-module** — normalise `X?.report` to `[]` with `Array.isArray`, matching what the newer `compare-tool` / `compare-document` / `compare-record` already do. The `?.` anticipated an absent report node; the loop right after it did not.

**wipe-requests-dialog** — initialise `requests` to `[]` like all four sibling dialogs, and reset it in the load error arm so the empty state renders instead of the body staying permanently broken.

**compare.component** — route all eleven load error callbacks through one helper that sets `this.error`, so the template's existing `@if (error)` banner is reached on an HTTP failure rather than only from the local "Invalid params" pre-check.

## Tests

Four new spec files (16 tests). Run against `develop` with only the source changes reverted, **11 of them fail**:

- `ComparePolicyComponent` / `CompareSchemaComponent` / `CompareModuleComponent` — "survives a diff whose report is missing" (3)
- `CompareComponent` — banner set for document/schema/module/policy/tool failures, plus the no-message fallback (6)
- `WipeRequestsDialogComponent` — empty initial list, and the empty state after a failed load (2)

The 5 that pass either way are the companion assertions (populated reports still compute `_collapse`, a successful load leaves the banner clear, the header total is read) — they are there to show the guards did not change the working path.

### DIFF
--- a/frontend/src/app/modules/analytics/compare/compare.component.ts
+++ b/frontend/src/app/modules/analytics/compare/compare.component.ts
@@ -241,6 +241,17 @@ export class CompareComponent implements OnInit {
         return results;
     }
 
+    /**
+     * Every load error used to be swallowed into console.error, leaving a stopped
+     * spinner over a blank page: the template's `@if (error)` banner was only ever
+     * reachable from the local "Invalid params" pre-check.
+     */
+    private onLoadFailed(message?: string): void {
+        this.loading = false;
+        this.error = message || 'The comparison could not be loaded. Please try again.';
+        console.error(message);
+    }
+
     private loadDocument() {
         this.error = null;
         const options = {
@@ -262,8 +273,7 @@ export class CompareComponent implements OnInit {
                 this.loading = false;
             }, 500);
         }, ({ message }) => {
-            this.loading = false;
-            console.error(message);
+            this.onLoadFailed(message);
         });
     }
 
@@ -287,8 +297,7 @@ export class CompareComponent implements OnInit {
             }
             this.loading = false;
         }, ({ message }) => {
-            this.loading = false;
-            console.error(message);
+            this.onLoadFailed(message);
         });
     }
     private loadOriginalPolicy(policyId: string) {
@@ -307,8 +316,7 @@ export class CompareComponent implements OnInit {
                 this.loading = false;
             }, 500);
         }, ({ message }) => {
-            this.loading = false;
-            console.error(message);
+            this.onLoadFailed(message);
         });
     }
 
@@ -333,8 +341,7 @@ export class CompareComponent implements OnInit {
                 this.loading = false;
             }, 500);
         }, ({ message }) => {
-            this.loading = false;
-            console.error(message);
+            this.onLoadFailed(message);
         });
     }
 
@@ -358,8 +365,7 @@ export class CompareComponent implements OnInit {
             }
             this.loading = false;
         }, ({ message }) => {
-            this.loading = false;
-            console.error(message);
+            this.onLoadFailed(message);
         });
     }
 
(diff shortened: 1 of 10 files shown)


frontend/src/app/modules/analytics/compare/compare.component.ts
```diff
@@ -241,6 +241,17 @@ export class CompareComponent implements OnInit {
         return results;
     }
 
+    /**
+     * Every load error used to be swallowed into console.error, leaving a stopped
+     * spinner over a blank page: the template's `@if (error)` banner was only ever
+     * reachable from the local "Invalid params" pre-check.
+     */
+    private onLoadFailed(message?: string): void {
```
Pyatakov, line 249: 1. Add an export-specific handler next to `onLoadFailed`.
```ts
private onExportFailed(message?: string): void {
    this.loading = false;
    this.error = message || 'The report could not be exported. Please try again.';
    console.error(message);
}
```

2. Repoint the five export callbacks from `onLoadFailed` to `onExportFailed` – the ones at lines 300, 368, 413, 466 and 515. The six remaining call sites (276, 319, 344, 391, 440, 491) stay on `onLoadFailed`.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6720, opened 2026-08-19
fix(frontend): stop the Policy Wizard producing policies with broken $ref chains

### LINKED ISSUE #6719: Policy Wizard accepts schemas with missing sub-schema refs, hangs on a failed lookup, and renders its schema multiselect detached
Three problems in `frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/`.

## 1. No dependency-completeness check before Create Policy

The wizard resolves each selected schema with `getSchemaWithSubSchemas` but carries only the **parent** forward — `onSelectedSchemaChange` creates one form control and one tree node per selected schema from `data.schema`, and `data.subSchemas` is discarded.

Nothing verifies that every schema in the referenced `$ref` chain is itself present. A schema whose sub-schemas were deleted is accepted, and the failure only surfaces afterwards, on the assembled policy:

> Schema with id '#6d0f197f-…' refers to non-existing schema '#999ebbb8-…&1.0.0'
> Schema with id '#283a34b4-…' refers to invalid schema '#8debf876-…'
> Schema with id '#8debf876-…' refers to invalid schema '#5cc0ce56-…'

From the user's side: the wizard let them select the schemas, so why is the result invalid?

## 2. A failed sub-schema lookup hangs the wizard

Both `getSchemaWithSubSchemas` subscribes have a **next callback only** — `policy-wizard-dialog.component.ts:354` (preset restore) and `:690` (selection):

```ts
const resolvedSchema: Schema = await new Promise<Schema>((resolve) => {
    this.schemaService.getSchemaWithSubSchemas(category, id, topicId).subscribe((data) => {
        ...
        resolve(addedSchema);
    });                       // <-- no error arm
});
...
Promise.all(schemaPromises).then(() => {
    this.selectedSchemas = value;
    this.matTree?.refreshTree();
});
```

If the request errors, the promise never settles, so `Promise.all` never resolves, `selectedSchemas` is never assigned and the tree never refreshes. The wizard simply stops responding, with no error anywhere in the UI.

## 3. The schema multiselect renders detached and full-width

`policy-wizard-dialog.component.html:261` — the grouped `p-multiSelect` uses `appendTo="body"` while living **inside** the wizard's PrimeNG DynamicDialog. The overlay is appended to `<body>`, escaping the dialog's positioning/stacking context, so the align pass mispositions it relative to the trigger. `guardian-multiselect-panel` sets no `width` / `max-width`, so when align fails the panel stretches to the body width — a full-viewport band with the options pinned to the far left, overlapping the wizard.

## Reproduce

1. Open a policy in the configurator, launch the Policy Wizard, go to **Policy Schemas**.
2. Open the **Select Schema** multiselect → the panel renders detached and full-width.
3. Select schema(s) whose sub-schema `$ref` chain is not fully present, finish the wizard → the created policy is invalid with "refers to non-existing / invalid schema", with no earlier warning.
4. Make `getSchemaWithSubSchemas` fail → the wizard stops responding with no message.

vshvets-bc:
Fixes #6719

## Fixes

**Dependency completeness** — keep the `subSchemas` the backend already returns, walk the transitive `$ref` chain of everything selected before Create Policy, and name the unresolved IRIs in the dialog instead of handing back a policy the backend will reject. `#GeoJSON` and `#SentinelHUB` are whitelisted, matching `TemplateUtils.checkSchemaVariables`.

**Hanging lookup** — add the missing error arm at **both** call sites (`:354` preset restore and `:690` selection). Each resolves with the unresolved schema and records it, so `Promise.all` settles and the failure is reported as a dependency that could not be verified rather than silently freezing the wizard.

**Detached multiselect** — drop `appendTo="body"` so the overlay stays in the dialog's positioning context, and cap `guardian-multiselect-panel` so a future align failure cannot reproduce the full-width band.

## Tests

New `policy-wizard-dialog.component.spec.ts`, 8 tests, **all 8 fail on `develop`**:

- blocks creation and names the unresolved ref
- follows the chain transitively
- creates when the whole chain resolves
- treats the built-in defs as resolvable
- survives a self-referencing schema (no infinite walk)
- a failed lookup still settles the selection
- a failed lookup is reported as an unverifiable dependency
- the resolved sub-schemas are registered, not just the parent

## Not addressed

The duplicated OWNER / Trust Chain nodes in the step tree, also reported in #6719 — that is step-tree generation and wants its own change.

### DIFF
--- a/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.html
+++ b/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.html
@@ -17,6 +17,17 @@
 </div>
 <div class="dialog-footer">
   {{this.dataForm.errors}}
+  @if (missingSubSchemas.length) {
+    <div class="wizard-missing-schemas">
+      <div>The selected schemas reference sub-schemas that are not available in this policy:</div>
+      <ul>
+        @for (iri of missingSubSchemas; track iri) {
+          <li>{{ iri }}</li>
+        }
+      </ul>
+      <div>Add the missing schemas, or deselect the schemas that depend on them, before creating the policy.</div>
+    </div>
+  }
   <div class="action-buttons">
     <div class="create-policy-button">
       <div class="dialog-button">
@@ -258,7 +269,6 @@
       [group]="true"
       [ngModel]="selectedSchemas"
       [options]="groupedSchemas"
-      appendTo="body"
       placeholder="Select Schema"
       >
     </p-multiSelect>
--- a/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.scss
+++ b/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.scss
@@ -434,3 +434,17 @@ form {
 .guardian-input-container {
     margin-bottom: 24px;
 }
+.wizard-missing-schemas {
+    margin-bottom: 16px;
+    padding: 12px 16px;
+    border-radius: 8px;
+    background: var(--guardian-red-transparent, rgba(255, 0, 0, 0.06));
+    color: var(--guardian-red, #c00);
+    font-size: 13px;
+
+    ul {
+        margin: 8px 0;
+        padding-left: 20px;
+        word-break: break-all;
+    }
+}
--- a/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.scss
+++ b/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.scss
@@ -237,6 +237,19 @@ form {
     overflow-y: auto;
 }
 
+/*
+ * The schema multiselect renders its overlay inline (see the template), so it lives
+ * inside .dialog-body's scroll box. Bounding the list keeps the panel short enough to
+ * open within that box instead of running past its fold.
+ */
+::ng-deep .guardian-multiselect-panel {
+    .p-multiselect-items-wrapper,
+    .p-multiselect-list-container {
+        max-height: min(220px, 30vh);
+        overscroll-behavior: contain;
+    }
+}
+
 ::ng-deep {
     .p-dialog-title {
         font-family: Poppins, sans-serif;
@@ -438,8 +451,8 @@ form {
     margin-bottom: 16px;
     padding: 12px 16px;
     border-radius: 8px;
-    background: var(--guardian-red-transparent, rgba(255, 0, 0, 0.06));
-    color: var(--guardian-red, #c00);
+    background: var(--color-accent-red-2, rgba(255, 0, 0, 0.06));
+    color: var(--color-accent-red-1, #c00);
     font-size: 13px;
 
     ul {
--- a/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
+++ b/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
@@ -41,6 +41,10 @@ export class PolicyWizardDialogComponent implements OnInit, AfterViewInit {
         schemas: Schema[];
     }[] = [];
     public selectedSchemas: Schema[] = [];
+    public missingSubSchemas: string[] = [];
+    // parent + every sub-schema the backend resolves, keyed by iri
+    private resolvedSchemas = new Map<string, Schema>();
+    private failedSchemaLookups = new Set<string>();
     public mintedSchemas: Schema[] = [];
     public selectedTrustChainRoles: string[] = [];
 
@@ -358,7 +362,13 @@ export class PolicyWizardDialogComponent implements OnInit, AfterViewInit {
                             addedSchema = new Schema(data.schema);
                         }
 
+                        this.registerResolvedSchemas(addedSchema, data?.subSchemas);
                         resolve(addedSchema);
+                    }, (error) => {
+                        // without an error arm the promise never settles and the wizard hangs
+                        console.error(error);
+                        this.failedSchemaLookups.add(schema?.iri || id);
+                        resolve(schema);
                     });
                 });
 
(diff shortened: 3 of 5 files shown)


frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
```diff
@@ -694,7 +775,17 @@ export class PolicyWizardDialogComponent implements OnInit, AfterViewInit {
...
+                    this.registerResolvedSchemas(addedSchema, data?.subSchemas);
                     resolve(addedSchema);
+                }, (error) => {
+                    /*
+                     * Without an error arm this promise never settles, so Promise.all
+                     * below never resolves, selectedSchemas is never assigned and the
+                     * tree never refreshes - the wizard simply stops responding.
+                     */
+                    console.error(error);
+                    this.failedSchemaLookups.add(schema?.iri || id);
```
Pyatakov, line 798: `failedSchemaLookups` is only ever added to (here and at the line 370) and read (711). The deselect branch above filters `selectedSchemas`, `currentNode.children`, the form array and `mintedSchemas` but never this set, and the success arm only writes `resolvedSchemas`. So one transient network error leaves the IRI in the set forever, `findMissingSubSchemas` re-adds it on every `onCreate`, and Create silently refuses for the rest of the dialog – even after the user deselects that schema or reselects it successfully. Needs a `delete` in the success arm and one per deleted schema. Suggestion:
   - Add `this.failedSchemaLookups.delete(addedSchema?.iri || id)` in the success arm (after line 779)
   - Add `this.failedSchemaLookups.delete(deletedSchema.iri)` in the deselect loop (inside lines 727-760) for each removed schema

frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.html
```diff
@@ -258,7 +269,6 @@ <h3>Important parameters</h3>
       [group]="true"
       [ngModel]="selectedSchemas"
       [options]="groupedSchemas"
-      appendTo="body"
```
Pyatakov, line 261: Removing `appendTo="body"` moves the overlay back inside `.dialog-body`, which is `max-height: 75vh; overflow-y: auto` (scss 234-238). `overflow-y: auto` with `overflow-x: visible` computes to auto/auto, so the panel is now clipped by that scroll container instead of floating over the dialog.
If the width fix is what's wanted, keeping `appendTo="body"` plus the panel `max-width` gets it without the clipping. Suggestion:
   - **Option A (keep the fix):** keep `appendTo="body"` – the `max-width` on `.guardian-multiselect-panel` alone may be sufficient and avoids the clipping
   - **Option B (justify the removal):** if `appendTo="body"` was causing issues, add CSS to handle the clipping (e.g., scoped overflow rules or a repositioning) and add a short comment why the removal was necessary
   - Test with a long schema list to confirm either path works

frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
Pyatakov: Test run logs two unhandled
`TypeError: Cannot read properties of undefined (reading 'filter')` at policy-wizard-dialog.component.ts:842 – `schema.fields.filter` in `onSelectedSchemaChange`.
That call sits *after* the awaited promise inside the async map callback, so the throw rejects that element's promise, `Promise.all` rejects, and the `.then` at line 795 never assigns `selectedSchemas` or refreshes the tree. Added specs pass only because none of them assert on the tree for that case.
**Suggestion**:
Default the `fields` – `schema.fields?.filter` or `(schema.fields || []).filter`

frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.scss
```diff
@@ -433,4 +433,18 @@ form {
...
 .guardian-input-container {
     margin-bottom: 24px;
-}
\ No newline at end of file
+}
+.wizard-missing-schemas {
+    margin-bottom: 16px;
+    padding: 12px 16px;
+    border-radius: 8px;
+    background: var(--guardian-red-transparent, rgba(255, 0, 0, 0.06));
```
Pyatakov: `--guardian-red` / `--guardian-red-transparent` aren't defined anywhere in `themes`; the existing tokens are `--color-accent-red-1/2`. Replace `--guardian-red` -> `--color-accent-red-1` and `--guardian-red-transparent`-> `--color-accent-red-2`

frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
```diff
@@ -637,10 +647,81 @@ export class PolicyWizardDialogComponent implements OnInit, AfterViewInit {
...
+            missing.add(iri);
+        }
+        return Array.from(missing);
+    }
+
     onCreate() {
         if (!this.dataForm.valid) {
             return;
         }
+        this.missingSubSchemas = this.findMissingSubSchemas();
```
Pyatakov, line 722: `missingSubSchemas` is only recomputed in `onCreate`, so the banner keeps showing stale IRIs while the user fixes the selection.
Add `this.missingSubSchemas = []` at the top of `onSelectedSchemasChange` (around line 733) so the banner disappears as soon as the user starts fixing the selection, not only on the next Create click

Pyatakov (changes requested):

frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.html
vshvets-bc: You are right that the inline overlay sits inside `.dialog-body` (`max-height: 75vh; overflow-y: auto`) and can be clipped by it. I went a third way, because neither option as stated fixes the width.

Option A does not bound anything: appended to `body`, `max-width` resolves against the viewport, so `100vw` is a no-op there exactly as it was inline. And restoring `appendTo="body"` reintroduces the symptom the ticket reported - the panel rendered *detached* from its trigger. This dialog carries transforms, and PrimeNG's absolute positioning measures offsets that a transformed ancestor invalidates, which is what pulled the panel away from the field.

So:

- The `max-width` came back out of `.guardian-multiselect-panel` entirely. That class is shared by eight other components (`schema-rule-config-dialog` alone has seven), so it was never a PR-local rule - and since it never bound anything, removing it loses nothing.
- The overlay stays inline, keeping it anchored to the trigger.
- The list is bounded in this dialog's own scope (`min(220px, 30vh)`), short enough to open inside the 75vh box rather than past its fold.

This is reasoned from the CSS rather than confirmed in a browser, so if you have seen the detached behaviour resolve some other way I will take that instead.

Pyatakov (approved):

PR #6722, opened 2026-08-19
fix(frontend): confirm a button-block decision and restore the row on failure

### LINKED ISSUE #6721: button block: an accepted Approve/Reject gives no feedback, and a rejected one leaves no buttons at all
`frontend/src/app/modules/policy-engine/policy-viewer/blocks/button-block/button-block.component.ts`

## 1. No feedback on success

`onSelect(button)` submits via `setBlockData` with an **empty success callback** (`:245-257`):

```ts
this.policyEngineService.setBlockData(this.id, this.policyId, { document: this.data, tag: button.tag })
    .subscribe(
        () => { },                              // success: nothing at all
        (e) => { console.error(e.error); this.loading = false; this.cdref.detectChanges(); }
    );
```

Clicking Approve on a review step (e.g. as VVB) produces no toast, no message, nothing — indistinguishable from the click not registering.

## 2. A rejected submit leaves the user with no buttons

`onSelect` optimistically hides the button row with `this.commonVisible = false` (`:244`) **before** the request. The error arm only sets `loading = false` — it never restores `commonVisible`.

So when the submit is rejected the buttons stay hidden and the only trace is a `console.error`. The user is left with no controls, no explanation, and no way to retry short of reloading.

## Reproduce

1. In a dry-run or published policy, reach a review step with Approve / Reject.
2. Click **Approve** → no confirmation of any kind.
3. Make the `setBlockData` call fail → the button row disappears permanently and nothing is shown.

## Expected

A successful decision is confirmed, and a rejected one restores the buttons and says why.

## Note

There is a related report that the buttons *reappear* after a successful approve. That is a different thing: visibility is recomputed from the block state the server pushes back (`loadData` sets `commonVisible = true`, `setData` recomputes each `button.visible`), so if the refresh returns the same actionable state the buttons legitimately come back. Forcing them hidden client-side would hide a control that may still be actionable, so this issue does not cover that.

vshvets-bc:
Fixes #6721

## Fixes

- the empty success callback now confirms the submitted decision by button name (falling back to `title`, then `tag`)
- the error arm restores `commonVisible`, which `onSelect` cleared optimistically before the request, and reports the failure as a sticky toast instead of only `console.error`

## Tests

New `button-block.component.spec.ts`, 5 tests. Run against `develop` with the source change reverted, **3 fail**:

- confirms an accepted decision by name
- falls back to the tag when the button has no name
- restores the buttons and reports a rejected submit

The other 2 (the `{document, tag}` payload, and the row hiding while a decision is in flight) pass either way — they are there to show the change did not alter the submit path or the optimistic hide.

## Deliberately not included

Forcing the decision buttons to stay hidden after a successful submit. Their visibility is computed from the block state the server pushes back, so overriding it client-side would hide a control that may genuinely still be actionable — see the note at the end of #6721.

### DIFF
--- a/frontend/src/app/modules/policy-engine/policy-viewer/blocks/button-block/button-block.component.ts
+++ b/frontend/src/app/modules/policy-engine/policy-viewer/blocks/button-block/button-block.component.ts
@@ -233,7 +233,12 @@ export class ButtonBlockComponent implements OnInit {
         return result;
     }
 
-    onSelect(button: any) {
+    /*
+     * `restorePoint` is captured by onSelectDialog before it appends the comment,
+     * so a rejected submit can undo that too. Snapshotting here would keep it.
+     */
+    onSelect(button: any, restorePoint?: string) {
+        const snapshot = restorePoint ?? JSON.stringify(this.data ?? null);
         void this.clearOutgoingHideEventsState()
             .then(() => {
                 return this.writeOutgoingHideEventsState(button);
@@ -264,6 +269,13 @@ export class ButtonBlockComponent implements OnInit {
                      * left with no controls and no explanation.
                      */
                     this.commonVisible = true;
+                    try {
+                        this.data = JSON.parse(snapshot);
+                    } catch (parseError) {
+                        console.error(parseError);
+                    }
+                    // the persisted hide-events value records a decision that never landed
+                    void this.clearOutgoingHideEventsState().catch((err) => console.error(err));
                     this.loading = false;
                     this.toastService.error(
                         e?.error?.message || 'The action could not be submitted.',
@@ -285,6 +297,7 @@ export class ButtonBlockComponent implements OnInit {
             },
         })!;
 
+        const restorePoint = JSON.stringify(this.data ?? null);
         dialogRef.onClose.subscribe((result) => {
             if (result) {
                 let comments = this.getObjectValue(
@@ -303,7 +316,7 @@ export class ButtonBlockComponent implements OnInit {
                     button.dialogResultFieldPath || this._commentField,
                     comments
                 );
-                this.onSelect(button);
+                this.onSelect(button, restorePoint);
             }
         });
     }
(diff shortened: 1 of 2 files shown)


frontend/src/app/modules/policy-engine/policy-viewer/blocks/button-block/button-block.component.ts
Pyatakov, line 250: The issue requires reverting **two optimistic updates** that happen on line 245 and lines 237–240, not just the UI visibility.

**What currently happens:**

1. User clicks a button -> `onSelect` is called
2. Line 245: `this.data` is mutated with the new value
3. Lines 237–240: Hide-events state is written to IndexedDB
4. Request is sent with `commonVisible = false`
5. If request **fails**, the error handler (line 259–274) only restores the UI (`commonVisible = true`, `loading = false`)

**The problem:**

After a failed submit, `this.data` still holds the value from the failed action, so the next click sends a document that already includes it. And the persisted hide-events value records a decision that never landed, and when `loadData` runs next, `applyIncomingHideEventsVisibility` applies that stale state to the buttons.

**What needs fixing:**

In the error handler, you need to:

1. **Restore `this.data`** to its state before line 245. Capture the previous field value before mutating it, then revert on error:
   ```ts
   const previousValue = this.getObjectValue(this.data, button.field);
   // ... existing line 245 ...
   (e) => {
       this.setObjectValue(this.data, button.field, previousValue);
   ```

2. **Clear the IndexedDB hide-events write** by calling `clearOutgoingHideEventsState()` in the error handler (add it before the line 273). This undoes the persisted value that `applyIncomingHideEventsVisibility` would otherwise apply:
   ```ts
   void this.clearOutgoingHideEventsState().catch((err) => console.error(err));
   ```

Then restore the UI visibility/loading state as the PR already does.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6724, opened 2026-08-19
fix(frontend): clear "Unsaved changes" when an edit is reverted

### LINKED ISSUE #6723: Schema editor keeps showing "Unsaved changes" after the edit is reverted
## Symptom

In the schema editor, add a change (e.g. a condition) and then remove it again so the schema is back to its last-saved state. The header still shows amber **"Unsaved changes"** and the red **Save all** button, as if edits were pending.

## Root cause

Dirty tracking is **add-only** — there is no comparison against the saved state. `markDirty()` in `frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts` only adds the schema's id to `dirtySchemaIds`; nothing ever removes it when the edit is reversed:

```ts
public get hasUnsavedChanges(): boolean {
    if (this.isTemplateConfigMode) { return this.templateConfigDirty; }
    return this.dirtySchemaIds.size > 0;   // once added, stays until save
}
```

`dirtySchemaIds` is cleared only on a successful save, or when a brand-new unsaved schema is deleted. Reverting a change so it matches what was loaded does not clear it, so the flag is a one-way latch: any edit marks the schema dirty until the user saves.

## Reproduce

1. Open a saved schema. Add a condition (or edit any field) — "Unsaved changes" appears.
2. Delete that condition / undo the edit so the schema matches its saved state.
3. Actual: "Unsaved changes" and the red Save all remain. Expected: with no net change, the editor returns to "All saved".

## Impact

Low correctness risk — Save all is idempotent — but it erodes trust in the indicator and pushes users to save no-op changes, each of which is a real write.

## Note for whoever fixes this

The failure mode to avoid is the opposite one. Declaring a schema clean when it is not would hide Save all and silently discard the user's work, which is far worse than the stale indicator. Whatever comparison is used should fail *toward* dirty whenever it cannot prove the schema matches.

vshvets-bc:
Fixes #6723

## Fix

Keep a signature of the last-saved state per schema — taken when a schema loads (single load and list load) and again after a successful save — and have `markDirty()` drop the id when the current state matches it.

The signature covers what the editor can actually change: the schema's name/description/entity, and the full field tree including nested `fields` and `conditions`.

## The failure mode this avoids

The signature deliberately **fails toward dirty**. A cycle, a depth blow-out, or anything else it cannot model returns `null`, and callers treat `null` as "cannot prove clean" and leave the schema dirty.

A false clean would hide Save all and silently discard the user's work — far worse than the stale indicator being fixed. Schemas that have never been saved have no baseline and are always dirty.

## Tests

8 new tests in the existing `schemas-configuration.component.spec.ts`. Run against `develop` with the source change reverted, **6 fail**:

- marks dirty on edit and clean again on revert
- notices an added and then removed field
- notices an added and then removed condition
- stays dirty when only part of the edit is reverted
- stays dirty when the field tree is cyclic
- a successful save becomes the new baseline

The 2 that pass either way pin the fail-toward-dirty behaviour (no baseline, and never-saved schemas) — they must keep passing, which is the point of including them.

`39 passing` on this branch.

### DIFF
--- a/frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
+++ b/frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
@@ -1115,64 +1115,106 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
         return schemaConfig.fields[fieldKey];
     }
 
+    /*
+     * Counts every object and array level, not just field nesting. A field one level
+     * deeper costs two (its parent's `fields` array, then the field object), so this is
+     * double the ~12 levels of field nesting actually intended. Scalars return before
+     * the check and cost nothing. Too low is not harmless: the throw means "cannot
+     * prove clean", so a legitimately deep schema would latch "Unsaved changes" forever
+     * - the exact failure this fix exists to remove.
+     */
+    private static readonly SIGNATURE_MAX_DEPTH = 24;
+
+    /* Recomputed by the editor rather than edited by the user; everything else is hashed. */
+    private static readonly SIGNATURE_IGNORED_KEYS = new Set([
+        'errors', 'path', 'fullPath', 'parent', 'controlKey', '_id',
+    ]);
+
     /**
      * A stable signature of everything the editor can change.
+     *
+     * This walks the schema generically instead of hashing a hand-picked list of
+     * properties. An allow-list fails in the dangerous direction: any editable
+     * property nobody remembered to add is invisible to the check, the dirty flag
+     * clears, `saveAll()` early-returns on `!hasUnsavedChanges`, and the edit is lost
+     * on navigation with no toast and no guard dialog. Ignoring only what is provably
+     * derived inverts that - an unrecognised property makes the schema look dirty,
+     * which merely costs a redundant save.
+     *
+     * It also removes two whole classes of blind spot for free: a condition built with
+     * AND/OR has no `ifCondition.field`, so every predicate edit used to hash to the
+     * same null; and `then`/`else` targets are now reached like anything else.
      *
      * Returns null on anything it cannot model (cycle, depth blow-out, throw). Callers
      * treat null as "cannot prove clean" and leave the schema dirty: a false clean would
      * hide Save all and silently discard the user's work, which is far worse than the
      * stale "Unsaved changes" this fixes.
      */
-    private static readonly SIGNATURE_MAX_DEPTH = 12;
-
     private schemaSignature(schema: Schema | null | undefined): string | null {
         if (!schema) {
             return null;
         }
         try {
-            const seen = new WeakSet<object>();
-            const field = (f: any, depth: number): any => {
-                if (!f || typeof f !== 'object') {
-                    return f ?? null;
+            /*
+             * Path-local rather than one set for the whole traversal. A global set
+             * rejects any object reached twice, not just cycles - and two ref fields
+             * pointing at the same sub-schema IRI share their field objects, so the
+             * signature would return null and leave that schema permanently dirty.
+             */
+            const ancestors = new Set<object>();
+            const walk = (value: any, depth: number): any => {
+                if (value === undefined || typeof value === 'function') {
+                    return null;
                 }
-                if (depth > SchemasConfigurationComponent.SIGNATURE_MAX_DEPTH || seen.has(f)) {
+                if (value === null || typeof value !== 'object') {
+                    return value;
+                }
+                if (depth > SchemasConfigurationComponent.SIGNATURE_MAX_DEPTH || ancestors.has(value)) {
                     throw new Error('unbounded');
                 }
-                seen.add(f);
-                return [
-                    f.name, f.title, f.description, f.type, f.format, f.pattern, f.unit,
-                    f.unitSystem, f.property, f.customType, f.enumName, f.comment,
-                    f.textSize, f.expression, f.remoteLink, f.templateFieldId,
-                    !!f.required, !!f.isArray, !!f.isRef, !!f.readOnly, !!f.autocalculate,
-                    f.enum ?? null,
-                    f.availableOptions ?? null,
-                    f.dependency ?? null,
-                    (f.fields ?? []).map((sub: any) => field(sub, depth + 1)),
-                    (f.conditions ?? []).map((c: any) => condition(c, depth + 1)),
-                ];
-            };
-            const condition = (c: any, depth: number): any => {
-                if (!c || typeof c !== 'object') {
-                    return c ?? null;
+                ancestors.add(value);
+                try {
+                    if (Array.isArray(value)) {
+                        return value.map((item) => walk(item, depth + 1));
+                    }
+                    const out: any = {};
+                    for (const key of Object.keys(value).sort()) {
+                        if (SchemasConfigurationComponent.SIGNATURE_IGNORED_KEYS.has(key)) {
+                            continue;
+                        }
+                        const item = value[key];
+                        /*
+                         * JSON.stringify drops undefined- and function-valued keys, so the
+                         * saved schema cannot tell those from absent ones. Clearing a
+                         * property back to unset has to read as clean, or it leaves a
+                         * dirty flag nothing can ever clear.
+                         */
+                        if (item === undefined || typeof item === 'function') {
+                            continue;
+                        }
+                        out[key] = walk(item, depth + 1);
+                    }
+                    return out;
+                } finally {
+                    ancestors.delete(value);
                 }
-                if (depth > SchemasConfigurationComponent.SIGNATURE_MAX_DEPTH || seen.has(c)) {
-                    throw new Error('unbounded');
-                }
-                seen.add(c);
-                return [
-                    c.ifCondition?.field?.name ?? null,
-                    c.ifCondition?.fieldValue ?? null,
-                    (c.thenFields ?? []).map((f: any) => field(f, depth + 1)),
-                    (c.elseFields ?? []).map((f: any) => field(f, depth + 1)),
-                ];
             };
-            return JSON.stringify([
-                schema.name ?? null,
-                schema.description ?? null,
-                schema.entity ?? null,
-                (schema.fields ?? []).map((f: any) => field(f, 0)),
-                (schema.conditions ?? []).map((c: any) => condition(c, 0)),
-            ]);
+            /*
+             * The walk is generic, but it can only reach what this root names, so the
+             * root is the one remaining allow-list and needs the same scrutiny.
+             * `arrayDependencies` is user-editable (addArrayDependency /
+             * removeArrayDependency, both of which call markDirty) and is persisted on
+             * the model, so editing only an array dependency used to recompute a
+             * signature identical to the baseline and lose the edit.
+             */
+            return JSON.stringify(walk({
+                name: schema.name ?? null,
+                description: schema.description ?? null,
+                entity: schema.entity ?? null,
+                fields: schema.fields ?? [],
+                conditions: schema.conditions ?? [],
+                arrayDependencies: schema.arrayDependencies ?? [],
+            }, 0));
         } catch {
             return null;
         }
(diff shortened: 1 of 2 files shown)


vshvets-bc:
Rebased onto current `develop` — it had gone conflicting against #6749 (schema-editor error surfacing), which touches the same file and has since merged.

Only the spec collided: both changes append an independent `describe` at the same anchor. Both are kept — `backend failures are surfaced` from #6749 and `reverting an edit clears the dirty flag` from this PR. The component itself merged cleanly.

Re-verified after the rebase:
- **43/43 pass** on this branch
- with the component reverted to `develop`, **6 of the 8 new tests fail** — unchanged from the original claim in the description

The 2 that pass either way are still the fail-toward-dirty guards (no baseline, never-saved schemas), which must keep passing.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1109,6 +1115,100 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+                    f.unitSystem, f.property, f.customType, f.enumName, f.comment,
+                    f.textSize, f.expression, f.remoteLink, f.templateFieldId,
+                    !!f.required, !!f.isArray, !!f.isRef, !!f.readOnly, !!f.autocalculate,
+                    f.enum ?? null,
+                    f.availableOptions ?? null,
+                    f.dependency ?? null,
+                    (f.fields ?? []).map((sub: any) => field(sub, depth + 1)),
+                    (f.conditions ?? []).map((c: any) => condition(c, depth + 1)),
+                ];
+            };
```
Pyatakov: When a condition uses AND/OR operators, `ifCondition` is structured as `{AND: [...]}` or `{OR: [...]}`, so reading `ifCondition.field?.name` and `ifCondition.fieldValue` returns `undefined` both times. Changes to predicate rows are invisible to the signature.

**Suggestion:**
- Create a helper to recursively serialize the entire `ifCondition` structure, not just extract two fields
- It should handle:
  - **SINGLE** predicates: `[field.name, fieldValue, fieldPath]`
  - **AND** predicates: `['AND', [...array of predicates...]]`
  - **OR** predicates: `['OR', [...array of predicates...]]`
- Use that in the condition signature

Example approach:
```ts
const predicate = (p: any): any => {
    if (!p || typeof p !== 'object') { return p ?? null; }
    if (Array.isArray(p.AND)) { return ['AND', p.AND.map(predicate)]; }
    if (Array.isArray(p.OR)) { return ['OR', p.OR.map(predicate)]; }
    return [p.field?.name ?? null, p.fieldValue ?? null, p.fieldPath ?? null];
};
```

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1109,6 +1115,100 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+     */
+    private static readonly SIGNATURE_MAX_DEPTH = 12;
+
+    private schemaSignature(schema: Schema | null | undefined): string | null {
+        if (!schema) {
+            return null;
+        }
+        try {
+            const seen = new WeakSet<object>();
+            const field = (f: any, depth: number): any => {
```
Pyatakov: The signature's allow-list is incomplete, these editable properties are missing: `hidden` (`setFieldRequiredMode`), `default`/`suggest`/`examples` (`onRefPresetFormChange`), `textColor`/`textBold` (`resetHelpText`, plus the `textColor` ngModel), `isUpdatable` (`setFieldType`). Switching a field's mode from "none" to "hidden" leaves `required` and `autocalculate` both `false`, so the signature doesn't move and the flag clears.

Beyond adding these, a hand-picked key list is the wrong shape for a check whose failure mode is data loss. Instead of maintaining a hand-picked list (which breaks every time a new editable property is added), derive the signature from the field's own keys – e.g., a sorted walk of `Object.keys(f)` excluding known-transient ones (like internal Angular properties or UI-only state). That way new editable properties are covered by default.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1109,6 +1115,100 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+                if (!c || typeof c !== 'object') {
+                    return c ?? null;
+                }
+                if (depth > SchemasConfigurationComponent.SIGNATURE_MAX_DEPTH || seen.has(c)) {
+                    throw new Error('unbounded');
+                }
+                seen.add(c);
+                return [
+                    c.ifCondition?.field?.name ?? null,
+                    c.ifCondition?.fieldValue ?? null,
```
Pyatakov: When a condition uses AND/OR operators, `ifCondition` is structured as `{AND: [...]}` or `{OR: [...]}`, so reading `ifCondition.field?.name` and `ifCondition.fieldValue` returns `undefined` both times. Changes to predicate rows are invisible to the signature.

**Suggestion:**
- Create a helper to recursively serialize the entire `ifCondition` structure, not just extract two fields
- It should handle:
  - **SINGLE** predicates: `[field.name, fieldValue, fieldPath]`
  - **AND** predicates: `['AND', [...array of predicates...]]`
  - **OR** predicates: `['OR', [...array of predicates...]]`
- Use that in the condition signature

Example approach:
```ts
const predicate = (p: any): any => {
    if (!p || typeof p !== 'object') { return p ?? null; }
    if (Array.isArray(p.AND)) { return ['AND', p.AND.map(predicate)]; }
    if (Array.isArray(p.OR)) { return ['OR', p.OR.map(predicate)]; }
    return [p.field?.name ?? null, p.fieldValue ?? null, p.fieldPath ?? null];
};
```

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1109,6 +1115,100 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+                    return f ?? null;
+                }
+                if (depth > SchemasConfigurationComponent.SIGNATURE_MAX_DEPTH || seen.has(f)) {
+                    throw new Error('unbounded');
+                }
+                seen.add(f);
+                return [
+                    f.name, f.title, f.description, f.type, f.format, f.pattern, f.unit,
+                    f.unitSystem, f.property, f.customType, f.enumName, f.comment,
+                    f.textSize, f.expression, f.remoteLink, f.templateFieldId,
```
Pyatakov: The signature's allow-list is incomplete, these editable properties are missing: `hidden` (`setFieldRequiredMode`), `default`/`suggest`/`examples` (`onRefPresetFormChange`), `textColor`/`textBold` (`resetHelpText`, plus the `textColor` ngModel), `isUpdatable` (`setFieldType`). Switching a field's mode from "none" to "hidden" leaves `required` and `autocalculate` both `false`, so the signature doesn't move and the flag clears.

Beyond adding these, a hand-picked key list is the wrong shape for a check whose failure mode is data loss. Instead of maintaining a hand-picked list (which breaks every time a new editable property is added), derive the signature from the field's own keys – e.g., a sorted walk of `Object.keys(f)` excluding known-transient ones (like internal Angular properties or UI-only state). That way new editable properties are covered by default.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1109,6 +1115,100 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+    private snapshotSchema(schema: Schema | null | undefined, key?: string): void {
+        const id = key || schema?.id || (schema as any)?._id;
+        if (!id) {
+            return;
+        }
+        const signature = this.schemaSignature(schema);
+        if (signature === null) {
+            this.savedSignatures.delete(id);
+        } else {
+            this.savedSignatures.set(id, signature);
```
Pyatakov, line 1232: `dirtySchemaIds` gets pruned after list loads via `pruneDirtySchemaIds()`, but `savedSignatures` only grows. Over time it accumulates signatures for schemas no longer in the list.

**What needs fixing:**
- Extend the existing `pruneDirtySchemaIds()` method (or create a paired `pruneSavedSignatures()`) to also clean up `savedSignatures` against the current schema list
- Call it in the same places where `pruneDirtySchemaIds()` is called (after list reloads)

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1109,6 +1115,100 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+        try {
+            const seen = new WeakSet<object>();
+            const field = (f: any, depth: number): any => {
+                if (!f || typeof f !== 'object') {
+                    return f ?? null;
+                }
+                if (depth > SchemasConfigurationComponent.SIGNATURE_MAX_DEPTH || seen.has(f)) {
+                    throw new Error('unbounded');
+                }
+                seen.add(f);
```
Pyatakov: `seen` is one global WeakSet for the whole traversal, so it rejects any object reached *twice* (not just cycles). When two ref fields point at the same sub-schema IRI, they share the field objects, `field()` visits the same object twice, the signature returns `null`, and `reconcileDirty` leaves the schema permanently dirty.

**Suggestion:**
- Change from a global `seen` set to tracking the ancestor *chain* (the path from root to current node)
- On each recursive call, create a new set from the current path and add the current object
- Only throw if the object is in the ancestor chain (actual cycle), not if it's been seen elsewhere in the tree

Example approach:
```ts
const field = (f: any, depth: number, path: Set<object>): any => {
    if (!f || typeof f !== 'object') { return f ?? null; }
    if (depth > SchemasConfigurationComponent.SIGNATURE_MAX_DEPTH || path.has(f)) { throw new Error('unbounded'); }
    const next = new Set(path);
    next.add(f);
    // ... recurse with `next` instead of `seen`
}
```

Pyatakov (changes requested):

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
vshvets-bc: Adopted, and taken further as you suggested: the signature is now derived from the object's own keys rather than a hand-picked list, ignoring only what the editor recomputes (`errors`, `path`, `fullPath`, `parent`, `controlKey`, `_id`).

That inverts the failure direction, which was the real problem - an unrecognised property now makes the schema look dirty and costs a redundant save, instead of disappearing from the check and costing the user their edit. All seven you named are covered, each with a spec.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
vshvets-bc: Fixed. Rather than a `predicate` helper, the generic walk descends `{AND: [...]}` / `{OR: [...]}` with no special case - so a future operator (or any other nesting) is covered without touching this code again. `thenFields`/`elseFields` targets are reached the same way. Specs added for an edit inside an AND group, a row added to an OR group, and an edit to a then-target.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
vshvets-bc: Fixed, using the ancestor chain as you described. Rather than allocating a new `Set` per node, the object is added before recursing and removed in a `finally`, which is the same semantics at O(1) per visit.

The spec for it builds two ref fields sharing one sub-schema object, which is exactly the case you identified: it returned `null` before and left the schema permanently dirty.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
vshvets-bc: Fixed - `savedSignatures` is now pruned alongside `dirtySchemaIds` in the same method, as you suggested.

One thing worth flagging for the review: extending `pruneDirtySchemaIds()` alone is not sufficient, because it opens with `if (!this.dirtySchemaIds.size) { return; }`. A baseline can outlive the last dirty mark, so the prune would have been skipped in the common case. The guard now covers both maps.

Dropping a baseline can only fail safe: `reconcileDirty` treats a missing one as "cannot prove clean" and leaves the schema dirty, and anything still reachable is re-snapshotted by the next list load.

vshvets-bc:
All four review points are fixed in `c7dac797e`, replied to individually above. The branch is also rebased onto current `develop` (it was 35 commits behind, and its parent commit has since merged).

Three further defects surfaced while making those changes, so flagging them explicitly rather than burying them in the diff:

**`arrayDependencies` was missing from the signature root.** It is user-editable (`addArrayDependency` / `removeArrayDependency`, both of which call `markDirty`) and persisted on the model, but the root object the walk starts from never named it. Editing only an array dependency therefore recomputed a signature identical to the baseline, `reconcileDirty` cleared the flag, `saveAll()` early-returned on `!hasUnsavedChanges`, and the edit was lost on navigation with no toast and no guard dialog. Same silent-data-loss shape as the missing properties above. Worth noting the generic walk does not save us here - the root is the one remaining allow-list, so it needs the same scrutiny.

**`SIGNATURE_MAX_DEPTH` stopped meaning what it did.** The old walk charged one unit per field-nesting level. The generic walk descends every object *and* array level, so a field one level deeper now costs two. Left at `12`, usable nesting would have dropped from ~13 levels to 6 - and past that the walk throws, the signature is `null`, and the schema latches "Unsaved changes" permanently, which is the exact bug this PR exists to remove. Raised to `24` and documented in terms of what it now counts. Scalars return before the check, so they cost nothing.

**Undefined-valued keys had to be skipped.** `JSON.stringify` drops `undefined`- and function-valued keys, so the saved schema cannot distinguish them from absent ones. Without skipping them, clearing a property back to unset left a key the baseline did not have and the flag could never clear again. This one showed up as a genuine test failure rather than by inspection.

Verification: 61 specs pass; 13 of the new ones fail against the previous implementation, one per defect - so none of them is vacuous.

Pyatakov (approved):

PR #6726, opened 2026-08-19
fix(frontend): remove two controls the user cannot use

### LINKED ISSUE #6725: Schema Templates nav item is a dead link for schema-read roles; Export Schema dialog has a dead Copy button and no way to close
Two controls the user cannot use.

## 1. "Schema Templates" is a dead link for a schema-read role

`frontend/src/app/views/new-header/menu.model.ts` pushes the item behind:

```ts
if (
    user.SCHEMAS_SCHEMA_READ ||
    user.SCHEMAS_SYSTEM_SCHEMA_READ
) {
    childItems.push({ title: 'Schema Templates', routerLink: '/schema-templates' });
}
```

but the route requires a **different** permission — `frontend/src/app/app-routing.module.ts` guards `/schema-templates` with `Permissions.TEMPLATES_TEMPLATE_READ`.

So a custom role holding schema-read but not template-read sees the menu entry, clicks it, and is bounced to "Access Restricted" by `PermissionsGuard`.

Note the converse also has to be handled: `TEMPLATES_TEMPLATE_READ` is not in the condition that opens the **Manage** section at all, so simply swapping the gate would hide the item from exactly the roles that *can* use it.

## 2. Export Schema dialog: dead Copy button, and no way to close

`frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.html:32-37`

```html
<button (click)="handleCopyToClipboard(schema.messageId)" [disabled]="!schema.messageId"
  class="button secondary" label="Copy message identifier" pButton></button>
```

A draft has no Hedera message identifier — it is not published — so on every draft this renders permanently greyed out with no explanation of why. There is nothing to copy in that state; the control should not be offered.

The same dialog has **no Cancel/Close button**. The component has the handler:

```ts
onClose(): void {
    this.ref.close(false);
}
```

but nothing in the template calls it. The footer holds only Copy / Save to file / Save to Excel, so the dialog closes only *programmatically*, after a successful save.

## Reproduce

1. As a custom role with schema-read but not template-read, open the nav → "Schema Templates" is listed → clicking it lands on "Access Restricted".
2. Open Export Schema on a DRAFT schema → Copy is greyed out with no explanation, and there is no way to dismiss the dialog short of saving.

vshvets-bc:
Fixes #6725

## Fixes

**Schema Templates nav item** — gate it on `TEMPLATES_TEMPLATE_READ`, the permission `/schema-templates` actually requires, and add that permission to the condition that opens the **Manage** section. Both halves are needed: without the second, the new gate would hide the item from the roles that can use it.

**Export Schema dialog** — show "Copy message identifier" only for a schema that has one, instead of rendering it permanently disabled on every draft; and wire the existing `onClose()` to a Cancel button so the dialog can be dismissed without saving.

## Tests

Two new spec files, 9 tests. Run against `develop` with the source changes reverted, **5 fail**:

- Schema Templates hidden from a schema reader (1)
- Schema Templates hidden from a system-schema reader (2)
- Manage opens for a template reader alone (3)
- Copy hidden on a draft (4)
- Cancel present and closing the dialog (5)

The 4 that pass either way are the companions — Copy still offered on a published schema, both save actions still present, the item still shown when both permissions are held, and an empty menu for a user with no permissions.

### DIFF
--- a/frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.html
+++ b/frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.html
@@ -29,12 +29,19 @@
 
 <div class="dialog-footer">
   <div class="action-buttons">
-    <div class="copy-button">
-      <button (click)="handleCopyToClipboard(schema.messageId)" [disabled]="!schema.messageId"
-        class="button secondary"
-        label="Copy message identifier"
-        pButton
-      ></button>
+    <!-- a draft has no Hedera message identifier, so this used to render permanently
+         greyed out with no explanation -->
+    @if (schema?.messageId) {
+      <div class="copy-button">
+        <button (click)="handleCopyToClipboard(schema.messageId)"
+          class="button secondary"
+          label="Copy message identifier"
+          pButton
+        ></button>
+      </div>
+    }
+    <div>
+      <button (click)="onClose()" class="button secondary" label="Cancel" pButton style="margin-right: 15px"></button>
     </div>
     <div>
       <button (click)="saveToFile()" class="button" label="Save to file" pButton style="margin-right: 15px"></button>
--- a/frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.html
+++ b/frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.html
@@ -29,8 +29,7 @@
 
 <div class="dialog-footer">
   <div class="action-buttons">
-    <!-- a draft has no Hedera message identifier, so this used to render permanently
-         greyed out with no explanation -->
+    <!-- only a published schema has a Hedera message identifier -->
     @if (schema?.messageId) {
       <div class="copy-button">
         <button (click)="handleCopyToClipboard(schema.messageId)"
--- a/frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.spec.ts
+++ b/frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.spec.ts
@@ -29,8 +29,7 @@ describe('ExportSchemaDialog', () => {
             .find((button) => button.getAttribute('label') === label);
     }
 
-    // A draft is not published and so has no Hedera message identifier: the button used
-    // to render permanently greyed out with no explanation of why.
+    // Drafts have no Hedera message identifier since they are not yet published
     it('hides Copy message identifier on a draft', () => {
         expect(buttonByLabel(render({ name: 'S' }), 'Copy message identifier')).toBeUndefined();
     });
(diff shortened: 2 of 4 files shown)


frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.html
```diff
@@ -29,12 +29,19 @@
...
 <div class="dialog-footer">
   <div class="action-buttons">
-    <div class="copy-button">
-      <button (click)="handleCopyToClipboard(schema.messageId)" [disabled]="!schema.messageId"
-        class="button secondary"
-        label="Copy message identifier"
-        pButton
-      ></button>
+    <!-- a draft has no Hedera message identifier, so this used to render permanently
+         greyed out with no explanation -->
```
Pyatakov:
```suggestion
    <!-- only a published schema has a Hedera message identifier -->
```

frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        return fixture;
+    }
+
+    function buttonByLabel(fixture: ComponentFixture<ExportSchemaDialog>, label: string): HTMLButtonElement | undefined {
+        return Array.from(fixture.nativeElement.querySelectorAll('button') as NodeListOf<HTMLButtonElement>)
+            .find((button) => button.getAttribute('label') === label);
+    }
+
+    // A draft is not published and so has no Hedera message identifier: the button used
+    // to render permanently greyed out with no explanation of why.
```
Pyatakov:
```suggestion
    // Drafts have no Hedera message identifier since they are not yet published
```

frontend/src/app/modules/schema-engine/export-schema-dialog/export-schema-dialog.component.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    it('hides Copy message identifier on a draft', () => {
+        expect(buttonByLabel(render({ name: 'S' }), 'Copy message identifier')).toBeUndefined();
+    });
+
+    it('offers Copy message identifier once the schema is published', () => {
+        expect(buttonByLabel(render({ name: 'S', messageId: '1.2.3' }), 'Copy message identifier')).toBeDefined();
+    });
+
+    // onClose() existed but was never wired to anything, so the dialog closed only
+    // programmatically after a successful save.
```
Pyatakov:
```suggestion
    // Users can close the dialog with Cancel instead of only through a successful save.
```

Pyatakov (changes requested):

Pyatakov (approved):

PR #6728, opened 2026-08-19
fix(frontend): put misplaced overlays, tooltips and empty states where they belong

### LINKED ISSUE #6727: Misplaced overlays and tooltips: toolbar tooltips detached, unreadable clipped field path, empty state over the toolbar, detached search icon, un-dimmed modal band
Five layout defects of the same kind — an element positioned against something that is not the box it is meant to sit on.

## 1. Manage Schemas toolbar tooltips render detached from their buttons

`frontend/src/app/views/schemas/schemas.component.html:140-172`

```html
<p-button (click)="onNewInEditor()" class="toolbar-btn-primary"
    [pTooltip]="'Create a Schema'" tooltipPosition="bottom"> ... </p-button>
```

(same shape for Generate Excel Template `:152`, Import `:161`, Compare `:169`)

`pTooltip` on a `<p-button>` binds to the **component host** element, and PrimeNG computes the tooltip position from that host's bounding rect, which does not match the visible button — so the bubble lands well below and left of the control, over empty grid space. These also omit the `tooltipStyleClass="guardian-tooltip"` that the correctly-rendered tooltips elsewhere in the same file use (e.g. `:355`, `:569-571`, which are on plain elements).

## 2. interfaceDocumentsSourceBlock field path is clipped with no tooltip

`frontend/src/app/modules/policy-engine/policy-configuration/blocks/documents/document-viewer-config/document-viewer-config.component.html:51-52`

```html
<td class="propRowCell not-editable-text">
  <span>{{ field.name }}</span>
```

The `not-editable-text` cell clips the content and the span has no tooltip binding, so a value like `document.credentialSubject.0.projectId` renders as `document.credentialS…` and cannot be read at all. The same component already uses the tooltip pattern for other labels (`:22`).

## 3. Manage Policies empty state overlaps the toolbar

`frontend/src/app/modules/policy-engine/policies/policies.component.scss` declares `.not-exist` **twice**:

```scss
.not-exist {           /* :120 */
    position: absolute;
    left: 50%;
    top: 50%;
    transform: translate(-50%, -50%);
    ...
}
...
.not-exist {           /* :923 */
    display: flex;
    flex-direction: column;
    align-items: center;
    justify-content: center;
    ...
}
```

The later rule wins for the properties they share, but `position: absolute` only appears in the first, so it survives — the empty state is lifted out of flow and renders on top of the toolbar action buttons. The second rule already centres it with flex, so the first contributes nothing except the overlap.

## 4. Relayer Accounts search icon renders detached below-left of the input

`frontend/src/app/views/relayer-accounts/relayer-accounts.component.scss:104-114`

```scss
.search-filter {
    width: 400px;
    position: relative;
    .search-icon { position: absolute; right: 12px; top: 12px; }
}
```

The wrapper has no layout of its own and the icon is placed with a magic `top: 12px` that assumes a 40px input, so the icon detaches from the input rather than sitting inside it.

## 5. Add Credential modal leaves an un-dimmed band

`frontend/src/app/components/credentials/credentials-panel/credentials-panel.component.html:154`

```html
<p-dialog header="Add Credential" [(visible)]="showDialog" [modal]="true" [style]="{width:'500px'}" [closable]="true">
```

With no `appendTo`, the modal mask is created next to the dialog inside the page layout, so it covers only that container — leaving a white, un-dimmed band across the bottom of the page. The other `p-dialog` usages in the app append to body.

vshvets-bc:
Fixes #6727

## Fixes

1. **Manage Schemas toolbar tooltips** — anchor each on an `inline-flex` wrapper around the `p-button` instead of on the component host, and add `tooltipStyleClass="guardian-tooltip"` to match the grid-row tooltips that render correctly.
2. **Field path in the documents block config** — add the tooltip the clipped cell needs, using the same pattern already used for the other labels in that component.
3. **Manage Policies empty state** — delete the obsolete first `.not-exist` rule. The later rule already centres it with flex and supersedes every other declaration; all the first contributed was the `position: absolute` that lifted it onto the toolbar. The remaining rule gets `width: 100%` and vertical padding so it flows below the toolbar.
4. **Relayer Accounts search icon** — centre it on the input (`top: 50%` + `translateY(-50%)`) instead of a magic `top: 12px`, make `.search-filter` an explicit flex box so it is unambiguously the offset parent, and reserve `padding-right` so the text cannot run under the icon.
5. **Add Credential dialog** — `appendTo="body"` so the mask spans the viewport, plus the `draggable`/`resizable`/`styleClass` flags the other dialogs use.

## Verification — please read

These are markup and CSS only, so there is nothing meaningful to unit-test; **I have not visually verified them in a browser**. Each change is justified by the mechanism described in #6727 rather than by a screenshot, and the toolbar-tooltip and search-icon fixes in particular would benefit from a reviewer confirming them visually.

I also could not use a full `ng build` as a check: `develop` currently fails to compile in my environment with 27 `TS7053`/`TS7006` errors, all in `frontend/src/app/views/first-steps-panel/first-steps-markdown.ts` (marked's token typings). That file is untouched by this PR and the failure reproduces without these changes.

### DIFF
--- a/frontend/src/app/modules/policy-engine/policies/policies.component.scss
+++ b/frontend/src/app/modules/policy-engine/policies/policies.component.scss
@@ -103,14 +103,9 @@
     font-size: 20px;
 }
 
-.not-exist {
-    position: absolute;
-    left: 50%;
-    top: 50%;
-    transform: translate(-50%, -50%);
-    font-size: 20px;
-    color: darkgrey;
-}
+/* The .not-exist rule further down already centres the empty state with flex and
+   supersedes every declaration this block set; all it contributed was the absolute
+   positioning that lifted the empty state out of flow, on top of the toolbar. */
 
 .ml-8 {
     margin-left: 8px;
@@ -922,6 +917,8 @@
     align-items: center;
     row-gap: 8px;
     justify-content: center;
+    width: 100%;
+    padding: 64px 0;
 }
 
 .info-text {
--- a/frontend/src/app/modules/policy-engine/policies/policies.component.scss
+++ b/frontend/src/app/modules/policy-engine/policies/policies.component.scss
@@ -103,10 +103,6 @@
     font-size: 20px;
 }
 
-/* The .not-exist rule further down already centres the empty state with flex and
-   supersedes every declaration this block set; all it contributed was the absolute
-   positioning that lifted the empty state out of flow, on top of the toolbar. */
-
 .ml-8 {
     margin-left: 8px;
 }
--- a/frontend/src/app/components/credentials/credentials-panel/credentials-panel.component.html
+++ b/frontend/src/app/components/credentials/credentials-panel/credentials-panel.component.html
@@ -151,7 +151,18 @@
   </p-tabs>
 </div>
 
-<p-dialog header="Add Credential" [(visible)]="showDialog" [modal]="true" [style]="{width: '500px'}" [closable]="true">
+<!-- without appendTo the mask is created next to the dialog inside the page layout, so
+     it covers only that container and leaves an un-dimmed band below it -->
+<p-dialog
+    header="Add Credential"
+    [(visible)]="showDialog"
+    [modal]="true"
+    appendTo="body"
+    [draggable]="false"
+    [resizable]="false"
+    styleClass="guardian-dialog"
+    [style]="{width: '500px'}"
+    [closable]="true">
   <div class="dialog-body">
     <div class="guardian-input-container">
       <label class="form-label">Service</label>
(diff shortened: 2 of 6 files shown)


frontend/src/app/modules/policy-engine/policies/policies.component.scss
```diff
@@ -103,14 +103,9 @@
...
-    position: absolute;
-    left: 50%;
-    top: 50%;
-    transform: translate(-50%, -50%);
-    font-size: 20px;
-    color: darkgrey;
-}
+/* The .not-exist rule further down already centres the empty state with flex and
+   supersedes every declaration this block set; all it contributed was the absolute
+   positioning that lifted the empty state out of flow, on top of the toolbar. */
```
Pyatakov: Remove the comment. Over time, without the PR context, it won't be clear what this statement does.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6730, opened 2026-08-19
fix(frontend): say which operation started, and show a failed search once

### LINKED ISSUE #6729: Async progress toast never says which operation started; a failed policy search is shown twice
Two feedback problems.

## 1. "Operation started" says nothing about the operation

Starting any async task (e.g. a data migration) shows a blue toast reading **"Operation started"**, with the action name only as the small title. It reads like a positive confirmation but is info-coloured, and the wording is identical for every async action, so it never identifies what began.

The backend creates the notification with a fixed message for all task actions — `common/src/notification/notification-events.ts:226`:

```ts
this.notify = await notificationHelper.progress(this.action, 'Operation started', this.taskId);
```

and the frontend toasts it verbatim — `frontend/src/app/components/notification/notification.component.ts:91-92`:

```ts
toastProgress(notification: any) {
    this.toastService.info(notification.message, notification.action);
}
```

`this.action` is the TaskAction (e.g. `Migrate data`). Nothing is wrong with the mechanism, but the result is a toast that carries no information.

## 2. A failed policy search is reported twice

Searching for a remote policy that fails shows the same error twice — inline in the Search Policy dialog **and** as a global toast.

`frontend/src/app/modules/policy-engine/dialogs/search-external-policy-dialog/search-external-policy-dialog.component.ts` handles the HTTP error itself and sets an inline message (`this.error = e?.error?.message`, `:48`/`:70`), rendered in the dialog body. Meanwhile the global `HandleErrorsService` interceptor toasts every HTTP error, so `MESSAGE_NOT_FOUND: Message 1 could not be retrieved…` also appears as a "404 Other Error" toast.

There is already a mechanism for this: `SILENT_HTTP_ERRORS` (`frontend/src/app/constants/http-context-tokens.ts`), which the interceptor checks and which `branding.service` and `auth.service` already use for requests they handle themselves.

## Reproduce

1. Start a data migration → blue "Operation started" toast that does not say what started.
2. Manage Policies → Remote Policies → Search, with a message id that fails → the same error inline and as a toast.

vshvets-bc:
Fixes #6729

## Fixes

**Progress toast** — when the message is the generic `'Operation started'` constant (or absent), name the action and phrase it as in-progress. A backend message that is *not* the constant is passed through untouched, so anything more specific a task sends still wins. The toast stays info-styled: it is progress, not success.

Done on the frontend rather than by changing the backend constant, so it applies to every existing task action without touching `notification-events.ts`.

**Duplicated search error** — put `preview` and `import` behind the existing `SILENT_HTTP_ERRORS` context flag, the same mechanism `branding.service` and `auth.service` already use for requests they handle themselves. The dialog keeps its inline message; the global interceptor stops adding a second copy. Every other `external-policy` call keeps the global toast.

## Tests

Two new spec files, 7 tests. Run against `develop` with the source changes reverted, **4 fail**:

- replaces the generic constant with the action
- handles a progress notification with no message at all
- falls back to a heading when there is no action
- marks preview and import as handled inline

The 3 that pass either way pin what must **not** change: a specific backend message is passed through untouched, `preview` still posts the `messageId`, and other external-policy calls still reach the global toast.

### DIFF
--- a/frontend/src/app/services/external-policy.service.ts
+++ b/frontend/src/app/services/external-policy.service.ts
@@ -1,7 +1,8 @@
 import { Injectable } from '@angular/core';
-import { HttpClient, HttpParams, HttpResponse } from '@angular/common/http';
+import { HttpClient, HttpContext, HttpParams, HttpResponse } from '@angular/common/http';
 import { Observable } from 'rxjs';
 import { API_BASE_URL } from './api';
+import { SILENT_HTTP_ERRORS } from '../constants';
 
 /**
  * Services for working from labels and separate blocks.
@@ -48,12 +49,20 @@ export class ExternalPoliciesService {
         return this.http.get<any>(`${this.url}`, { observe: 'response', params });
     }
 
+    /**
+     * The Search Policy dialog renders these failures inline, so the global interceptor
+     * toast is a second copy of the same message.
+     */
+    private silent(): HttpContext {
+        return new HttpContext().set(SILENT_HTTP_ERRORS, true);
+    }
+
     public preview(messageId: string): Observable<any> {
-        return this.http.post<any>(`${this.url}/preview`, { messageId });
+        return this.http.post<any>(`${this.url}/preview`, { messageId }, { context: this.silent() });
     }
 
     public import(messageId: string): Observable<any> {
-        return this.http.post<any>(`${this.url}/import`, { messageId });
+        return this.http.post<any>(`${this.url}/import`, { messageId }, { context: this.silent() });
     }
 
     public approve(messageId: string): Observable<any> {
--- a/frontend/src/app/components/notification/notification.component.spec.ts
+++ b/frontend/src/app/components/notification/notification.component.spec.ts
@@ -1,0 +1,54 @@
+import { NotificationComponent } from './notification.component';
+
+describe('NotificationComponent', () => {
+    let toastService: any;
+
+    function create(): any {
+        toastService = jasmine.createSpyObj('ToastService', ['info', 'success', 'warn', 'error']);
+        return new NotificationComponent(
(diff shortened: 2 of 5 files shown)


frontend/src/app/services/external-policy.service.ts
```diff
@@ -48,12 +49,20 @@ export class ExternalPoliciesService {
...
+
     public preview(messageId: string): Observable<any> {
-        return this.http.post<any>(`${this.url}/preview`, { messageId });
+        return this.http.post<any>(`${this.url}/preview`, { messageId }, { context: this.silent() });
     }
 
     public import(messageId: string): Observable<any> {
-        return this.http.post<any>(`${this.url}/import`, { messageId });
+        return this.http.post<any>(`${this.url}/import`, { messageId }, { context: this.silent() });
     }
```
Pyatakov, line 66: The dialog's error handlers only read `e?.error?.message` (frontend/src/app/modules/policy-engine/dialogs/search-external-policy-dialog/search-external-policy-dialog.component.ts:70 and `:48`), and the template renders step 0 as a bare `{{error}}` div with no fallback. For responses without a JSON `message` – gateway HTML 502, connection reset, timeout – the user now sees an empty error box, and the interceptor toast that used to cover that case is suppressed. Suggested fix in the dialog:

**Line 48**, in `onSubmit`'s error handler:
```ts
this.error = e?.error?.message || 'Import failed. Please try again.';
```
**Line 70**, in `onSearch`'s error handler:
```ts
this.error = e?.error?.message || 'Search failed. Please try again.';
```

Pyatakov (changes requested):

Pyatakov (approved):

PR #6732, opened 2026-08-19
fix(frontend): give Schema Templates the shared paginator

### LINKED ISSUE #6731: Schema Templates grid has no items-per-page control (uses the p-table pager, not the shared paginator)
## Symptom

The Schema Templates list paginates with PrimeNG's default `p-table` pager — just the page arrows. There is no rows-per-page selector, so the page size cannot be changed. Every other grid in the app shows the shared paginator with a 10 / 25 / 50 / 100 / 500 selector, so this grid is both inconsistent and less usable.

## Where

`frontend/src/app/views/schema-templates/schema-templates.component.html:88-89` enables the table's own paginator:

```html
<p-table
    [value]="templates"
    [paginator]="true"
    [rows]="pageSize"
    ...
```

The convention elsewhere is the shared `<app-paginator>` (`frontend/src/app/modules/common/paginator/`), as used by the Schemas grid.

## Note

`SchemaTemplatesComponent.onPage` currently reads PrimeNG's `{first, rows}` payload:

```ts
public onPage(event: any): void {
    this.pageIndex = Math.floor((event.first || 0) / (event.rows || this.pageSize));
    this.pageSize = event.rows || this.pageSize;
    this.loadTemplates();
}
```

`app-paginator` emits `{pageIndex, pageSize}` instead, so the handler has to change with the template. The default `pageSize = 20` is also not one of the paginator's options, so it would be appended to the selector's list.

vshvets-bc:
Fixes #6731

## Fix

Replace `p-table`'s built-in pager with the shared `app-paginator`, using the standard 10/25/50/100/500 options, so the grid gets a rows-per-page control and matches every other grid in the app.

`onPage` now reads `{pageIndex, pageSize}` instead of PrimeNG's `{first, rows}`, and resets to the first page when the size changes — the same behaviour as the Schemas grid.

The default `pageSize` moves 20 → 25 so it is one of the offered options; otherwise `app-paginator.ngOnChanges` appends 20 to its own list and the selector shows a value no other grid offers.

## Tests

New spec covering the payload change: moving page, and the reset-to-first-page on a size change. Both fail on `develop`, where `onPage` reads `event.first` / `event.rows` and leaves `pageIndex` at 0 / `pageSize` unchanged for this payload.

### DIFF
--- a/frontend/src/app/views/schema-templates/schema-templates.component.html
+++ b/frontend/src/app/views/schema-templates/schema-templates.component.html
@@ -271,7 +271,6 @@
               [pageIndex]="pageIndex"
               [pageSize]="pageSize"
               [length]="total"
-              [options]="[10, 25, 50, 100, 500]"
               (page)="onPage($event)"
             ></app-paginator>
           </div>
--- a/frontend/src/app/views/schema-templates/schema-templates.component.html
+++ b/frontend/src/app/views/schema-templates/schema-templates.component.html
@@ -86,11 +86,6 @@
         <div class="guardian-grid-container">
           <p-table
             [value]="templates"
-            [paginator]="true"
-            [rows]="pageSize"
-            [totalRecords]="total"
-            [lazy]="true"
-            (onPage)="onPage($event)"
             class="guardian-grid-table"
             [scrollable]="true">
             <ng-template pTemplate="header">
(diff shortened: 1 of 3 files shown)


frontend/src/app/views/schema-templates/schema-templates.component.html
```diff
@@ -270,6 +265,16 @@
               </tr>
             </ng-template>
           </p-table>
+          <div class="guardian-grid-paginator">
+            <app-paginator
+              class="guardian-grid-paginator"
+              [pageIndex]="pageIndex"
+              [pageSize]="pageSize"
+              [length]="total"
+              [options]="[10, 25, 50, 100, 500]"
```
Pyatakov: `[options]="[10, 25, 50, 100, 500]"` is a fresh array each change-detection run and `PaginatorComponent.ngOnChanges` mutates it in place; no other grid passes `[options]`. Drop it or move it to a `readonly` field.

Pyatakov (changes requested):

Pyatakov (approved):

Issue #5227, opened 2025-08-19
Update IWA TTF properties to v3 and implement a versioning system for backward compatibility


Neurone:
## Problem description

[The current implementation of the IWA TTF properties](https://docs.hedera.com/guardian/guardian/standard-registry/schemas/property-glossary) is not updated with the most recent [v3 version of the standard](https://interworkalliance.github.io/TokenTaxonomyFramework/dmrv/spec/index.html).

In the context of mapping methodologies with the IWA TTF properties correctly, we need to source code to support the current v3 specs, and the future ones.

## Requirements

- Introduce a versioning system for the Guardian IWA TTF properties, so that
  - policy developers can define a specific standard to use for the policy (a single spec for each policy).
  - users can recognise what IWA TTF specs version is used by the policy both reading the network for published policies or via importing the .policy file.
- Flag the current implementation (v1) as _legacy_, so all policies non-compliant with the new versioning system will be recognized by the system as following the legacy specs. 
- Update the list of properties in the source code.
- Update the documentation to support multiple versions of the specs.

prernaadev01:
@Neurone Can you please let us know if you are already working on this one?

Pyatakov:
@Neurone Now that [IWA TTF Version 3.0](https://interworkalliance.github.io/TokenTaxonomyFramework/dmrv/spec/index.html) has been released, it would be great to revise the requirements for this ticket and perhaps include CADT/CDOP as well.

Neurone:
I updated the requirements to comply with v3 and introducing a versioning system.
About CADT/CDOP, I would suggest to still consider them a _transformation_ from IWA TTF v3, more than standards that can be used inside the schema definition.

PR #6739, opened 2026-08-20
feat: Introduce IWA dMRV v3 support and upgrade functionality

gayanath8:
### Issue Label
[https://github.com/hashgraph/guardian/issues/5227](https://github.com/hashgraph/guardian/issues/5227)

## Problem description
[The current implementation of the IWA TTF properties](https://docs.hedera.com/guardian/guardian/standard-registry/schemas/property-glossary) is not updated with the most recent [v3 version of the standard](https://interworkalliance.github.io/TokenTaxonomyFramework/dmrv/spec/index.html).
In the context of mapping methodologies with the IWA TTF properties correctly, we need to source code to support the current v3 specs, and the future ones.

## Requirements

- Introduce a versioning system for the Guardian IWA TTF properties, so that
  - ~~policy developers can define a specific standard to use for the policy (a single spec for each policy)~~ → **implemented at schema level, not policy level**
  - users can recognise what IWA TTF specs version is used both when reading the network for published policies and when importing a `.policy` file. ✅
- Flag the current implementation (v1) as _legacy_, so all policies non-compliant with the new versioning system will be recognized by the system as following the legacy specs. ✅
- Update the list of properties in the source code. ✅
- Update the documentation to support multiple versions of the specs. ✅

## Summary of Changes

A per-schema IWA specification version tag was introduced, the IWA v3.0.0 property namespace was added alongside the existing v1 list, and a guided bulk-remap action was added so an author can move a draft schema from v1 to v3 without editing every field by hand.

- Two new REST endpoints were added under `/schemas/:schemaId/`; **one existing endpoint** gained an optional query parameter. No existing endpoint changed its route, method, or required parameters.
- Full property reference: [Property Glossary](docs/guardian/workspace/schemas/property-glossary.md)
- Every schema now carries an `iwaVersion` tag.
- Absent means v1. Schemas created before this change carry no tag and are read as IWA v1, never as the current default. This is the "legacy" flag the issue asks for.
- New schemas default to v3.
- The tag travels with the schema. It is written to the Hedera SchemaMessage body as iwa_version, so a schema published on one Guardian instance and imported into another keeps its version. `.policy` file imports route through the same SchemaImport path and are defaulted to v1 unless the file says otherwise.
- The tag decides which property list the editor offers, so a schema cannot accidentally mix the two namespaces.
- Every read path resolves per schema, so mixed-version policies work: the field editor calls loadProperties(resolveIwaVersion(schema)), and project search resolves the schema's version before matching properties.

### 1. `GET /api/v1/schemas/:schemaId/iwa-upgrade/preview` *(new)*

Reports, per field, what an upgrade would do. **Does not modify the schema.**

- Requires `SCHEMAS_SCHEMA_READ`
- `404` - schema not found
- `422` - schema is published, or already on v3

**Request:**
```
GET /api/v1/schemas/6a7d5ca0e908b1fee4ffdc22/iwa-upgrade/preview
Authorization: Bearer <accessToken>
```

**Response `200 OK`:**
```json
{
  "renamed": [
    { "field": "postalCode", "from": "Address.zip", "to": "Address.postalCode" },
    { "field": "orgCountry", "from": "AccountableImpactOrganization.country", "to": "ActivityImpactModule.country" },
    { "field": "checkpoints.spanPackage", "from": "ImpactClaimCheckpoint.spanDataPackage", "to": "Checkpoint.dataPackages" }
  ],
  "unchanged": [
    { "field": "projectTitle", "from": "ActivityImpactModule.name", "to": "ActivityImpactModule.name" }
  ],
  "unmappable": [
    { "field": "legacyExtension", "from": "MrvExtension.typedExtension", "to": null }
  ]
}
```
- `renamed`: the property path v3 moved or renamed it to
- `unchanged`: identical in both versions
- `unmappable` : v3 removed the property with no equivalent; `to` is `null`, and applying the upgrade clears it
- Walks nested objects and array items, not just top-level fields, a property inside an array of objects is reported too

image

### 2. `POST /api/v1/schemas/:schemaId/iwa-upgrade` *(new)*

Rewrites every field property to its v3 equivalent and tags the schema `3.0.0`.

- Requires `SCHEMAS_SCHEMA_UPDATE`
- `404` - schema not found
- `422` - schema is published, or already on v3
- Body: `{}` - the endpoint takes no input, but a JSON content-type with a completely empty body is rejected by the gateway

**Request:**
```
POST /api/v1/schemas/6a7d5ca0e908b1fee4ffdc22/iwa-upgrade
Authorization: Bearer <accessToken>
Content-Type: application/json

{}
```

**Response `200 OK`:** the same report as the preview, plus the updated schema.
image

### 3. `GET /api/v1/projects/properties` *(existing)*

Gained an **optional** `iwaVersion` query parameter.

```
GET /api/v1/projects/properties                    → 632 rows (both versions)
GET /api/v1/projects/properties?iwaVersion=1.0.0   → 273 rows
GET /api/v1/projects/properties?iwaVersion=3.0.0   → 359 rows
GET /api/v1/projects/properties?iwaVersion=9.9.9   → 0 rows (no fallback to the full list)
```

Each row now also carries its `iwaVersion`:
```json
{ "title": "ImpactClaim.startDate", "value": "Start Date", "iwaVersion": "3.0.0", "id": "..." }
```

image
image
image

---

## Impact on existing endpoints

| Endpoint | Impact |
|---|---|
| `GET /projects/properties` | **Behaviour change when called without the new parameter** |
| `POST /projects/search` | None. The three IWA paths the project index uses (`AccountableImpactOrganization.name`, `ActivityImpactModule.projectScope`, `ActivityImpactModule.name`) are identical in v1 and v3, so schemas of either version resolve unchanged. |
| `GET /schema/:schemaId`, `GET /schemas`, `GET /schemas/list/all` | Additive only, responses gain an `iwaVersion` field. No field removed or renamed. |
| `POST /schemas/:topicId` (create) | Additive, accepts an optional `iwaVersion`. Omitted ⇒ defaults to `3.0.0`. |
| Schema publish / import | Additive, `iwa_version` added to the Hedera message body. Messages published before this change simply lack the key and resolve to v1. |

### Schema editor version-aware property dropdown

The **Property** dropdown lists only the properties belonging to the schema's own IWA version. A property already saved on a field but absent from the current version's list is preserved and still displayed, so opening and saving a v1 schema cannot silently blank its property values.

Version 1 List
image

Version 3 List
image

### Schema editor "Upgrade to IWA v3" action

Shown on draft schemas that are still on v1. Opens a confirmation dialog built from the existing `compare-schema` and `guardian-grid-table` patterns:
Nothing is written until **Upgrade** is confirmed.

image
image
image

## DB Changes 
New migration v3-7-0
1. Tags every existing `policy_property` row as `1.0.0`
2. Seeds the 359 v3 property rows, **guarded by a lookup so re-running cannot duplicate them** (the predecessor `v2-18-0` inserts unconditionally and would)
3. Tags every existing `schema` document as `1.0.0`

Two nullable columns added `schema.iwaVersion` and `policy_property.iwaVersion`. Both nullable because rows written before this migration have no such field.

### Unit Tests

image


gayanath8:
Extended Implementation of [https://github.com/hashgraph/guardian/issues/5227](https://github.com/hashgraph/guardian/issues/5227)

## Problem description

Building on the versioned IWA property list, schema developers still have to tag every field's glossary property entirely by hand, picking from a flat catalogue of 359 (v3) / 273 (v1) machine readable paths such as `ActivityImpactModule.projectScope`. The catalogue has no descriptions, and there is no assistance for which property a given field should use.

## Requirements

- Suggest, per schema field, which glossary property it should be tagged with, both while building a new schema and when reviewing an existing one. ✅
- Never let the AI invent a property that does not exist in the real catalogue. ✅
- Never block schema editing if the AI is unavailable, the manual property picker must keep working exactly as before. ✅
- Add real descriptions to the property catalogue, sourced from the actual IWA dMRV specification, and surface them wherever a property is shown. ✅

## Summary of Changes

An AI suggestion engine was added end to end, given a schema's fields, ai-service returns ranked glossary-property candidates per field, each with a confidence score, a rationale, and its description. This is surfaced in the schema builder as a new Glossary AI tab plus a per-field quick-suggest action next to the existing Property dropdown. The `PolicyProperty` catalogue itself was enriched with real descriptions and threaded through both the manual picker and the AI path.

- image

- image


- [ ] One new REST endpoint was added under /ai-suggestions/ and one existing endpoint gained an additive field on its response.
- [ ] The authoritative property list for the schema's iwaVersion is always fetched server-side inside ai-service, never trusted from the client, and the model is constrained to a JSON-Schema enum of the real property titles, so a hallucinated property name is structurally impossible.
- [ ] The rationale shown to the user is always one of four fixed sentence templates (name / description / type / current-match)
- [ ] On any failure in the LLM call (bad key, rate limit, network) the response degrades to `{available:false, results:[]}` instead of throwing, verified live against an actual provider quota error.

## API Changes

### 1. `POST /api/v1/ai-suggestions/schema-properties` *(new)*

Returns ranked glossary-property candidates for one or more schema fields.
**Authentication:** Bearer token required. **Permission:** `Permissions.SCHEMAS_SCHEMA_CREATE`, the same permission required to create a schema. Requests without it get `401`.

**Request:**
```
POST /api/v1/ai-suggestions/schema-properties
Content-Type: application/json

{
  "schemaId": "64f1e2b3c4d5e6f7a8b9c0d1",
  "fieldNames": ["orgCountry"]
}
```
`schemaTitle`, `iwaVersion` and the per-field payload (`title`/`description`/`type`/`currentProperty`) are no longer part of the request — `ai-service` fetches the schema itself by `schemaId` and derives all of that from it, so the client can't influence what context the model sees.

**Response `200 OK`:**
```json
{
  "available": true,
  "results": [
    {
      "fieldName": "orgCountry",
      "candidates": [
        {
          "title": "Address.country",
          "confidence": 0.9,
          "rationale": "The field name \"orgCountry\" corresponds to Address.country.",
          "description": "The country of the address."
        }
      ]
    }
  ]
}
```
- `available: false` means the AI is temporarily unavailable, not an error, `results` is `[]` and the caller should fall back to manual tagging
- `candidates` is ranked best-first, capped at 3, and may be empty for a field with no confident match
- `description` is omitted when the matched property has none (all current IWA v1 properties)

image

### 2. `GET 1` *(existing)*

Each row now also carries its `description` when one exists:
```json
{ "title": "Address.country", "value": "Country", "iwaVersion": "3.0.0", "description": "The country of the address.", "id": "..." }
```

image

---

## Impact on existing endpoints

| Endpoint | Impact |
|---|---|
| `GET /projects/properties` | Additive only, responses gain an optional `description` field. No field removed or renamed. |
| `GET /ai-suggestions/ask`, `PUT /ai-suggestions/rebuild-vector` | None, the new suggestion path is entirely separate from the existing RAG/vector methodology-search feature and doesn't touch its vector store or chain state. |

## UI changes

### Schema editor "Glossary AI" tab

A new pill tab alongside Fields / Conditions / Repeatable links. Suggestions load automatically for every field in the schema and are cached per schema/sub-schema for the session, switching schemas or drilling into a sub-schema and back restores previous results instantly instead of re-querying the AI. Each row shows the field, its current property, the AI's top candidate with confidence and description, and a rationale line, with Accept / Change / Dismiss per row and an "Apply all high-confidence (N)" bulk action.

image

### Per-field quick suggest

A sparkle button next to the existing **Property** dropdown runs the same suggestion for just the selected field. "See more" expands the remaining ranked candidates inline; "Move to Glossary AI" switches to the full tab with that field's row highlighted.

image

### Property descriptions everywhere

The manual Property dropdown, the Glossary AI tab, and the per-field hint all now show each property's description alongside its name and, in the AI views, the rationale, styled distinctly (name bold, description muted, rationale muted italic) so the three read clearly at a glance.

image
image


### Graceful degradation

If `ai-service` is down or the model call fails for any reason, the tab and the per-field hint show a clear "unavailable" message; the manual Property dropdown is completely unaffected.

image

## DB Changes

New migration `v3-8-0`
1. Reads the property CSVs with a quote-aware parser (descriptions contain commas) and backfills `description` onto already-seeded `policy_property` rows, matched by `title` (+ `iwaVersion`).
2. Enrichment only, never inserts a row, never overwrites a description with a blank one.
3. Idempotent and safe to re-run.

One nullable column added, `policy_property.description`. Nullable because rows written before this migration, and all IWA v1 rows, have no such field.

guardian-service/src/migrations/v3-8-0.ts
Pyatakov, line 1: This should be moved to v3.7.1.ts

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -549,12 +636,224 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+        this.rightPanelSuggestLoading = true;
+        this.rightPanelSuggestUnavailable = false;
+        const contextSchema = this.currentContextSchema;
+        const request = {
+            schemaId: contextSchema?.id || (contextSchema as any)?._id,
+            fieldNames: [field.name]
+        };
+        this.aiSearchService.suggestSchemaProperties(request)
+            .pipe(takeUntil(this.destroy$))
+            .subscribe({
```
Pyatakov, line 799: The right-panel suggestion response is assigned with no check that `selectedField` is still the field it was requested for. The `selectedField` setter clears the spinner but does not cancel the request, and the template's Apply button calls `acceptSuggestion(selectedField.name, …)`, so selecting another field during the (could be up to 45s) AI call silently writes the first field's suggested IWA property onto the second.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -549,12 +636,224 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+        const key = this.getContextSchemaCacheKey(schema);
+        this.suggestionsLoading = true;
+        const request = {
+            schemaId: schema?.id || (schema as any)?._id,
+            fieldNames: fields.map((field) => field.name)
+        };
+        this.aiSearchService.suggestSchemaProperties(request)
+            .pipe(takeUntil(this.destroy$))
+            .subscribe({
+                next: (response) => {
```
Pyatakov, line 737: `loadSuggestions()` returns from both handlers *before* clearing `suggestionsLoading` when the context key changed, and nothing else ever sets it false. Switching schema (or drilling into a sub-schema) mid-load leaves the tab spinning forever, Refresh and "Apply all high-confidence" permanently disabled, and `autoLoadSuggestionsFor()` blocked for the rest of the session.

common/src/database-modules/database-server.ts
```diff
@@ -3266,8 +3266,32 @@ export class DatabaseServer extends AbstractDatabaseServer {
...
+    /**
+     * Build the policy-property lookup filter for an IWA version.
+     *
+     * Rows seeded before IWA v3 support carry no iwaVersion, so a v1 request
+     * must also match untagged rows — otherwise the property list comes back
+     * empty on an instance where the backfill migration has not run yet.
+     */
+    private static buildPolicyPropertyFilter(iwaVersion?: string): FilterObject<PolicyProperty> {
+        if (!iwaVersion) {
+            return {};
```
Pyatakov, line 3284: An unversioned lookup returns `{}`, i.e. v1 *and* v3 rows merged now that v3-7-0 seeds both into `policy_property`. `SchemaService.properties()` was left unversioned, so the Policy Labels, Statistic Definition and Schema Rules property pickers get both namespaces at once.
I'm not sure if that was intentional, but worth to highlight.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -549,12 +636,224 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+                        },
+                    });
+                    if (!dialogRef) { return; }
+                    dialogRef.onClose.pipe(takeUntil(this.destroy$)).subscribe((result: string) => {
+                        if (result !== 'Upgrade') { return; }
+                        this.schemaService.upgradeToIwaV3(id)
+                            .pipe(takeUntil(this.destroy$))
+                            .subscribe({
+                                next: () => this.schemaLoad$.next(id),
+                                error: () => {}
```
Pyatakov, line 688: Both IWA-upgrade subscriptions use `error: () => {}`. The 403/422s from `loadUpgradableSchema` are reachable from a stale UI, and the button then appears to do nothing. The component already has `showError()`.

frontend/src/app/modules/schema-engine/iwa-upgrade-dialog/iwa-upgrade-dialog.component.html
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  <!-- Colour keys double as filters, as in the compare-schema view.
+       Real buttons so they are keyboard-reachable and announce their state. -->
+  <div class="legend">
+    @if (removedCount > 0) {
+      <button type="button" class="legend-item" [class.is-off]="!showRemoved"
+              [attr.aria-pressed]="showRemoved" (click)="toggle('removed')">
+        <span class="legend-icon legend-removed"></span>
+        <span class="legend-text">No v3 equivalent &mdash; will be cleared ({{ removedCount }})</span>
```
Pyatakov, line 35: `&mdash;` in UI text; same at schemas-configuration.component.html:1659.

Pyatakov (changes requested):

Pyatakov:
@gayanath8 please resolve merge conflicts and address review comments. Thanks!

frontend/src/app/modules/schema-engine/iwa-upgrade-dialog/iwa-upgrade-dialog.component.html
gayanath8: Fixed. replaced &mdash with UI dash text

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
gayanath8: Fixed. Added up reportError toast

common/src/database-modules/database-server.ts
gayanath8: This was intentional

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
gayanath8: Fixed. Clear loading flag without condition

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
gayanath8: Fixed. Added stale field response guard

guardian-service/src/migrations/v3-8-0.ts
gayanath8: Merged into v3-7-1.ts

gayanath8:
Hi @Pyatakov @Neurone 
 Related to the Issue #5227, merge conflict has been resolved, and the review comments have also been addressed. Let me know if anything needs to be changed.
 
 Thanks

frontend/src/app/modules/schema-engine/iwa-upgrade-dialog/iwa-upgrade-dialog.component.html
Pyatakov: @gayanath8 "—" is still exists at iwa-upgrade-dialog.component.html:35 and schemas-configuration.component.html:1834, should be replaced with "–".

Pyatakov (approved):

frontend/src/app/modules/schema-engine/iwa-upgrade-dialog/iwa-upgrade-dialog.component.html
gayanath8: @Pyatakov can you check now two places have replaced "—" with "–".

PR #6741, opened 2026-08-20
fix(policy-service): a group restriction with no group must not match everything

### LINKED ISSUE #6740: documentValidatorBlock: an "own group only" source restriction matches every document when the user has no group
## Summary

`documentValidatorBlock`'s source filter drops its group restriction entirely when the user has no active group, so a validation that is meant to narrow the search to the user's own group instead widens it to **every document in the policy**.

## Where

`policy-service/src/policy-engine/blocks/document-validator-block.ts`

```ts
if (sourceValidation.onlyOwnByGroupDocuments && user?.group) {
    filter.group = { $eq: user.group };
}
...
if (sourceValidation.onlyAssignByGroupDocuments && user?.group) {
    filter.assignedToGroup = { $eq: user.group };
}
```

The `&& user?.group` guard means: when the user has no group, no clause is added at all.

## Why it matters

A user has no active group in ordinary situations — a non-group policy, or any point before group selection. In those cases an entry configured "own group only" runs with no group filter, so **another participant's document can satisfy it**. The restriction exists precisely to prevent that.

It is also inconsistent within the same block: the direct document checks fail *closed* when the user has no group, while these fail *open*.

## Expected

When the restriction is configured but there is no group to scope to, the filter should match nothing rather than everything — an impossible filter is the honest translation of "only my group" when there is no group.

## Reproduce

Configure a `documentValidatorBlock` source validation with **Only own by group documents**, and run it as a user with no active group. Every document in the policy is a candidate; one owned by a different participant satisfies the validation.

vshvets-bc:
Fixes #6740

## Fix

Apply the group clauses whenever the restriction is configured, scoping to the user's group when there is one and to an impossible filter when there is not:

```ts
if (sourceValidation.onlyOwnByGroupDocuments) {
    filter.group = user?.group ? { $eq: user.group } : { $in: [] };
}
```

Same for `onlyAssignByGroupDocuments`. This matches how the direct document checks in the same block already behave — they fail closed when the user has no group.

## Tests

New `policy-service/tests/unit-tests/blocks/document-validator-source-filter.test.mjs`, 4 tests. On `develop` with the source change reverted, **the two fail-open cases fail**:

- "own group only" with no active group matches nothing rather than everything
- same for the assigned-by-group variant

The other two pin what must not change: the filter still scopes to the group when the user has one, and no clause is added when the option is off.

## Note

This came out of a wider audit of the block's source validations. Two other findings from the same audit are separate: the `in`/`not_in` ordering, already fixed in #6714, and a one-sided type-coercion mismatch in the DB filter, which I can raise separately if useful.

### DIFF
--- a/policy-service/src/policy-engine/blocks/document-validator-block.ts
+++ b/policy-service/src/policy-engine/blocks/document-validator-block.ts
@@ -108,14 +108,24 @@ export class DocumentValidatorBlock {
         if (sourceValidation.onlyOwnDocuments && user?.did) {
             filter.owner = { $eq: user.did };
         }
-        if (sourceValidation.onlyOwnByGroupDocuments && user?.group) {
-            filter.group = { $eq: user.group };
+        /*
+         * A group restriction with no group must not widen the search. These used to be
+         * skipped entirely when the user had no active group (a non-group policy, or
+         * before group selection), so an "own group only" entry ran with NO group filter
+         * and every document in the policy became a candidate - another participant's
+         * document could satisfy it. The direct document checks in this same block fail
+         * closed in that situation, so it was inconsistent as well as fail-open. An
+         * impossible filter is the honest translation of "only my group" when there is
+         * no group.
+         */
+        if (sourceValidation.onlyOwnByGroupDocuments) {
+            filter.group = user?.group ? { $eq: user.group } : { $in: [] };
         }
         if (sourceValidation.onlyAssignDocuments && user?.did) {
             filter.assignedTo = { $eq: user.did };
         }
-        if (sourceValidation.onlyAssignByGroupDocuments && user?.group) {
-            filter.assignedToGroup = { $eq: user.group };
+        if (sourceValidation.onlyAssignByGroupDocuments) {
+            filter.assignedToGroup = user?.group ? { $eq: user.group } : { $in: [] };
         }
 
         for (const f of (sourceValidation.filters || [])) {
--- a/policy-service/src/policy-engine/blocks/document-validator-block.ts
+++ b/policy-service/src/policy-engine/blocks/document-validator-block.ts
@@ -113,19 +113,24 @@ export class DocumentValidatorBlock {
          * skipped entirely when the user had no active group (a non-group policy, or
          * before group selection), so an "own group only" entry ran with NO group filter
          * and every document in the policy became a candidate - another participant's
-         * document could satisfy it. The direct document checks in this same block fail
-         * closed in that situation, so it was inconsistent as well as fail-open. An
-         * impossible filter is the honest translation of "only my group" when there is
-         * no group.
+         * document could satisfy it.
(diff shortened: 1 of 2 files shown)


policy-service/src/policy-engine/blocks/document-validator-block.ts
```diff
@@ -108,14 +108,24 @@ export class DocumentValidatorBlock {
...
+        /*
+         * A group restriction with no group must not widen the search. These used to be
+         * skipped entirely when the user had no active group (a non-group policy, or
+         * before group selection), so an "own group only" entry ran with NO group filter
+         * and every document in the policy became a candidate - another participant's
+         * document could satisfy it. The direct document checks in this same block fail
+         * closed in that situation, so it was inconsistent as well as fail-open. An
+         * impossible filter is the honest translation of "only my group" when there is
+         * no group.
+         */
```
Pyatakov: Compact justification

policy-service/src/policy-engine/blocks/document-validator-block.ts
```diff
@@ -108,14 +108,24 @@ export class DocumentValidatorBlock {
...
+        if (sourceValidation.onlyOwnByGroupDocuments) {
+            filter.group = user?.group ? { $eq: user.group } : { $in: [] };
```
Pyatakov: In a policy that has no groups, this makes the source validation unsatisfiable while the block's own direct check on the same document still accepts it.
The inconsistency is mirrored, not removed: on a group-less policy the direct check accepts the document and the source validation can never find a source.
Suggested alternative that keeps the security fix (documents with a real group uuid still never match a group-less user):
```suggestion
            filter.group = { $eq: user?.group ?? null };
```

Pyatakov (changes requested):

policy-service/src/policy-engine/blocks/document-validator-block.ts
```diff
@@ -108,14 +108,29 @@ export class DocumentValidatorBlock {
...
+         * document could satisfy it.
+         *
+         * Match the direct checks rather than inverting them: `document.group !== userGroup`
+         * (:353) and `document.assignedToGroup !== userGroup` (:363) both ACCEPT when
+         * neither side has a group, which is every document in a group-less policy. An
+         * impossible filter ($in: []) would reject exactly what they accept - the same
+         * inconsistency mirrored, not removed. `$eq: null` matches a null or missing
+         * group, so a group-less user sees group-less documents and still never matches
+         * a document carrying a real group uuid.
+         */
```
Pyatakov: Short 1-2 lines comment explaining what was changed is enough

vshvets-bc: Trimmed to two lines (e0cc9ad64).

Pyatakov (approved):

PR #6743, opened 2026-08-20
fix(interfaces): enforce a condition's cross-schema targets

### LINKED ISSUE #6742: validateConditionFields ignores a condition's cross-schema targets, so branch-inactive fields go unchecked
## Summary

`SchemaHelper.validateConditionFields` — the runtime condition check reached from `VCJS.conditionErrors` — validates only `thenFields` / `elseFields`, the fields a condition declares itself, looked up by name in the local object:

```ts
const active = evaluate(condition) ? thenFields : elseFields;
for (const field of active) {
    if (field.required && !present(data[field.name])) {
        errors.push(`Field "${field.name}" is required.`);
    }
}
```

It never reads `thenTargets` / `elseTargets` — the **cross-schema** half of a condition, where the target lives at a `fieldPath` inside another sub-schema.

## Why that leaves a gap

`buildDocument` does compile targets: `buildCrossRequired` emits nested `properties` / `required`, and `buildCrossForbidden` emits `false` for the other branch's targets. So targets are enforced wherever those compiled keywords apply — and unenforced everywhere they do not, with nothing in the runtime check to cover the difference.

The consequence that matters is the forbidden side: a document can carry a target belonging to the **inactive** branch and nothing rejects it.

## Reproduce

Condition: *if `kind = full` then `sub.detail` appears* (target `detail` inside sub-schema `sub`).

```js
SchemaHelper.validateConditionFields(conditions, { kind: 'partial', sub: { detail: 'smuggled' } });
// actual: []   expected: an error - the branch that reveals `detail` is not active
SchemaHelper.validateConditionFields(conditions, { kind: 'full', sub: {} });
// actual: []   expected: an error - `detail` is required on the active branch
```

## Out of scope here, but worth recording

A condition whose trigger and target sit in two **paired repeatable** sub-schemas (`subA[i].type` → `subB[i].detail`) cannot be validated by this function as it stands: correlating entry *i* of one array with entry *i* of another needs the schema's `arrayDependencies`, and `validateConditionFields(conditions, data)` does not receive them. Closing that would mean changing the signature, so I have left it out of the fix rather than half-doing it.

vshvets-bc:
Fixes #6742

## Fix

Validate `thenTargets` / `elseTargets` with the same semantics `buildDocument` compiles them to:

- on the **active** branch, a required target must be present (`buildCrossRequired`)
- the **inactive** branch's targets must be absent (`buildCrossForbidden`)
- optional targets are only ever hidden, never required
- a target with no usable `fieldPath` is ignored

Targets are resolved with the `resolve()` helper already in the function, so no new path logic is introduced.

## Tests

New `interfaces/tests/schema-condition-cross-targets.test.mjs`, 6 tests. On `develop` with the source change reverted, **2 fail**:

- a required target missing on the active branch
- a target smuggled in while its branch is inactive ← the one that matters

The other 4 pin what must not change: a satisfied target passes, an absent target on an inactive branch passes, optional targets stay unenforced, and a target with no path is ignored.

The 89 existing tests across `schema-condition-reachability`, `json-to-schema-conditions` and `schema-helper-conditions-serialize` still pass.

## Deliberately not included

The paired-repeatable case (`subA[i].type` → `subB[i].detail`) — correlating entry *i* across two arrays needs the schema's `arrayDependencies`, which this function does not receive. It would need a signature change, so it is described in #6742 rather than half-implemented here.

### DIFF
--- a/interfaces/src/helpers/schema-helper.ts
+++ b/interfaces/src/helpers/schema-helper.ts
@@ -665,6 +665,22 @@ export class SchemaHelper {
             const thenFields = condition.thenFields || [];
             const elseFields = condition.elseFields || [];
 
+            /*
+             * thenTargets / elseTargets are the cross-schema half of a condition: a
+             * target lives at a path inside another sub-schema rather than being a field
+             * of the condition itself. buildDocument compiles them into nested
+             * properties/required (buildCrossRequired) and `false` (buildCrossForbidden),
+             * so they are enforced for a plain object instance, but nothing here checked
+             * them - which left them unverified wherever the compiled keyword cannot
+             * apply.
+             */
+            const thenTargets = (condition as any).thenTargets || [];
+            const elseTargets = (condition as any).elseTargets || [];
+            const targetPath = (t: any): string[] | null =>
+                (t?.fieldPath?.length > 1) ? t.fieldPath : null;
+            const targetName = (t: any): string =>
+                t?.field?.name || (t?.fieldPath ? t.fieldPath[t.fieldPath.length - 1] : '');
+
             if (!reachable) {
                 for (const field of [...thenFields, ...elseFields]) {
                     if (present(data[field.name])) {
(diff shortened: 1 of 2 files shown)


dariy-rised:
Hello. I don't think #6742 reproduces; can you check my tests [there](https://github.com/hashgraph/guardian/issues/6742#issuecomment-5441806056). Both repro cases already return 422 on the real submission path.

Two notes if it proceeds anyway:

- On object containers it doubles the error — ajv already returns `must have required property 'detail'`, this adds `Field "detail" is required.` for the same violation.
- With a repeatable container it rejects valid documents: `resolve()` (`schema-helper.ts:615`) only walks object keys, so `{ kind: 'full', field_1: [{ detail: 'a' }, { detail: 'b' }] }` → `Field "detail" is required.` even though every entry has it. That lands in `conditionErrors`, which flips `res.ok` and throws at `request-vc-document-block.ts:247` — so submission fails with no way to clear the error (drafts still save).

vshvets-bc:
Closing — @dariy-rised is right that #6742 doesn't reproduce, and verifying it turned up more against this PR than the review found. Full analysis in https://github.com/hashgraph/guardian/issues/6742#issuecomment-5451888116.

Short version:

- The cross-schema targets **are** enforced, via `buildCrossRequired`/`buildCrossForbidden` compiled into the `allOf` branch plus `prepareSchema`'s def-clone and `$ref` repoint. My repro called `validateConditionFields` directly, where an empty result never meant the document was accepted.
- On an object container this PR double-reports alongside ajv.
- On a **repeatable** container it is wrong in both directions: it rejects a valid document (every entry carries the target) and stays silent on the smuggled one — because `resolve()` walks named keys and an array returns `undefined` for one.

The real gap is that neither builder emits `items`, so both compiled branches go inert against a repeatable container. That needs its own fix; a path-walking runtime check is not it.

Thanks for the careful read on both.

PR #6745, opened 2026-08-20
fix(policy-service): match source filter values as stored, not only as coerced

### LINKED ISSUE #6744: documentValidatorBlock source filters coerce only one side, so they miss string-stored values (fail-closed and fail-open)
## Summary

`documentValidatorBlock` coerces a filter's value before building the Mongo query, but the stored document is not coerced. Mongo comparisons are type-bracketed, so a coerced number never matches a string-stored value — and the mismatch breaks in **both** directions.

## Where

`policy-service/src/policy-engine/blocks/document-validator-block.ts`, in `buildSourceFilter`:

```ts
const value = this.coerceValue(raw);

switch (f.type) {
    case 'not_equal': filter[f.field] = { $ne: value }; break;
    ...
    case 'gte':       filter[f.field] = { $gte: value }; break;
```

`coerceValue` turns `'100'` into the number `100` and `'2024-06-01'` into epoch milliseconds. VC JSON stores those fields as **strings**.

## Both failure directions

- **fail-closed** — `gte '2024-01-01'` compares epoch-ms against strings and matches nothing, so valid source documents are never found and the validation fails for documents that should pass.
- **fail-open** — `$ne` / `$nin` against string-stored values matches nothing to exclude, so the documents the filter exists to exclude remain candidates.

The in-memory condition phase (`evaluateFieldCondition`) coerces **both** sides, so the two phases disagree — which is what makes the DB phase's one-sided coercion a mismatch rather than a deliberate choice.

## Reproduce

A source validation filtering `document.credentialSubject.0.date` with `gte` `'2024-01-01'`, against documents whose `date` is stored as `'2024-06-01'`: no source document matches, and the block reports "No matching source documents found".

The same filter with `not_equal` `'100'` against `'100'` stored as a string excludes nothing.

## Expected

The filter should match the value as stored as well as coerced. Note range operators cannot express both in a single field predicate — type bracketing again — so they need `$or`.

vshvets-bc:
Fixes #6744

## Fix

Where `coerceValue` actually changes the value, offer both representations:

- equality and `in` → `$in` over both; `not_equal` and `not_in` → `$nin` over both
- range operators cannot express both in one field predicate (type bracketing), so they go through `$and` / `$or` and qualify a document that compares true as **either** type
- where coercion changes nothing, the single predicate is kept unchanged

## Tests

New `policy-service/tests/unit-tests/blocks/document-validator-filter-coercion.test.mjs`, 5 tests. On `develop` with the source change reverted, **4 fail**:

- both representations offered for equality
- both excluded for `not_equal` (the fail-open direction)
- a date range compared as either type (the fail-closed direction)
- every element of an `in` list widened

The fifth pins what must not change: when coercion changes nothing, a single predicate is still emitted and no `$and` is added.

All 17 tests across the block's existing suites pass.

## Related

Same audit as #6740 (group fail-open) and #6714 (`in`/`not_in` ordering). This is the last of the four findings that reproduces upstream; the fourth was a missing custom Fail Message, which is a feature rather than a fix.

### DIFF
--- a/policy-service/src/policy-engine/blocks/document-validator-block.ts
+++ b/policy-service/src/policy-engine/blocks/document-validator-block.ts
@@ -118,34 +118,63 @@ export class DocumentValidatorBlock {
             filter.assignedToGroup = { $eq: user.group };
         }
 
+        /*
+         * The DB filter has to match the value as STORED, not only as coerced.
+         *
+         * coerceValue turns '100' into the number 100 and '2024-06-01' into epoch
+         * milliseconds, but VC JSON stores those fields as strings and Mongo comparisons
+         * are type-bracketed. So `gte '2024-01-01'` compared a number against strings and
+         * matched nothing - fail-closed, valid sources never found - while $ne against
+         * string-stored values matched nothing to exclude and left the excluded documents
+         * as candidates: fail-open. The in-memory condition phase coerces both sides,
+         * which is what makes the DB phase's one-sided coercion a mismatch.
+         *
+         * Where coercion actually changes the value, both representations are offered.
+         */
+        const alternatives = (rawValue: any, coerced: any): any[] =>
+            (rawValue === coerced || rawValue === undefined || rawValue === null)
+                ? [coerced]
+                : [coerced, rawValue];
+        const rangeFilters: any[] = [];
+
         for (const f of (sourceValidation.filters || [])) {
             const raw = f.typeValue === 'variable'
                 ? PolicyUtils.getObjectValue(document, f.value)
                 : f.value;
             const value = this.coerceValue(raw);
+            const both = alternatives(raw, value);
 
             switch (f.type) {
-                case 'not_equal': filter[f.field] = { $ne: value }; break;
-                case 'in': {
-                    const arr = f.typeValue === 'variable'
-                        ? (Array.isArray(raw) ? raw.map((e: any) => this.coerceValue(e)) : [value])
-                        : String(f.value).split(',').map((v: string) => this.coerceValue(v.trim()));
-                    filter[f.field] = { $in: arr };
-                    break;
-                }
+                case 'not_equal': filter[f.field] = { $nin: both }; break;
+                case 'in':
                 case 'not_in': {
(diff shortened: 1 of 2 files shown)


dariy-rised:
`buildSourceFilter` writes `filter[f.field]` directly for `equal/not_equal/in/not_in`.
This PR fixes the same-field-overwrite problem for gte/lte (and gt/lt) when the values need widening, since those now route through $and instead of a direct key write, so multiple range filters on one field compose correctly. But that fix doesn't extend to the other operators, or to range filters whose value doesn't need widening, e.g. `projectId gte 'PRJ-100' `+ `projectId not_equal 'PRJ-150'` (accept an ID range but exclude one flagged project) would lose one constraint, since alphanumeric IDs never trigger widening. Might be worth a follow-up issue to cover the remaining cases.

dariy-rised:
Would be good to add a test with typeValue: 'variable', none of the current tests exercise that path.

```
  it('widens a variable-resolved range value', () => {
      const document = { currentAmount: '100' };
      const filter = block().buildSourceFilter(
          { filters: [{ field: 'amount', type: 'gte', value: 'currentAmount', typeValue: 'variable' }] },
          ref, document, {}
      );
  
      const [{ $or }] = filter.$and;
      assert.deepEqual($or, [
          { amount: { $gte: 100 } },
          { amount: { $gte: '100' } }
      ]);
  });
```

vshvets-bc:
Both points taken, thanks.

**Test added** (8bec45096) — your snippet, plus the `in` list case since that path resolves the variable differently:

```
values resolved from the document
  ✔ widens a variable-resolved range value
  ✔ widens every element of a variable-resolved in list
```

The suite is now **7 tests, 6 of which fail on `develop`** (was 5 and 4).

**Same-field overwrite** — you're right that this PR only composes the subset that routes through `$and`, and that `equal` / `not_equal` / `in` / `not_in` and non-widening ranges still write the key directly, so your `projectId gte 'PRJ-100'` + `not_equal 'PRJ-150'` example still loses a constraint. Filed as #6779 with that repro rather than widening this PR.

dariy-rised (approved):

PR #6747, opened 2026-08-20
fix(frontend): name the wizard's role steps for what they configure

### LINKED ISSUE #6746: Policy Wizard step tree shows the same role twice as "<role> configuration", so the flow reads as a loop
## Symptom

In the Policy Wizard's left-hand step tree, **"OWNER configuration"** appears twice — once nested under **Policy Schemas** and again under **Trust Chain** — with nothing to distinguish them. Walking Next past the second one reads as having gone in a circle, so the user cannot tell where the flow ends.

## Cause

Two different node builders emit the identical label:

`frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts`

```ts
// onSchemaRoleConfigChange - a role's config for one schema, under Policy Schemas
name: `${role} configuration`,
...
// onSelectedTrustChainRoleChange - a role's trust-chain config, under Trust Chain
name: `${trustChainRole} configuration`,
```

## Worth stating: the navigation itself is fine

`SeparateStepperComponent.getNextNode()` walks the tree depth-first and returns nothing at the end, so Next terminates properly — there is **no** wrap-around. The "looping" impression comes purely from two stops sharing a name. This does not need a navigation change.

## Expected

Each step says what it configures, e.g. `OWNER in Project Description` and `OWNER trust chain`.

vshvets-bc:
Fixes #6746

## Fix

Name each role step for what it configures:

- schema role → `OWNER in Project Description` (the schema its parent node carries)
- trust-chain role → `OWNER trust chain`

A schema role node falls back to the previous `<role> configuration` if its parent carries no schema, so nothing breaks if a node is built outside the normal path.

The step walk is untouched — `getNextNode()` already terminates correctly, so there is no navigation bug to fix here (see #6746).

## Tests

New `policy-wizard-dialog.component.spec.ts`, 4 tests. On `develop` with the source change reverted, **3 fail**:

- names a schema role node after the schema it configures
- names a trust-chain role node after the trust chain
- gives the same role different labels in the two sections

The fourth pins the fallback and passes either way by design.

### DIFF
--- a/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
+++ b/frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
@@ -521,7 +521,14 @@ export class PolicyWizardDialogComponent implements OnInit, AfterViewInit {
         const newNode: any = {
             id: `${node.id}.${node.children.length + 1}`,
             icon: node.icon,
-            name: `${role} configuration`,
+            /*
+             * This and the trust-chain role node were both called "<role>
+             * configuration", so the same label appeared under Policy Schemas and again
+             * under Trust Chain. The step walk is a plain depth-first pass with no
+             * wrap-around, but two identically named stops read as one the user keeps
+             * returning to, which is why the flow felt like a loop with no end.
+             */
+            name: node.schema?.name ? `${role} in ${node.schema.name}` : `${role} configuration`,
             parent: node,
             template: this.schemaRoleConfig,
             fields,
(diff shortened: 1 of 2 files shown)


frontend/src/app/modules/policy-engine/dialogs/policy-wizard-dialog/policy-wizard-dialog.component.ts
```diff
@@ -521,7 +521,14 @@ export class PolicyWizardDialogComponent implements OnInit, AfterViewInit {
...
             id: `${node.id}.${node.children.length + 1}`,
             icon: node.icon,
-            name: `${role} configuration`,
+            /*
+             * This and the trust-chain role node were both called "<role>
+             * configuration", so the same label appeared under Policy Schemas and again
+             * under Trust Chain. The step walk is a plain depth-first pass with no
+             * wrap-around, but two identically named stops read as one the user keeps
+             * returning to, which is why the flow felt like a loop with no end.
+             */
```
Pyatakov: Please compact this comment

Pyatakov (changes requested):

Pyatakov (approved):

PR #6751, opened 2026-08-20
fix(guardian-service): prove the Hedera key pair instead of probing for a balance

### LINKED ISSUE #6750: The "Check hedera key" step never verifies the key, and 73b17e259 removes the side effect it relies on
## Summary

The block labelled `// Check hedera key` in registration, profile setup and organization setup does not verify that the supplied private key controls the account. It infers the pairing from a side effect — and `73b17e259` removes that side effect.

## Current mechanism

Three sites run the same idiom (`guardian-service/src/api/helpers/profile-helper.ts:372`, `guardian-service/src/api/helpers/organization-helper.ts:211`, `guardian-service/src/api/profile.service.ts:322`):

```ts
const workers = new Workers();
AccountId.fromString(hederaAccountId);
PrivateKey.fromString(hederaAccountKey);
await workers.addNonRetryableTask({
    type: WorkerTaskType.GET_USER_BALANCE,
    data: { hederaAccountId, hederaAccountKey }
}, { ... });
} catch (error) {
    throw new Error(`Invalid Hedera account or key.`);
}
```

The result is discarded — this is a probe, not a balance read. `PrivateKey.fromString()` proves only that the key **parses**. The pairing is assumed because `GET_USER_BALANCE` currently builds a signing `HederaSDKHelper` client, so a key that does not match the account makes the task throw.

## What 73b17e259 changes

That commit converts `balance()` to a keyless mirror-node REST call, and the worker case becomes:

```ts
case WorkerTaskType.GET_USER_BALANCE: {
    const { hederaAccountId } = task.data;      // hederaAccountKey never destructured
    result.data = await HederaSDKHelper
        .setNetwork(networkOptions)
        .balance(hederaAccountId);
```

The callers still send `hederaAccountKey`; the worker ignores it. After that commit the `catch` can only fire when the **account id** is unknown to the mirror node — never for a wrong key.

## Impact

Not a permanent acceptance of a bad key: the next step is `TopicHelper.create(...)`, which signs, so a wrong key still fails there. What is lost is the diagnosis and the ordering — `STEP_RESOLVE_ACCOUNT` reports success, and the user then gets an SDK error from topic creation instead of "Invalid Hedera account or key."

## The tool is already in the tree

`common/src/hedera-modules/utils.ts`:

```ts
export function checkHederaKey(privateKey: string, publicKey: string): boolean {
    const _privateKey = PrivateKey.fromStringDer(privateKey);
    const _publicKey  = _privateKey.publicKey;
    const _infoKey    = PublicKey.fromString(publicKey);
    return _publicKey.equals(_infoKey);
}
```

Read the account's on-chain public key over the keyless `GET_ACCOUNT_INFO_REST` task, derive the public key from the supplied private key, compare. That proves the pair directly instead of inferring it, and needs no signing, no fee and no mutation.

## Note

The second probe in `profile-helper.ts:649` already sends `data: { hederaAccountId }` with no key, so it is an account-exists check rather than a key check. The two sites already disagree about what they verify.

vshvets-bc:
Fixes #6750

## Fix

New `validateHederaAccountKey` helper: read the account's on-chain public key over the keyless `GET_ACCOUNT_INFO_REST` task, derive the public key from the supplied private key, and compare with the existing `checkHederaKey` from `common`. It throws the same `Invalid Hedera account or key.` the probe threw, so callers are unchanged.

All three probe sites now use it:

- `guardian-service/src/api/helpers/profile-helper.ts:372`
- `guardian-service/src/api/helpers/organization-helper.ts:211`
- `guardian-service/src/api/profile.service.ts:322`

This proves the pair directly rather than inferring it from a signing client being constructed, and needs no signing, no fee and no mutation.

It also unblocks #73b17e259: once these checks no longer depend on `GET_USER_BALANCE` building a client, consolidating onto a single keyless REST balance task is safe.

## Left alone

The second probe in `profile-helper.ts:649` already sends `data: { hederaAccountId }` with no key — it is an account-exists check, not a key check, so it is out of scope here.

## Tests

New `guardian-service/tests/unit/hedera-key-validator.test.mjs`, 6 tests, all passing:

- accepts a key matching the account's on-chain public key
- **rejects a well-formed key that does not control the account** ← the case the probe cannot see once it goes keyless
- sends the account id over `GET_ACCOUNT_INFO_REST` and asserts the key is **never** sent to the worker
- rejects a malformed account id
- rejects when the account cannot be read
- rejects when the account has no key on chain

To be straight about verification: these cover a helper that does not exist on `develop`, so "fails without the fix" is not meaningful for them in the usual way. The behaviour they pin is the wrong-key rejection, which today only works as a side effect of client construction and stops working with 73b17e259. `guardian-service` builds clean, and there are no existing tests over these paths to regress.

### DIFF
--- a/guardian-service/src/api/helpers/hedera-key-validator.ts
+++ b/guardian-service/src/api/helpers/hedera-key-validator.ts
@@ -1,0 +1,46 @@
+import { checkHederaKey, Workers } from '@guardian/common';
+import { AccountId, PrivateKey } from '@hiero-ledger/sdk';
+import { WorkerTaskType } from '@guardian/interfaces';
+
+/**
+ * Prove that a private key controls an account.
+ *
+ * This used to be inferred from a GET_USER_BALANCE probe whose result was discarded:
+ * the pairing was assumed because that task built a signing client and did not throw.
+ * PrivateKey.fromString() beside it only proves the key parses.
+ *
+ * Deriving the public key from the supplied private key and comparing it with the
+ * account's on-chain public key proves the pair directly, and needs no signing, no fee
+ * and no mutation - the account is read over the keyless GET_ACCOUNT_INFO_REST task.
+ *
+ * @throws Error('Invalid Hedera account or key.')
+ */
+export async function validateHederaAccountKey(
+    hederaAccountId: string,
+    hederaAccountKey: string,
+    options?: { userId?: string; interception?: string }
+): Promise<void> {
+    try {
+        AccountId.fromString(hederaAccountId);
+        PrivateKey.fromString(hederaAccountKey);
+
+        const info = await new Workers().addNonRetryableTask({
+            type: WorkerTaskType.GET_ACCOUNT_INFO_REST,
+            data: {
+                hederaAccountId,
+                payload: { userId: options?.userId ?? null }
+            }
+        }, {
+            priority: 20,
+            userId: options?.userId ?? null,
+            interception: options?.interception ?? null
+        });
+
+        // the mirror node returns key as { _type, key }
+        if (!checkHederaKey(hederaAccountKey, info?.key?.key)) {
+            throw new Error('Key does not match the account');
+        }
+    } catch (error) {
+        throw new Error('Invalid Hedera account or key.');
+    }
+}
--- a/guardian-service/src/api/helpers/hedera-key-validator.ts
+++ b/guardian-service/src/api/helpers/hedera-key-validator.ts
@@ -2,29 +2,30 @@ import { checkHederaKey, Workers } from '@guardian/common';
 import { AccountId, PrivateKey } from '@hiero-ledger/sdk';
 import { WorkerTaskType } from '@guardian/interfaces';
 
+// key types the mirror node reports as one comparable public key
+const COMPARABLE_KEY_TYPES = ['ED25519', 'ECDSA_SECP256K1'];
+
 /**
  * Prove that a private key controls an account.
  *
- * This used to be inferred from a GET_USER_BALANCE probe whose result was discarded:
- * the pairing was assumed because that task built a signing client and did not throw.
- * PrivateKey.fromString() beside it only proves the key parses.
+ * The profile-setup "Check hedera key" step only parsed the key and probed for the
+ * account; the first real check came later at TopicHelper.create(). Comparing the
+ * derived public key with the account's on-chain one proves the pair directly, with
+ * no signing, fee or mutation.
  *
- * Deriving the public key from the supplied private key and comparing it with the
- * account's on-chain public key proves the pair directly, and needs no signing, no fee
- * and no mutation - the account is read over the keyless GET_ACCOUNT_INFO_REST task.
- *
- * @throws Error('Invalid Hedera account or key.')
+ * @throws Error('Invalid Hedera account or key.') with the underlying error as `cause`
  */
 export async function validateHederaAccountKey(
     hederaAccountId: string,
     hederaAccountKey: string,
     options?: { userId?: string; interception?: string }
 ): Promise<void> {
+    let info: any;
     try {
         AccountId.fromString(hederaAccountId);
         PrivateKey.fromString(hederaAccountKey);
 
-        const info = await new Workers().addNonRetryableTask({
+        info = await new Workers().addNonRetryableTask({
             type: WorkerTaskType.GET_ACCOUNT_INFO_REST,
             data: {
                 hederaAccountId,
@@ -35,12 +36,22 @@ export async function validateHederaAccountKey(
             userId: options?.userId ?? null,
             interception: options?.interception ?? null
         });
-
-        // the mirror node returns key as { _type, key }
-        if (!checkHederaKey(hederaAccountKey, info?.key?.key)) {
-            throw new Error('Key does not match the account');
-        }
     } catch (error) {
-        throw new Error('Invalid Hedera account or key.');
+        // keep the cause: a NATS timeout and a wrong key otherwise read identically
+        throw new Error('Invalid Hedera account or key.', { cause: error });
+    }
+
+    // a key-list account has no single key to compare against, and PublicKey.fromString
+    // throws on its ProtobufEncoded hex - defer it to the signing operations downstream
+    const keyType = info?.key?._type;
+    if (keyType && !COMPARABLE_KEY_TYPES.includes(keyType)) {
+        return;
+    }
+
+    // the mirror node returns key as { _type, key }
+    if (!checkHederaKey(hederaAccountKey, info?.key?.key)) {
+        throw new Error('Invalid Hedera account or key.', {
+            cause: new Error(`Key does not match account '${hederaAccountId}'`),
+        });
     }
 }
--- a/guardian-service/src/api/helpers/organization-helper.ts
+++ b/guardian-service/src/api/helpers/organization-helper.ts
@@ -29,6 +29,7 @@ import {
 } from '@guardian/common';
 import { AccountId, PrivateKey } from '@hiero-ledger/sdk';
 import { getGlobalTopic } from './profile-helper.js';
+import { validateHederaAccountKey } from './hedera-key-validator.js';
 
 /**
  * Lightweight shape used by guardian-service to round-trip Organization records over NATS.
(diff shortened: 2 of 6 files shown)


guardian-service/src/api/helpers/hedera-key-validator.ts
```diff
@@ -0,0 +1,46 @@
...
+    hederaAccountId: string,
+    hederaAccountKey: string,
+    options?: { userId?: string; interception?: string }
+): Promise<void> {
+    try {
+        AccountId.fromString(hederaAccountId);
+        PrivateKey.fromString(hederaAccountKey);
+
+        const info = await new Workers().addNonRetryableTask({
+            type: WorkerTaskType.GET_ACCOUNT_INFO_REST,
```
Pyatakov, line 29: The `HederaSDKHelper.accountInfo()` static method (which `GET_ACCOUNT_INFO_REST` calls) does a single `axios.get()` with no retry. The mirror node returns 404 for accounts it hasn't ingested yet, and `ONBOARD_USER_ASYNC` creates an account then immediately validates it.

**Suggested:**
Add the same `MIRROR_NODE_INDEXING_ATTEMPTS` loop to `accountInfo()` that `balance()` has (worker-service/src/api/helpers/hedera-sdk-helper.ts:2317–2334 on develop – introduced at #6736 with some further improvements made in #6737). Retry up to 10 times with a 1-second delay between attempts if the result is empty.

Alternatively, if you want to avoid modifying the helper, use the `balance()` call as the account-exists gate and add `checkHederaKey` comparison on top of it (cheaper than an extra REST call, and you already know the retry logic works).

guardian-service/src/api/helpers/hedera-key-validator.ts
```diff
@@ -0,0 +1,46 @@
...
+            userId: options?.userId ?? null,
+            interception: options?.interception ?? null
+        });
+
+        // the mirror node returns key as { _type, key }
+        if (!checkHederaKey(hederaAccountKey, info?.key?.key)) {
+            throw new Error('Key does not match the account');
+        }
+    } catch (error) {
+        throw new Error('Invalid Hedera account or key.');
```
Pyatakov: The catch wraps the entire account read, so any NATS timeout, mirror 5xx, or `@timeout` rejection gets rewritten to "Invalid Hedera account or key." with the original error discarded.

**Suggested:**
Chain or log the cause
```ts
} catch (error) {
    await logger.error(error, ['GUARDIAN_SERVICE']);
    throw new Error('Invalid Hedera account or key.', { cause: error });
}
```

Or be more surgical and only rewrite the message for the three causes that are the user's fault (malformed ID, nonexistent account, key mismatch), and let infrastructure failures propagate:

```ts
catch (error) {
    if (error.message?.includes('bad account') || 
        error.message?.includes('Invalid account') ||
        error.message?.includes('Key does not match')) {
        throw new Error('Invalid Hedera account or key.');
    }
    throw error;  // Mirror down, NATS timeout, etc. — propagate as-is
}
```

guardian-service/src/api/helpers/hedera-key-validator.ts
```diff
@@ -0,0 +1,46 @@
...
+ * This used to be inferred from a GET_USER_BALANCE probe whose result was discarded:
+ * the pairing was assumed because that task built a signing client and did not throw.
+ * PrivateKey.fromString() beside it only proves the key parses.
+ *
+ * Deriving the public key from the supplied private key and comparing it with the
+ * account's on-chain public key proves the pair directly, and needs no signing, no fee
+ * and no mutation - the account is read over the keyless GET_ACCOUNT_INFO_REST task.
+ *
+ * @throws Error('Invalid Hedera account or key.')
+ */
```
Pyatakov, line 17: **Rationale wrong**
Reword the comment to something like:
```ts
/**
 * Prove that a private key controls an account.
 *
 * The `// Check hedera key` step in profile setup used to parse the key and probe
 * for the account's existence, but never verified the key controlled it. The first
 * real check came later at TopicHelper.create(), which signs – making the "check"
 * step misleading and the error messages worse.
 *
 * This reads the account's on-chain public key over the keyless GET_ACCOUNT_INFO_REST
 * task, derives the public key from the supplied private key, and compares them.
 * It proves the pair directly, needs no signing, no fee, and no mutation.
 *
 * @throws Error('Invalid Hedera account or key.')
 */
```

guardian-service/tests/unit/hedera-key-validator.test.mjs
Pyatakov, line 1: **Tests don't exercise real behavior**
Add a test with real keys after the stubbed ones:
```ts
it('with real SDK: rejects a key from a different account', async () => {
    const { PrivateKey, PublicKey } = await import('@hiero-ledger/sdk');
    const { checkHederaKey } = await import('@guardian/common');
    
    const keyA = PrivateKey.generateED25519();
    const keyB = PrivateKey.generateED25519();
    
    // checkHederaKey should accept the matching public key
    assert.isTrue(checkHederaKey(keyA.toStringDer(), keyA.publicKey.toStringDer()));
    
    // and reject a key from a different account
    assert.isFalse(checkHederaKey(keyA.toStringDer(), keyB.publicKey.toStringDer()));
});
```
This will catch the DER-encoding and key-comparison bugs that the stubbed version misses.

guardian-service/src/api/helpers/organization-helper.ts
```diff
@@ -29,6 +29,7 @@ import {
 } from '@guardian/common';
 import { AccountId, PrivateKey } from '@hiero-ledger/sdk';
```
Pyatakov: `AccountId, PrivateKey` are dead imports now (no uses after the removed block)

guardian-service/src/api/profile.service.ts
```diff
@@ -25,6 +25,7 @@ import { Controller, Module } from '@nestjs/common';
 import { ClientsModule, Transport } from '@nestjs/microservices';
 import { AccountId, PrivateKey } from '@hiero-ledger/sdk';
```
Pyatakov: `AccountId` is dead import now – no use after the removed block
(`PrivateKey` still in use at lines 267 and 627).

guardian-service/src/api/helpers/hedera-key-validator.ts
```diff
@@ -0,0 +1,46 @@
...
+                payload: { userId: options?.userId ?? null }
+            }
+        }, {
+            priority: 20,
+            userId: options?.userId ?? null,
+            interception: options?.interception ?? null
+        });
+
+        // the mirror node returns key as { _type, key }
+        if (!checkHederaKey(hederaAccountKey, info?.key?.key)) {
```
Pyatakov: **Key-list accounts rejected**
The mirror node returns `key: { _type: 'ProtobufEncoded', key: '<hex>' }` for threshold/key-list accounts. `PublicKey.fromString()` throws on protobuf hex, and the helper rejects the user even though a 1-of-N list would sign fine later.

**Suggested:**
In `validateHederaAccountKey`, check the key type before calling `checkHederaKey`:
```ts
if (info?.key?._type && !['ED25519', 'ECDSA_SECP256K1'].includes(info.key._type)) {
    // Account has a key list or other composite key; we can't verify it against a single private key
    // but it's still readable and will be checked when TopicHelper.create signs
    return;
}
if (!checkHederaKey(hederaAccountKey, info?.key?.key)) {
    throw new Error('Key does not match the account');
}
```

Pyatakov (changes requested):

vshvets-bc:
Thanks — all seven were correct, and two of them changed behaviour for real users. Pushed in 6ceb9d4.

**Key-list accounts were rejected.** You're right that `PublicKey.fromString()` throws on `ProtobufEncoded` hex, so the helper refused a user whose 1-of-N list would have signed fine later. Only `ED25519` and `ECDSA_SECP256K1` are compared now; anything else has no single key to compare against, so the check doesn't apply and those accounts defer to the signing operations downstream — which is where they were always going to be proven.

**`accountInfo()` had no retry.** Added the `MIRROR_NODE_INDEXING_ATTEMPTS` loop as you suggested, on the same schedule `balance()` uses. One refinement: only a **404** is waited on. Auth failures, 5xx and network errors won't resolve by retrying, and retrying them for 10 seconds would just turn a fast clear failure into a slow vague one.

**The catch no longer discards the cause.** Message unchanged for callers, original attached as `cause`. I also moved the key comparison *out* of the `try`, so a genuine mismatch is no longer caught by the handler meant for the account read — otherwise the two failure modes stay conflated even with the cause attached.

**Reworded the doc comment** along the lines you gave — it described the old behaviour's rationale rather than what the step now does.

**Dead imports** dropped: both in `organization-helper`, `AccountId` only in `profile.service` (`PrivateKey` is still used at `:267` and `:627`).

**Tests** — this was the most useful point. The existing ones stub `checkHederaKey` as string concatenation, so they'd pass against a validator that never validates. Added four cases on the real SDK and the real `checkHederaKey` with generated keypairs: accepts the controlling key, rejects a key from a different account, accepts ECDSA, and doesn't reject a key-list account — plus one asserting the cause survives. 11 passing, and I checked the key-list case genuinely fails without the new guard rather than passing for some unrelated reason.

Pyatakov:
@vshvets-bc please delete any unnecessary comments (where code is self explanatory) and shorten the text where it needs to remain but could be less verbose

vshvets-bc:
Done — comments cut from 53 lines to 17 across the four files (diff is now +278/-61, was +314/-61).

- `COMPARABLE_KEY_TYPES`: 4-line block → one line.
- Function doc: 14 lines → 8, keeping the reason the step changed and the `@throws`/`cause` contract.
- The catch: 5 lines → one.
- The key-list guard: 7 lines → 2.
- `accountInfo()` retry rationale: 6 lines → 2, and the 404-only note to one.
- Test file: the two block headers and the key-list note down to one or two lines each.

11 tests still passing, and both services build.

Pyatakov (approved):

PR #6754, opened 2026-08-24
fix: strip template field ids and add tests

dariy-rised:
**Description**:
- Fixed `Error: strict mode: unknown keyword: "templateFieldId" `

### DIFF
--- a/common/src/hedera-modules/vcjs/vcjs.ts
+++ b/common/src/hedera-modules/vcjs/vcjs.ts
@@ -545,10 +545,20 @@ export class VCJS {
             }
         };
 
-        stripProperties(schema?.properties);
+        const stripNode = (node: any) => {
+            stripProperties(node?.properties);
+            if (Array.isArray(node?.allOf)) {
+                for (const entry of node.allOf) {
+                    stripNode(entry?.then);
+                    stripNode(entry?.else);
+                }
+            }
+        };
+
+        stripNode(schema);
         if (schema?.$defs) {
             for (const nestedSchema of Object.values<any>(schema.$defs)) {
-                stripProperties(nestedSchema?.properties);
+                stripNode(nestedSchema);
             }
         }
     }
--- a/common/src/hedera-modules/vcjs/vcjs.ts
+++ b/common/src/hedera-modules/vcjs/vcjs.ts
@@ -381,6 +381,7 @@ export class VCJS {
      */
     private prepareSchema(schema: any) {
         this.stripIfOnly(schema);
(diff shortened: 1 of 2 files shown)


common/tests/unit-tests/hedera-modules/vcjs/vcjs-coverage.test.mjs
Pyatakov, line 1: Add a test case with a conditional field (e.g., `allOf: [{if: {...}, then: {properties: {condField: {type:'string', templateFieldId:'id'}}}}]`) to catch the regression.

common/src/hedera-modules/vcjs/vcjs.ts
```diff
@@ -518,6 +519,40 @@ export class VCJS {
...
+                if (property.properties) {
+                    stripProperties(property.properties);
+                }
+                if (property.items?.properties) {
+                    stripProperties(property.items.properties);
+                }
+            }
+        };
+
+        stripProperties(schema?.properties);
```
Pyatakov: **Condition branches (`allOf[].then/else`) are never stripped.**

The fix requires expanding `stripTemplateFieldIds` to walk into condition branches. Here's exactly what needs to change:

**Current code** (line 527–554) only handles:
- `schema.properties` 
- `schema.$defs[*].properties`

**but missing** `schema.allOf[].then.properties` and `schema.allOf[].else.properties` (and the same for nested schemas inside `$defs`)

To fix that, replace the current `stripTemplateFieldIds` method with one that:
1. **Creates a node-walker** that recursively processes any schema node (the root, `then`, `else`, or nested `$defs` entries)
2. **On each node**, calls `stripProperties` on `.properties`, then walks any `.allOf` array to process each entry's `.then` and `.else`
3. **Applies this walker** to both the root schema and each nested schema in `$defs`

**Pseudo-structure:**
```ts
private stripTemplateFieldIds(schema: any) {
    const stripProperties = (properties: any) => {
        // ... existing code ...
    };

    const stripNode = (node: any) => {
        stripProperties(node?.properties);
        if (Array.isArray(node?.allOf)) {
            for (const entry of node.allOf) {
                stripNode(entry?.then);
                stripNode(entry?.else);
            }
        }
    };

    stripNode(schema);
    if (schema?.$defs) {
        for (const nestedSchema of Object.values<any>(schema.$defs)) {
            stripNode(nestedSchema);
        }
    }
}
```

Pyatakov (changes requested):

Pyatakov (approved):

Issue #5308, opened 2025-09-17
Enhance Guardian UI Form to Support Rich Text


prernaadev01:
When authoring Guardian schemas and forms I want to emphasize important information in a structured and clear way using headings, lists, links, and text styles to avoid misinterpretation and support better readability.

## Requirements

Authors can format text so that rendered output will be styles. Integrates with third party systems such as project hub and the hedera network.

## Acceptance criteria

* support for headings, bold, italics
* bulleted and numbered lists
* hyperlinks
* formatting preserved when displayed in final output
* Existing plain text entries remain unaffected and display as expected
    

## Additional Notes 

* data stored as markdown
* additional field type for rich text support added to schemas
* authors of existing policies will need to be update their policies opt in to receive rich text fields
* we will not update any policies to automatcially change existing fields to rich text
* decided to leave existing policies alone and in VCS v5.0 fields can be updated via templates to include rich text

aouatefm:
Hi! 👋 I’d like to contribute to this issue.
While checking the linked image, I noticed that it’s returning a Non-Image content-type error — could you please re-upload it or share a valid image link?
I’m happy to start exploring the implementation for the Rich Text Editor enhancement once that’s fixed.

prernaadev01:
@aouatefm Thank you for your interest in contributing to this ticket. We are still working on the refinement of this ticket. Please give us sometime and we will reach out to you when it is ready.

prernaadev01:
@Dairus01 thank you for checking. Just wanted to let you know this task is still in our backlog. Great to see your interest.

PR #6777, opened 2026-08-26
feat: 5308 add Rich Text field type for schema forms

ihar-tsykala:
**Description**:

Every text field in a Guardian schema was plain text. An author writing instructions into a form had no way to mark a heading, emphasise a warning, or link to an external page: everything arrived as one flat paragraph, which readers misread.

Add a **Rich Text** field type. In the schema it is an ordinary string field carrying `customType: 'richText'`, so the value stays a plain JSON string and nothing downstream needs a new data type. It is stored as **Markdown**, which travels safely to Hedera and stays readable on its own. A policy form shows a small editor instead of a textarea — bold, italic, both list kinds, H1–H3 and links — and the document viewer, the documents grid and the Indexer render that Markdown as formatted output, with links opening in a new tab. Pasted content is reduced to the same set, so a copy from a word processor cannot bring fonts, tables or images into a document. The type is registered in the field-type picker and the Excel dictionary, and preset values use the same editor.

Underline is deliberately absent: Markdown cannot store it, and raw HTML is not mixed into the value. Nothing changes for a field without that custom type, and stored values are never rewritten.

Fixes #5308

**Checklist**


frontend/src/app/modules/schema-engine/rich-text-editor/markdown.ts
```diff
@@ -0,0 +1,163 @@
+import { isSafeHref } from './rich-text-sanitizer';
+
+export function escapeHtml(value: string): string {
+    return value
+        .replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;');
```
Pyatakov: `escapeHtml` escapes `&`, `<` and `>` but not `"`, and the link regex `([^)\s]+)` happily captures a quote. A stored value of `[click](https://x"onmouseover="window.__pwned=1)` renders as `<a href="https://x"onmouseover="window.__pwned=1" target="_blank" ...>` – the browser parses `onmouseover` as a real event handler. `rich-text-editor.component.ts:308` assigns this HTML directly to `el.innerHTML`, so Angular's sanitizer never sees it and the handler executes.

```suggestion
        .replace(/>/g, '&gt;')
        .replace(/"/g, '&quot;')
        .replace(/'/g, '&#39;');
```

The same fix is needed in `indexer-frontend/src/app/components/schema-form-view/markdown-view.ts`.

indexer-frontend/src/app/components/schema-form-view/markdown-view.ts
```diff
@@ -0,0 +1,65 @@
+import { isSafeHref } from './rich-text-view';
+
+export function escapeHtml(value: string): string {
+    return value
+        .replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;');
```
Pyatakov: Same defect as `frontend/.../markdown.ts`: `"` is not escaped, so a URL like `https://x"onmouseover="…` breaks out of the `href` attribute. The indexer only renders through Angular `[innerHTML]`, which strips the handler, so this corrupts the link rather than executing – but fix it here too so the two copies stay in step.

```suggestion
        .replace(/>/g, '&gt;')
        .replace(/"/g, '&quot;')
        .replace(/'/g, '&#39;');
```

frontend/src/app/modules/schema-engine/rich-text-editor/rich-text-editor.component.ts
```diff
@@ -0,0 +1,479 @@
...
+    private _currentBlockFormat(): string {
+        try {
+            return (document.queryCommandValue('formatBlock') || '').toLowerCase();
+        } catch {
+            return '';
+        }
+    }
+}
+
+function escapeText(text: string): string {
```
Pyatakov, line 474: `escapeText` escapes markup but leaves raw `\n` in place, and `insertHTML` collapses those as HTML whitespace – pasting three lines of plain text produces one line. Split on newlines and emit `<br>` (or wrap the lines in `<p>`):

```suggestion
function escapeText(text: string): string {
    const inert = document.implementation.createHTMLDocument('');
    const holder = inert.createElement('div');
    return text
        .replace(/\r\n/g, '\n')
        .split('\n')
        .map((line) => {
            holder.textContent = line;
            return holder.innerHTML;
        })
        .join('<br>');
}
```

frontend/src/app/modules/policy-engine/policy-viewer/blocks/documents-source-block/documents-source-block.component.html
```diff
@@ -73,6 +73,15 @@
                             class="text"
                           >{{ getText(element, option) }}</span>
                         }
+                        @if (option.type == 'richText') {
+                          @let richTextCellText = getRichTextCellText(element, option);
```
Pyatakov, line 77: `getRichTextCellText` runs a full markdown render plus `document.implementation.createHTMLDocument()` (in `richTextToText`) for every rich-text cell on every change-detection pass, and this component is not OnPush. With a few hundred rows that is a few hundred document allocations per tick – and the hover handlers added here make ticks frequent.

Compute the plain-text cell value once when the rows are built (alongside the other per-row fields in the data mapping) and read the cached string from the template instead.

interfaces/src/helpers/schema-json.ts
```diff
@@ -184,6 +184,10 @@ export class SchemaToJson {
             return 'HederaAccount';
         }
 
+        if (field.customType === 'richText') {
```
Pyatakov, line 188: The loop below already returns `'Rich Text'` for these fields – `FieldTypesDictionary.equal` compares `customType`, and the new dictionary entry has no `pattern`/`format` that could drift (unlike `hederaAccount`, which is why that one needs the early return). This branch can go.

Pyatakov (commented):
Please, address the comments below (more details added inline).

* `frontend/src/app/modules/schema-engine/rich-text-editor/markdown.ts:7` – escapeHtml leaves `"` unescaped, so a markdown link URL can inject attributes into the generated `<a>`
* `indexer-frontend/src/app/components/schema-form-view/markdown-view.ts:7` – Same unescaped-quote attribute injection in the indexer copy of escapeHtml
* `frontend/src/app/modules/schema-engine/rich-text-editor/rich-text-editor.component.ts:474` – Plain-text paste collapses every line break into a single line
* `frontend/src/app/modules/policy-engine/policy-viewer/blocks/documents-source-block/documents-source-block.component.html:77` – getRichTextCellText parses a fresh HTMLDocument per rich-text cell on every change detection cycle
* `interfaces/src/helpers/schema-json.ts:187` – Redundant richText special case – the loop below already resolves it

Pyatakov (changes requested):

frontend/src/app/modules/schema-engine/rich-text-editor/markdown.ts
ihar-tsykala: It was fixed. Could you check it. Thanks!

indexer-frontend/src/app/components/schema-form-view/markdown-view.ts
ihar-tsykala: It was fixed. Could you check it. Thanks!

frontend/src/app/modules/schema-engine/rich-text-editor/rich-text-editor.component.ts
ihar-tsykala: It was fixed. Could you check it. Thanks!

interfaces/src/helpers/schema-json.ts
ihar-tsykala: This branch can't be removed. The loop below walks FieldTypesDictionary.FieldTypes, while the Rich Text entry lives in CustomFieldTypes — together with hederaAccount, Prefix and Postfix.

The split is deliberate: FieldTypes holds the types a field can be recognised by from the schema alone — by type, format and pattern. CustomFieldTypes holds the ones that are indistinguishable from their base type in the schema: Rich Text is a plain string, and only the customType marker tells it apart. equal does compare customType, but every entry in FieldTypes has it empty, so there is no match.

Without the early return the field falls through to return 'String', and the schema is exported with type String instead of Rich Text. That is also why every entry from CustomFieldTypes has its own if here, hederaAccount included.

frontend/src/app/modules/policy-engine/policy-viewer/blocks/documents-source-block/documents-source-block.component.html
ihar-tsykala: It was fixed. Could you check it. Thanks!

Pyatakov (approved):

Issue #6783, opened 2026-08-28
verifySchema: conditional-required strip skipped for wrapper-nested subject schemas; only the first credential subject is checked


nazlo-climission:
verifySchema has two related defects found during an internal regression audit:

1. prepareSchema's conditional-required strip only walks the object it's handed directly. When verifySchema compiles the VC wrapper schema, the wrapper's root never carries `allOf` - the subject schema's `allOf` lives nested under `$defs`. A field that's base-required in the subject sub-schema but conditionally not required stays unconditionally required, so a document that's valid at request time can be rejected later on an external-verify path (external-data-block, upload-vc-document-block, external-topic-block, http-request-block).

2. verifySchema only restores ref-field `type`/`@context` (via `ContextHelper.setContext`) on `credentialSubject[0]`. A multi-subject VC is only checked consistently for its first subject.

**Fix**: run the conditional-required strip against every `$defs` entry that carries its own `allOf`, in addition to the root, keyed by scope so two containers stripping the same `$ref`'d entry along different paths don't collide. Loop `ContextHelper.setContext` over every subject. Reject an empty or all-null `credentialSubject` array instead of validating vacuously.

Issue #1743, opened 2023-02-02
CLI integration


ieumuzair:
Guardian-Cli is a useful command-line application developed for simplifying management of guardian project from cloning the repository to checkout to a specific release version, start/stop the project by docker and PM2 as well as cleaning the project and more features. It is supposed guardian-cli to be pushed into npm registry in order that users can install it as global package and use cli commands entire their machine or servers.

dubgeis:
This may become Hex Trust or another custodian @ieumuzair

PR #6785, opened 2026-08-28
fix: run the VC wrapper's conditional-required strip

nazlo-climission:
**Description**:

This PR fixes two `verifySchema` defects on the VC-wrapper validation path.

* Run the conditional-required strip against every `$defs` entry that carries its own `allOf`, not just the object handed to `prepareSchema` directly, so a condition nested in the subject definition (not the root) is honored
* Restore ref-field `type`/`@context` for every credential subject via `ContextHelper.setContext`, not just the first
* Reject an empty or all-null `credentialSubject` array instead of validating vacuously

**Related issue(s)**:

Fixes #6783

**Notes for reviewer**:

`tsc --noEmit` clean on `interfaces` and `common`. New/extended tests in `common/tests/unit-tests/hedera-modules/vcjs/vcjs-coverage.test.mjs`, logic-verified against the compiled dist via a standalone script (this checkout's `node_modules` is missing chai's transitive deps repo-wide, so `npm test` could not run as-is). 8/8 checks passed, including a regression check that the pre-existing "shared sub-schema sibling isolation" behavior is unchanged.

**Checklist**

### DIFF
--- a/common/src/hedera-modules/vcjs/vcjs.ts
+++ b/common/src/hedera-modules/vcjs/vcjs.ts
@@ -296,8 +296,12 @@ export class VCJS {
 
         const vcObject = JSON.parse(JSON.stringify(vc));
 
-        const subjects = vcObject.credentialSubject;
-        const subject = Array.isArray(subjects) ? subjects[0] : subjects;
+        const rawSubjects = vcObject.credentialSubject;
+        const subjects: any[] = Array.isArray(rawSubjects) ? rawSubjects : [rawSubjects];
+        if (!subjects.length || !subjects[0]) {
+            throw new Error('"credentialSubject" property is required.');
+        }
+        const subject = subjects[0];
 
         if (!this.schemaLoader) {
             throw new Error('Schema Loader not found');
@@ -318,7 +322,12 @@ export class VCJS {
 
         const schemaObject = Schema.fromVc(schema);
 
-        ContextHelper.setContext(subject, schemaObject);
+        // Every subject needs its ref-field type/@context restored before validation,
+        // not just the first - otherwise a multi-subject VC only gets subject[0]
+        // checked consistently with the schema.
+        for (const item of subjects) {
+            ContextHelper.setContext(item, schemaObject);
+        }
 
         const validate = await ajv.compileAsync(schema);
         const valid = validate(vcObject);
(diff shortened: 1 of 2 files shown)


ihar-tsykala (changes requested):
Hi @nazlo-climission,

The wrapper case is fixed, and the new tests do fail without the fix.

One problem blocks the merge. The new loop processes $defs in this order: Subject, then Detail, then Extra. While processing Subject, it creates Detail__copy and changes Subject to point to that copy. At this moment Detail__copy still points to the original Extra, where note is required.

The loop then processes the original Detail, creates Extra__copy without note in required, and changes only the original Detail to point to it. The already-created Detail__copy is not in the saved Object.keys($defs) list, so it is never processed. The validation path is therefore Subject -> Detail__copy -> Extra, not Subject -> Detail -> Extra__copy. Given an inactive condition in Detail, a document that correctly omits note is still rejected as missing a required field: https://github.com/hashgraph/guardian/issues/6783 again, one level deeper.

Fix this with a depth-first preparation: before cloning Detail, prepare Detail and its conditional references first. Keep a Set of prepared/in-progress definitions so each one is processed once and $ref cycles terminate.

ihar-tsykala (approved):

nazlo-climission:
> Hi @nazlo-climission,
> 
> The wrapper case is fixed, and the new tests do fail without the fix.
> 
> One problem blocks the merge. The new loop processes $defs in this order: Subject, then Detail, then Extra. While processing Subject, it creates Detail__copy and changes Subject to point to that copy. At this moment Detail__copy still points to the original Extra, where note is required.
> 
> The loop then processes the original Detail, creates Extra__copy without note in required, and changes only the original Detail to point to it. The already-created Detail__copy is not in the saved Object.keys($defs) list, so it is never processed. The validation path is therefore Subject -> Detail__copy -> Extra, not Subject -> Detail -> Extra__copy. Given an inactive condition in Detail, a document that correctly omits note is still rejected as missing a required field: #6783 again, one level deeper.
> 
> Fix this with a depth-first preparation: before cloning Detail, prepare Detail and its conditional references first. Keep a Set of prepared/in-progress definitions so each one is processed once and $ref cycles terminate.

fixed, pls check it

swidnikk:
@nazlo-climission please be sure the DCO check passes — thank you!

common/tests/unit-tests/hedera-modules/vcjs/vcjs-coverage.test.mjs
```diff
@@ -447,6 +335,168 @@ describe('VCJS coverage (offline paths)', function () {
             });
             assert.property(result, 'ok');
         });
+
+        // #1743: verifySchema used to restore ref-field type/@context (via
```
Pyatakov: New tests' comments cite #1743; the PR says it fixes #6783. Same on the `describe` title at line 415.
Drop it (preferable) or change to the OS one.

Pyatakov (changes requested):

common/src/hedera-modules/vcjs/vcjs.ts
```diff
@@ -399,26 +382,99 @@ export class VCJS {
...
+        // in defsObj[ref] when it clones a referenced def into a container (e.g.
+        // Subject cloning Detail into Detail__copy). If Detail's own conditional
+        // strip hasn't run yet, the clone carries a stale $ref to Detail's
+        // original (unstripped) dependency, and that dependency's fix never
+        // reaches the clone - it isn't among the original defsKeys, so the flat
+        // loop this replaced would never visit it. Preparing depth-first (a def's
+        // own $ref'd dependencies before the def itself) guarantees every def is
+        // fully stripped before anything clones it. The prepared/inProgress sets
+        // make sure each def runs once and $ref cycles terminate instead of
+        // recursing forever.
```
Pyatakov: Please keep comment short, clear, and focused on explaining what the code does.

Pyatakov (approved):

PR #6787, opened 2026-08-28
fix(guardian-service): stop recursive sub-tool import looping on a cycle

vshvets-bc:
## Problem

`importToolByMessage` recurses into the sub-tools listed inside the archive, and the only dedup check is the `DatabaseServer.getTool({ messageId })` lookup — for a row that `createTool` writes **after** the recursion has already returned. A tool on the current import path is therefore never visible to its own check.

So a self- or mutually-referencing archive recurses without end. Every call is awaited, so the stack never overflows and nothing is raised: the import simply never returns, writing **one orphan GridFS blob, one notifier step, one Hedera `getMessage` and one IPFS fetch per turn**.

Running the new suite against current `develop` exhausts a 2 GB heap:

```
[166386:0x28def1c0] 29817 ms: Scavenge (reduce) 2045.4 (2081.3) -> 2045.0 (2081.6) MB ... allocation failure
```

The sub-tool list is archive-controlled: `parseZipFile` returns every `tools/*.json` entry verbatim, `messageId` included, and nothing checks an entry against the tool being imported. **It does not take a malicious archive** — a mis-published tool that references itself is enough.

## Fix

**Cycle guard.** Thread a `visited` set through `importSubTools` / `importToolByMessage` holding the message ids on the current import *path*, and refuse an id already on it. The id is removed on the way out, so a legitimate diamond (A imports B and C, both importing D) still resolves D twice rather than being mistaken for a cycle. `MAX_TOOL_IMPORT_DEPTH` (32) is a backstop for a pathological but acyclic chain, and bounds the damage if a future edit breaks the cycle check.

**Orphan blobs.** Save the tool file to GridFS *after* the sub-tool import rather than before. `createTool` is only reached further down, so a blob saved ahead of the recursion is orphaned by anything that throws in between, with no reference and no cleanup path. That was one orphan per turn of the loop, and it is still one per failed import of any kind. Steps are looked up by name, so the reordering does not disturb the progress tree.

## Tests

`guardian-service/tests/tool-import-cycle.test.mjs` — 9 cases: self-reference, mutual (A→B→A) and longer (A→B→C→A) cycles, the depth ceiling, orphan-blob accounting, and the cases that must **not** be treated as cycles (diamond, repeated sibling), plus the plain and normally-nested tools as controls.

All 9 pass with the fix. Against current `develop` the cycle cases hang until the process runs out of memory, which is what the heap trace above is.

### DIFF
--- a/guardian-service/src/helpers/import-helpers/tool/tool-import-helper.ts
+++ b/guardian-service/src/helpers/import-helpers/tool/tool-import-helper.ts
@@ -5,11 +5,15 @@ import { SchemaImportExportHelper } from '../schema/schema-import-helper.js';
 import { ImportToolMap, ImportToolResult, ImportToolResults } from './tool-import.interface.js';
 import { resolveToolOverrides } from './tool-override-resolver.js';
 
+// backstop only - `visited` is what catches cycles
+const MAX_TOOL_IMPORT_DEPTH = 32;
+
 /**
  * Import tools by messages
  * @param owner
  * @param messages
  * @param notifier
+ * @param visited message ids on the current import path
  */
 export async function importSubTools(
     hederaAccount: IRootConfig,
@@ -19,7 +23,8 @@ export async function importSubTools(
     }[],
     user: IOwner,
     notifier: INotificationStep,
-    userId: string | null
+    userId: string | null,
+    visited: Set<string> = new Set<string>()
 ): Promise<ImportToolResults> {
     notifier.start();
 
@@ -40,7 +45,8 @@ export async function importSubTools(
                 message.messageId,
                 user,
                 step,
-                userId
+                userId,
+                visited
             );
             if (importResult.tool) {
                 tools.push(importResult.tool);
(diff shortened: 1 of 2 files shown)


ihar-tsykala (approved):

Pyatakov:
@vshvets-bc please delete any unnecessary comments (where code is self explanatory) and shorten the text where it needs to remain but could be less verbose

vshvets-bc:
Done — comments cut from 47 lines to 11 across both files (diff is now +230/-9, was +266/-9).

- `MAX_TOOL_IMPORT_DEPTH`: 6-line block → one line.
- The cycle-check rationale: 19 lines → 5, keeping only what the code cannot say — that `getTool` runs before `createTool` writes the row, and that `visited` is the path rather than everything seen.
- The save-file reorder: 9 lines → 2.
- The `finally` comment: deleted, `visited.delete(messageId)` says it.
- Test header: 11 lines → 3. The per-case one-liners are kept; they say which defect each case pins.

9 tests still passing.

PR #6788, opened 2026-08-28
fix: report an unresolvable artifact instead of throwing a TypeError

vshvets-bc:
## Problem

`PolicyImportExport.parseZipFile` matches each `artifacts/` entry to a record in `artifacts/metadata.json` with `find`, which returns `undefined` when nothing matches — and the next line dereferences it:

```ts
const artifactMetaData = metaDataBody.find(item => item.uuid === uuid);
return {
    name: artifactMetaData.name,   // <- TypeError when nothing matched
```

The user sees:

```
TypeError: Cannot read properties of undefined (reading 'name')
```

raised from inside `@guardian/common`, with no indication that their archive was the cause and no mention of which entry. Because the throw happens inside a `Promise.all` over every artifact, it also takes any well-formed sibling with it — one bad entry costs the whole import.

Three ways in, none requiring malice:

- `metadata.json` absent, so the `'[]'` fallback makes every lookup miss
- a hand-edited or truncated archive that omits a record
- a **nested** path such as `artifacts/sub/file`, which passes the `/^artifacts\/.+/` filter but whose `split('/')[1]` yields `sub` and matches no uuid

## Fix

Unresolvable entries are skipped and reported on the returned components as `artifactErrors`, naming the path:

```
Artifact metadata missing for "artifacts/orphan-uuid".
Artifact "artifacts/sub/file" is nested; artifacts must sit directly under artifacts/.
```

One malformed entry should not cost the user the rest of a valid archive.

**`artifactErrors` is deleted in `cleanBeforeHash`.** `getPolicyHash` stringifies the whole components object, so leaving it in would give a policy a different hash purely because its archive had a malformed artifact entry — and those are exactly the archives that get corrected and re-imported. The hash has to describe the policy's content, not how cleanly the zip parsed. There is a test pinning that invariance.

## Tests

`common/tests/unit-tests/import-export/artifact-metadata-mismatch.test.mjs` — 9 cases covering the missing record, absent `metadata.json`, the nested path, a well-formed sibling surviving a malformed one, the hash invariance, and controls for a clean archive and for preview mode (`includeArtifactsData: false`).

Against current `develop`, **6 of the 9 fail with the exact `TypeError` above**.


common/src/import-export/policy.ts
```diff
@@ -459,20 +459,65 @@ export class PolicyImportExport {
...
+             * yields 'sub' and matches no uuid.
+             *
+             * Skipped and reported rather than thrown: one malformed entry should
+             * not cost the user the rest of a valid archive, and the message names
+             * the path so support can act on it.
+             */
+            const artifactEntries = fileEntries.filter(
+                file => /^artifacts\/.+/.test(file[0]) && file[0] !== 'artifacts/metadata.json'
+            );
+            const resolved = await Promise.all(artifactEntries.map(async file => {
```
dariy-rised, line 523: Guards `find` returning `undefined`, but not `metaDataBody` being non-array to begin with:

```ts
const artifactMetaData = isNested
    ? undefined
    : metaDataBody.find(item => item.uuid === uuid);
```

If `artifacts/metadata.json` is valid JSON but not an array (e.g. `{}`), `.find`/`.map` throws `TypeError` uncaught — crashing the *entire* import again, in both `includeArtifactsData` branches (reproduced against the built PR branch). Guardian's own exporter always writes an array, so this only bites a hand-edited or corrupted archive, not normal usage — but that's the same class of input this PR is already hardening against, and here the blast radius is bigger than the bug being fixed: one malformed record now degrades gracefully, but a malformed `metaDataBody` shape still takes down every artifact in the archive.

Since this is the exact code block already being touched, suggest closing it here too: `Array.isArray(...) ? ... : []` when parsing `metaDataBody`. Not a blocker if you'd rather scope this PR to the reported issue and track it separately.

common/src/import-export/policy.ts
```diff
@@ -459,20 +459,65 @@ export class PolicyImportExport {
         const metaDataFile = (Object.entries(content.files).find(file => file[0] === 'artifacts/metadata.json'));
         const metaDataString = metaDataFile && await metaDataFile[1].async('string') || '[]';
         const metaDataBody: any[] = JSON.parse(metaDataString);
+        //Artifact entries this archive could not resolve.
+        const artifactErrors: string[] = [];
```
dariy-rised: Computed and attached to the returned object, but nothing surfaces it to the user:

- Confirm-import endpoints (`POLICY_IMPORT_FILE(_ASYNC)`) don't forward `policyToImport` at all — they return `result.errors`, an unrelated `ImportPolicyError[]` from a different pipeline stage.
- The preview endpoint does return the raw object, but `preview-policy-dialog.component.ts` destructures known fields only and never reads `.artifactErrors`.

Net effect: a bad archive imports silently with the artifact dropped, no indication to the user — which was half the bug this PR set out to fix.

dariy-rised (changes requested):

vshvets-bc:
Both taken. Pushed in `f05e3bf6`.

**Non-list `metadata.json`.** Closed here as you suggested - the parse now degrades to "no records" rather than throwing:

```ts
const metaDataBody: any[] = Array.isArray(parsedMetaData) ? parsedMetaData : [];
```

It also names the file as the cause once, instead of leaving the user with N identical "no metadata record" lines and no hint that the metadata file itself is the wrong shape. Both `includeArtifactsData` branches are covered.

**The errors reached nobody - agreed, that was half the point of the PR.** They now travel the same path as every other import failure: `artifactErrors` is declared on `IPolicyComponents` in the `{type,uuid,name,error}` shape, `PolicyImport` picks it up from `options.policyComponents`, and `getErrors()` appends it to the `ImportPolicyError[]` the confirm-import endpoints already return as `result.errors`. So a dropped artifact now shows up where a dropped schema or tool does.

I have left the preview dialog alone: surfacing it there means the dialog rendering a field it currently ignores, which is a frontend change of its own rather than part of this fix.

Tests: 11 passing, 8 of them failing on `develop`. New cases cover a non-list metadata file in both parse modes.

guardian-service/src/helpers/import-helpers/policy/policy-import.ts
```diff
@@ -916,8 +920,13 @@ export class PolicyImport {
             tests,
             formulas,
             schemaTemplateSnapshot,
+            artifactErrors,
         } = options.policyComponents;
 
+        //parse-time diagnostics: the artifact was dropped, so say so rather than
+        //importing a policy that silently lost it
+        this.artifactErrors = artifactErrors || [];
```
dariy-rised, line 1071:
```ts
if (result?.errors?.length) {
    const message = PolicyImportExportHelper.errorsMessage(result.errors);
    notifier.fail(message);   // or: return new MessageError(message);
    return;
}
```

`type: 'artifact'` matches neither the `schema` nor the `tool` branch of `errorsMessage`, so the user gets `Failed to import components: others: ["artifacts/orphan: No metadata record matches this artifact."]`.

To be fair on the baseline: these archives already failed on `develop`, so nothing that worked breaks. What changed is what is left behind.

So the archive's valid remainder does import, but the API says it did not, and `startDemo` / `originalTracking` are skipped. Since `updateUUIDs` calls `replaceConfig`, which renames on a name clash, importing again after that verdict yields a second policy. The user-visible outcome is unchanged from `develop`, which is the goal the commit message set: "One malformed entry should not cost the user the rest of a valid archive."

Schemas and tools share this commit-then-fail pattern, and there it is defensible - the policy depends on them. An artifact is a supplementary attachment, which is the distinction the fatal array erases.

Reachable from more than the file endpoints, since `loadPolicyMessage` (`policy-import-helper.ts:280`) also parses a zip: `policy-engine.service.ts:1956`, `:2021`, `:2140`, `:2218`, and `external-policies.service.ts:175`.

**Suggested fix** - return it beside `errors` rather than inside:

```ts
const errors = await this.getErrors(); // unchanged, no artifactErrors folded in
return { policy: row, errors, artifactErrors: this.artifactErrors };
```

`result.errors.length` then stays clean, and a `logger.warn` when the list is non-empty closes the original "reaches nobody" finding server-side without touching the preview dialog.

## Minor

The non-list `metadata.json` entry is pushed *in addition to* the per-entry ones, since `metaDataBody` becomes `[]` and every artifact then fails to resolve - 1 + N, not "once instead of N identical lines". Built the branch and ran it on a non-array `metadata.json` with three artifacts:

```
artifacts: 0
artifactErrors: 4
  - artifacts/metadata.json: Artifact metadata is not a list; no artifact in this archive could be resolved.
  - artifacts/u1: No metadata record matches this artifact.
  - artifacts/u2: No metadata record matches this artifact.
  - artifacts/u3: No metadata record matches this artifact.
```

The test asserts with `.some(...)`, so it passes either way. Also worth a guardian-service test that an unresolvable artifact still yields a *successful* import result - the new cases only cover `parseZipFile` in isolation, and CI has no unit-test job.

dariy-rised (changes requested):

vshvets-bc:
Both right, and the first one was a real regression I introduced in the last round. Pushed in `343d7bf2`.

**`artifactErrors` no longer goes into `errors`.** Your framing is the correct one: schemas and tools commit-then-fail defensibly because the policy depends on them, an artifact is supplementary, and folding it in made callers that branch on `errors.length` report a failure for an import that had committed - skipping `startDemo`/`originalTracking` and setting up the rename-on-retry you describe. Taken as suggested:

```ts
const errors = await this.getErrors();          // unchanged, nothing folded in
...
return { policy: row, errors, artifactErrors: this.artifactErrors };
```

`ImportPolicyResult.artifactErrors` is declared with a note saying why it sits beside rather than inside, and a `logger.warn` fires when the list is non-empty - so the original "reaches nobody" finding is closed server-side without the preview dialog. That also covers the `loadPolicyMessage` callers you listed, since they all go through the same result.

**The 1 + N count - you're right, and my claim was wrong.** The file-level entry was pushed *in addition to* the per-entry ones. Per-entry errors are now suppressed when the metadata file itself is unusable: the entries are not the fault, and three identical "no metadata record" lines add nothing to "the metadata is not a list". Your exact repro now yields:

```
artifacts: 0
artifactErrors: 1
  - artifacts/metadata.json: Artifact metadata is not a list; no artifact in this archive could be resolved.
```

The test asserted with `.some(...)`, which is why it passed either way - it now asserts the full list, so the count is pinned.

**guardian-service test added** (`tests/unit/policy-import-artifact-errors.test.mjs`, 4 cases), since the others only covered `parseZipFile` in isolation:
- an unresolvable artifact leaves `errors` empty
- the diagnostic is still carried, separately
- a genuine schema/tool failure is still reported, without the artifact entry joining it
- one asserting `errorsMessage(['artifact'])` buckets into `others:`, which is the concrete reason it does not belong in that list

11 passing in common, 4 in guardian-service.

guardian-service/src/helpers/import-helpers/policy/policy-import.ts
dariy-rised: Fixed in `343d7bf2` (2026-09-03) - verified against the source at that commit:

- `getErrors()` no longer appends `artifactErrors`; `import()` returns them as a sibling field (`{ policy: row, errors, artifactErrors: this.artifactErrors }`), matching the suggested fix. `errors.length` stays clean for an artifact-only drop, so `notifier.fail`/`MessageError` at the call sites won't fire for this case anymore. `ImportPolicyResult.artifactErrors` is now typed instead of just conditionally spread.
- The "reaches nobody" gap is closed with a `logger.warn` inside `import()` naming the policy id and each dropped path - non-fatal, and doesn't touch the preview dialog, as suggested.
- Minor: the non-list `metadata.json` path now guards the per-entry push with `metaDataUsable`, so it reports once instead of once-per-artifact. New test (`metadata.json that is not a list is reported once, not once per artifact`) pins the exact count.
- New `guardian-service/tests/unit/policy-import-artifact-errors.test.mjs` covers the `getErrors()` boundary directly (artifact-only -> clean errors array, real component failure -> still reported, artifact diagnostic preserved separately).

No remaining objections from me on this thread.

dariy-rised (approved):

common/src/import-export/policy.ts
```diff
@@ -458,21 +465,67 @@ export class PolicyImportExport {
 
         const metaDataFile = (Object.entries(content.files).find(file => file[0] === 'artifacts/metadata.json'));
         const metaDataString = metaDataFile && await metaDataFile[1].async('string') || '[]';
-        const metaDataBody: any[] = JSON.parse(metaDataString);
+        //Artifact entries this archive could not resolve.
+        const artifactErrors: IArtifactError[] = [];
+        const parsedMetaData = JSON.parse(metaDataString);
```
Pyatakov: Malformed `metadata.json` still throws out of `parseZipFile`. `JSON.parse` is unguarded; the new `Array.isArray` check only sees values it already returned. `artifacts/metadata.json` containing `[{"uuid":` raises a raw `SyntaxError`, the same whole-import abort this PR sets out to remove, and the PR's own comment names "a truncated archive" as a cause.
Suggest wrapping the parse in a `try` and falling through to the existing `metaDataUsable` branch closes it.

common/src/import-export/policy.ts
```diff
@@ -458,21 +465,67 @@ export class PolicyImportExport {
...
         if (includeArtifactsData) {
-            const data = fileEntries.filter(file => /^artifacts\/.+/.test(file[0]) && file[0] !== 'artifacts/metadata.json').map(async file => {
-                const uuid = file[0].split('/')[1];
-                const artifactMetaData = metaDataBody.find(item => item.uuid === uuid);
+            /*
+             * A missing metadata record used to dereference `undefined` inside a
+             * Promise.all, so one bad entry aborted the whole import with a raw
+             * TypeError. Three ways in, none needing malice: no metadata.json, a
+             * truncated archive, or a nested path whose split('/')[1] matches no uuid.
+             */
```
Pyatakov: Adjust verbose comments here and above at the lines 471-473

Pyatakov (changes requested):

vshvets-bc:
Correct, and it is the same failure the PR set out to remove - my `Array.isArray` check only ever saw what `JSON.parse` had already returned. Pushed in `66e376e9`.

The parse is inside the guard now, and both faults fall through the one branch:

```ts
try {
    parsedMetaData = JSON.parse(metaDataString);
    if (!Array.isArray(parsedMetaData)) {
        metaDataFault = 'is not a list';
    }
} catch (error) {
    metaDataFault = 'is not valid JSON';
}
```

The two are reported distinctly - `is not valid JSON` vs `is not a list` - since they call for different corrections to the archive.

Two tests added for the case you gave, `[{"uuid":`, one per parse mode: **13 passing**. Comments at 471-473 and 491 cut to two lines each.

Pyatakov (approved):

Issue #6791, opened 2026-08-28
Bulk schema delete silently skips schemas that are still referenced, and returns {} for the per-schema outcome


vshvets-bc:
`DELETE_SCHEMAS` (`guardian-service/src/api/schema.service.ts`) has two problems that combine badly: it skips schemas that are still referenced by a policy schema without telling anyone, and the per-schema outcome it does return cannot survive the message bus.

Both are present on `develop` (`6ff0019f7`).

## Defect 1: the result is a `Map`, so the caller receives `{}`

```ts
const results = new Map<string, boolean>();
...
    const result = await deleteSchema(schema.id, owner, deleteSchemaStep);
    results.set(schema.id, result);
...
notifier.result(results);
```

A `Map` has no JSON representation — `JSON.stringify(new Map([['a', true]]))` is `{}`. Once the task result crosses the message bus the caller gets an empty object, so the per-schema `true`/`false` outcome is unreachable by design.

## Defect 2: blocked schemas are skipped with no report

The handler computes `blockedChildren` in full — the same dependency walk `GET_SCHEMA_DELETION_PREVIEW` performs — and then discards it. Schemas in that set are skipped:

```ts
for (const schema of schemas) {
    if (blockedSchemaIds.has(schema.iri)) {
        continue;
    }
```

with no error, no mention anywhere, and the task reporting success. A user selects five schemas, sees a green task, and finds three of them still present with nothing explaining why.

Note the blocked set is not only *children* despite the name — the branch that adds a directly-requested schema to `blockedSchemaIds` pushes it to `blockedChildren` in the same step.

## The frontend is already wired to display this

`frontend/src/app/modules/common/async-progress/async-progress.component.ts` already reads `result?.errors` in `reportSchemaImportErrors` and renders a sticky toast naming each failed item, and `TaskAction.DELETE_SCHEMAS` is already in the case list that calls it. It never fires today only because `errors` is never populated on this path.

One caveat for whoever fixes this: that method's strings are hardcoded for the import path (`'1 schema was not imported'`). Populating `errors` for a delete without touching them makes a blocked delete report *"1 schema was not imported"*.

## Reproduce

1. Create schema A, and a policy schema B whose document `$ref`s A.
2. Bulk-delete A.
3. A is silently skipped — the task succeeds, no toast, nothing in the result explains it. The per-schema outcome is `{}`.

## Suggested fix

[...]

PR #6792, opened 2026-08-28
fix(schemas): report schemas a bulk delete skipped, and make the result serialisable

vshvets-bc:
Fixes #6791.

`DELETE_SCHEMAS` silently skips schemas still referenced by a policy schema, and returned `{}` for the per-schema outcome. Two problems in one handler.

## `results` was a `Map`, so the caller received `{}`

`JSON.stringify(new Map([['a', true]]))` is `{}`. Once the task result crossed the message bus the per-schema `true`/`false` outcome was unreachable by design. It is now an array of `{ id, name, deleted }`.

## Blocked schemas were skipped with no report

`blockedChildren` was computed in full — the same dependency walk `GET_SCHEMA_DELETION_PREVIEW` uses — and then discarded, while schemas in it were skipped via `blockedSchemaIds`. No error, no mention, task reports success. A user selects five schemas, sees a green task, and finds three still there.

The blocked set is now surfaced as an `errors` array naming each skipped schema and what blocks it.

Worth noting the set covers directly-requested schemas, not only children despite the name: the branch that adds a top-level schema to `blockedSchemaIds` pushes it to `blockedChildren` in the same step, so `errors` and the skip set cannot diverge.

## Toast wording, bundled deliberately

The frontend already had the receiving half — `reportSchemaImportErrors` reads `result.errors` and renders a sticky toast, and `TaskAction.DELETE_SCHEMAS` is already in the case list that calls it. It simply never fired, because nothing populated `errors` on that path.

But its strings were written for import and hardcoded, so a schema blocked during a **delete** would have surfaced as *"1 schema was not imported"*. Since this change is what first populates `errors` there, shipping the backend fix alone would activate that. The wording is parameterised in the same commit: `reportSchemaImportErrors` becomes `reportSchemaErrors(result, verb)`, the delete case passes `'deleted'`, every other caller keeps `'imported'`.

## Tests

Adds `guardian-service/tests/unit/bulk-schema-delete-blocked-result.test.mjs`, using the existing `_handler-harness.mjs`.

**Both cases fail on `develop`:**

| Case | On `develop` |
|---|---|
| reports a blocked schema instead of silently skipping it | **fails** — no `errors` array on the result |
| carries the per-schema outcome in a JSON-serialisable shape | **fails** — `results must be an array, not a Map` |

Verified by stashing only the source change, rebuilding `guardian-service`, and re-running: 2 failing. With the fix restored: 2 passing.

One note for anyone extending these: the handler calls `RunFunctionAsync(...)` without awaiting it and returns the task immediately, so the test has to hold the promise its stub creates and await that, or it races past the assertion.

## Risk

Confined to the `DELETE_SCHEMAS` task result and the toast wording. Deletion logic — which schemas actually get deleted — is unchanged; only what is reported about it. No schema or data migration.

### DIFF
--- a/frontend/src/app/modules/common/async-progress/async-progress.component.ts
+++ b/frontend/src/app/modules/common/async-progress/async-progress.component.ts
@@ -461,7 +461,7 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
             case TaskAction.DELETE_SCHEMAS:
             case TaskAction.IMPORT_SCHEMA_FILE:
             case TaskAction.IMPORT_SCHEMA_MESSAGE:
-                this.reportSchemaImportErrors(result);
+                this.reportSchemaErrors(result, this.action === TaskAction.DELETE_SCHEMAS ? 'deleted' : 'imported');
                 if (this.last) {
                     const schemaId = typeof result === 'string' && result ? result : null;
                     const lastWithSchema = schemaId
@@ -610,7 +610,12 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
      * ImportSchemaResult.errors is {type,uuid,name,error} and nothing downstream renders
      * it, so flatten it into one sticky toast naming each schema that failed.
      */
-    private reportSchemaImportErrors(result: any): void {
+    /**
+     * Report per-schema failures carried on a task result.
+     * @param result task result
+     * @param verb what did not happen to the schemas, e.g. 'imported' or 'deleted'
+     */
+    private reportSchemaErrors(result: any, verb: string = 'imported'): void {
         const errors = result?.errors;
         if (!Array.isArray(errors) || !errors.length) {
             return;
@@ -619,10 +624,10 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
             .map((e: any) => (e?.name ? `${e.name}: ${e.error}` : e?.error))
             .filter((line: any) => !!line)
             .join('\n');
-        const msg = text || 'Some schemas could not be imported.';
+        const msg = text || `Some schemas could not be ${verb}.`;
         this.toastService.warn(
             msg,
-            errors.length === 1 ? '1 schema was not imported' : `${errors.length} schemas were not imported`,
+            errors.length === 1 ? `1 schema was not ${verb}` : `${errors.length} schemas were not ${verb}`,
             { sticky: true, logMessage: msg }
         );
     }
--- a/frontend/src/app/modules/common/async-progress/async-progress.component.ts
+++ b/frontend/src/app/modules/common/async-progress/async-progress.component.ts
@@ -607,11 +607,8 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
     }
 
     /**
-     * ImportSchemaResult.errors is {type,uuid,name,error} and nothing downstream renders
-     * it, so flatten it into one sticky toast naming each schema that failed.
-     */
-    /**
-     * Report per-schema failures carried on a task result.
+     * Report per-schema failures carried on a task result. Nothing downstream renders
+     * `errors`, so flatten it into one sticky toast naming each schema that failed.
      * @param result task result
      * @param verb what did not happen to the schemas, e.g. 'imported' or 'deleted'
      */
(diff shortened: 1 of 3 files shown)


guardian-service/src/api/schema.service.ts
```diff
@@ -1922,7 +1922,16 @@ export async function schemaAPI(logger: PinoLogger): Promise<void> {
                 }
 
                 const stepMap = new Map<string, NotificationStep>();
-                const results = new Map<string, boolean>();
+                const results: Array<{ id: string, name: string, deleted: boolean }> = [];
+                // A schema still referenced by a policy schema is skipped below via
+                // blockedSchemaIds; without this the caller is never told, and the task
+                // reports success for a delete that did not happen.
+                const errors = blockedChildren.map(blocked => ({
```
dariy-rised:
```ts
const errors = blockedChildren.map(blocked => ({
    type: 'schema',
    uuid: blocked.schema.iri,
    name: blocked.schema.name,
    error: 'Still referenced by ' + blocked.blockingSchemas.map(s => s.name).join(', ')
}));
```

`blockedChildren` is computed unconditionally over both directly-requested schemas *and* their child dependency schemas (via the `else` branch at line 1882, which exists specifically to catch a policy schema blocking a *child* of a requested schema). Whether those children are ever candidates for deletion depends entirely on `includeChildren`:

```ts
if (includeChildren) {
    const deletableChildren = Array.from(childSchemas.values())
        .filter(child => !blockedSchemaIds.has(child.iri) && ...)
    ...
}
```

When `includeChildren` is `false` — the default in `schema-delete-dialog.component.ts:15`, and a state the UI already reaches routinely since the preview dialog shows `blockedChildren` as informational content before the user decides whether to check the box — a blocked child schema was never a delete candidate in the first place. But it still gets pushed into `errors`, so the caller gets a toast saying "1 schema was not deleted: C — Still referenced by D" for a schema C that nothing ever attempted to delete.

guardian-service/tests/unit/bulk-schema-delete-blocked-result.test.mjs
dariy-rised, line 1: The added test only covers a *directly requested* schema being blocked (`includeChildren: false`, target itself referenced by the blocking policy schema). It doesn't cover the child-schema case above — a child of a requested schema being blocked by an unrelated policy schema while `includeChildren` is `false`. That's the exact case that would have caught Comment 1; worth adding a fixture where the requested schema has a `$ref`'d child and a *different* schema (not the one requested) is what blocks the child.

frontend/src/app/modules/common/async-progress/async-progress.component.ts
```diff
@@ -610,7 +610,12 @@ export class AsyncProgressComponent implements OnInit, OnDestroy {
      * ImportSchemaResult.errors is {type,uuid,name,error} and nothing downstream renders
      * it, so flatten it into one sticky toast naming each schema that failed.
      */
-    private reportSchemaImportErrors(result: any): void {
+    /**
```
dariy-rised:
```ts
/**
 * ImportSchemaResult.errors is {type,uuid,name,error} and nothing downstream renders
 * it, so flatten it into one sticky toast naming each schema that failed.
 */
/**
 * Report per-schema failures carried on a task result.
 * @param result task result
 * @param verb what did not happen to the schemas, e.g. 'imported' or 'deleted'
 */
private reportSchemaErrors(result: any, verb: string = 'imported'): void {
```

The original doc comment for `reportSchemaImportErrors` was left in place when the method was renamed/generalized to `reportSchemaErrors`, and a new doc block was stacked directly above it. Minor, but it reads as two competing descriptions of the same method — worth merging into one comment before merge.

dariy-rised (changes requested):

vshvets-bc:
Both correct. Pushed in `4e48b2c9`.

**Blocked children were reported when they were never delete candidates.** `blockedChildren` is built regardless of `includeChildren`, so with the box unchecked the caller got a toast about a schema nothing had tried to delete. The error list is now filtered to schemas that were actually up for deletion:

```ts
const requestedIris = new Set(schemas.map(schema => schema.iri));
const errors = blockedChildren
    .filter(blocked => includeChildren || requestedIris.has(blocked.schema.iri))
    .map(...)
```

**Test added for exactly that case**, as you described - a requested schema with a `$ref`'d child, blocked by a different policy schema than the one blocking the target:
- `says nothing about a blocked child that was never a delete candidate` (`includeChildren: false` -> only the target is reported)
- `reports that same blocked child once children are in scope` (`includeChildren: true` -> both)

The first fails without the filter. The fixture harness now takes the requested/child/scope sets rather than hard-coding them.

**Stacked doc comments merged** into one on `reportSchemaErrors`.

4 passing.

dariy-rised (approved):

Issue #6793, opened 2026-08-31
PolicyContainer: a saturated policy-service fleet stops scaling, and failed spawns permanently strand policy slots


vshvets-bc:
`PolicyContainer` has four defects that each strand policy slots. Individually they leak capacity; together they produce a fleet that stops scaling while policies have nowhere to run, and requests to those policies wait out the full Block Timeout because nothing is hosting them.

All four are in [`policy-service/src/helpers/policy-container.ts`](https://github.com/hashgraph/guardian/blob/develop/policy-service/src/helpers/policy-container.ts) and none need special configuration to reach.

## 1. The unsubscribe timer cancels the wrong subscription

`unsubscribeFromModelGeneration()` dereferences `this.generatePolicySubscription` *inside* its 500 ms timer, so it cancels whatever is current when the timer fires — not what was live when it was scheduled:

```ts
private unsubscribeFromModelGeneration(): void {
    setTimeout(() => {
        this.generatePolicySubscription.unsubscribe();   // <- re-read 500ms later
    }, 500);
}
```

1. The service fills up → `addPolicy` refuses → schedules the unsubscribe.
2. Inside that window a policy exits → `subscribeForModelGeneration()` installs a **new** subscription.
3. The stale timer fires and tears down that **new** subscription.

The service is left below capacity and deaf to `GENERATE_POLICY`, recovering only if another policy happens to exit. It also still advertises free slots, which feeds defect 2.

## 2. Capacity is reported without checking whether the service is listening

```ts
free: this.processCount < this.maxPolicyInstances,
```

`processCount` says nothing about whether `GENERATE_POLICY` is subscribed. A service in the state above reports `free: true` forever, and because `checkForRunNewInstance` bails out when *any* peer reports free, one such service blocks scale-up for the entire fleet.

## 3. Scale-up is sealed for the lifetime of the process

`startNewPolicyServiceTriggered` is set when a replica is requested and cleared **only** in `subscribeForModelGeneration()`, which on this path runs only when a policy *exits*. So a service that fills up and stays full triggers scale-up exactly once, ever.

`checkForRunNewInstance` also gives up if a single free slot exists anywhere:

```ts
let hasFree = false;
for (const info of freeCheck) { hasFree = hasFree || info.free }
if (hasFree) { return; }
```

One spare slot against a fleet that is dozens of policies short is not headroom.

## 4. A failed spawn strands the slot; a crashing policy never gives up

```ts
p.once('error', (error) => {
[...]

PR #6794, opened 2026-08-31
fix(policy-service): keep a saturated fleet scaling and stop losing policy slots

vshvets-bc:
Fixes #6793.

## Problem

Four defects in `PolicyContainer`, each of which strands policy slots. Together they produce a fleet that stops growing while policies have nowhere to run — and a request to an unhosted policy then waits out the full Block Timeout, because there is genuinely no responder.

**The unsubscribe timer cancels whatever is current, not what it was scheduled for.** `unsubscribeFromModelGeneration()` re-reads `this.generatePolicySubscription` inside its 500 ms timer. A policy exiting inside that window calls `subscribeForModelGeneration()`, and the stale timer then tears down that *new* subscription. The service is left below capacity and deaf to `GENERATE_POLICY`.

**Capacity is reported from `processCount` alone**, so the service above still advertises `free: true`. `checkForRunNewInstance` bails when any peer reports free, so one such service suppresses scale-up for the entire fleet.

**`startNewPolicyServiceTriggered` is a permanent seal.** It is cleared only in `subscribeForModelGeneration()`, which on the scaling path runs only when a policy *exits*, so a service that fills up and stays full asks for a replica exactly once per lifetime. The `hasFree` bail-out compounds it: a single spare slot anywhere blocks scaling no matter how far short the fleet is.

**A failed spawn strands the slot.** Node emits `'error'` when a process cannot be spawned and `'exit'` may never follow; the handler only logged. `instance.process` is already assigned and `runPolicyProcess` skips instances that have one, so the policy stayed wedged forever while still counting against `processCount`. The non-zero-exit branch has the opposite problem — re-forking every 10 s with no backoff and no ceiling.

## Fix

**Generation counter for the unsubscribe.** `subscribeForModelGeneration()` bumps a counter; the pending timer no-ops if the generation changed, so a re-subscribe inside the window survives. It also drops any prior subscription before installing a new one, so repeated exit/respawn cycles cannot accumulate handlers.

**`canAcceptPolicies`** gates the advertised capacity on `intakeSubscribed && processCount < max`, and `count` reports 0 when the service cannot accept work. A service that is not listening no longer suppresses fleet-wide scaling.

**The seal becomes a timestamp** (`SCALE_COOLDOWN_MS`, default 60 s) so a fleet that stays saturated keeps scaling, and **`hasFree` becomes a headroom test** — scale when total free slots fall below one service's worth (`SCALE_HEADROOM_SLOTS`, defaulting to `MAX_POLICY_INSTANCES`).

Making the seal re-armable makes a stampede reachable where it previously was not — with a permanent seal, "one replica per window" was emergent, since each service could scale at most once ever. So the winner is now elected explicitly: services stand down if any peer is inside its cooldown, and otherwise the lowest `instanceId` scales. `startNewPolicyServiceTriggered` rides along in the existing scatter reply; no new transport. A synchronous in-flight guard covers re-entry, since the 1 s interval does not await `checkForRunNewInstance` and the scatter takes ~500 ms.

**Spawn failure clears `instance.process`** so the 1 s sweep re-forks on its next tick, and **crash respawns back off exponentially** (`POLICY_RESTART_BASE_DELAY_MS` → `POLICY_RESTART_MAX_DELAY_MS`) up to `POLICY_MAX_RESTART_ATTEMPTS`, after which the policy *releases* its slot rather than holding one it never uses. guardian-service can still assign it again through `GENERATE_POLICY`, so this is a local stop, not a permanent verdict. A clean exit resets the count.

**Reply bookkeeping.** The handler now drops replies for ids it is not collecting, and `getFreePolicyServices` registers the bucket before publishing and always resolves an array. That fixes both the per-late-reply Map leak and the `undefined` return that `checkForRunNewInstance` could not distinguish from "the fleet is full" — the new empty-set check treats no answers as no information rather than as zero capacity.

All five new env vars are optional and defaulted; behaviour with none of them set is the intended one.

## Tests

`policy-service/tests/unit-tests/helpers/policy-container-lifecycle.test.mjs` — 19 specs across four groups: the intake race and honest capacity reporting, scale-up (cooldown re-arm, headroom, empty reply set, peer cooldown, election, re-entry), spawn failure and crash-loop backoff/ceiling/reset, and the free-status reply bookkeeping.

19 pass with the fix. Against unmodified `develop`, **15 of the 19 fail**.

```
$ yarn workspace policy-service run build
$ node_modules/.bin/mocha --no-config --exit \
    --spec policy-service/tests/unit-tests/helpers/policy-container-lifecycle.test.mjs
  19 passing
```

`yarn lint` in `policy-service` is clean.

## Note, deliberately not included

`develop` sets the stop script from the wrong variable:

```ts
this.stopServiceScript = process.env.RUN_SERVICE_SCRIPT;   // policy-container.ts:160
```

so the scale-down path at the empty-service branch runs the **start** script. It is unrelated to these four defects and wants its own explanation, so it is left untouched here — happy to send it separately.

### DIFF
--- a/policy-service/src/helpers/policy-container.ts
+++ b/policy-service/src/helpers/policy-container.ts
@@ -78,6 +78,11 @@ export interface IPolicyServiceInfo {
      */
     requestId: string
 
+    /**
+     * Whether this service has recently asked for a new replica
+     */
+    startNewPolicyServiceTriggered?: boolean
+
 }
 
 /**
@@ -147,10 +152,81 @@ export class PolicyContainer extends NatsService {
     private generatePolicySubscription: Subscription;
 
     /**
+     * Identity of the live GENERATE_POLICY subscription. A pending unsubscribe
+     * timer only tears down the generation it was scheduled for.
+     * @private
+     */
+    private generateSubscriptionGeneration: number = 0;
+
+    /**
+     * Whether this service is currently listening for GENERATE_POLICY.
+     * @private
+     */
+    private intakeSubscribed: boolean = false;
+
+    /**
+     * A service has usable capacity only when it is below its limit AND actually
+     * listening. Reporting slots it will never fill keeps every other service's
+     * `hasFree` true, which suppresses scale-up across the whole fleet.
+     * @private
+     */
+    private get canAcceptPolicies(): boolean {
+        return this.intakeSubscribed && this.processCount < this.maxPolicyInstances;
+    }
+
+    /**
+     * Consecutive failed starts per policy, for the crash-loop backoff.
+     * @private
+     */
+    private readonly restartAttempts: Map<string, number> = new Map();
+    private readonly maxRestartAttempts: number = process.env.POLICY_MAX_RESTART_ATTEMPTS
+        ? parseInt(process.env.POLICY_MAX_RESTART_ATTEMPTS, 10)
+        : 10;
+    private readonly restartBaseDelayMs: number = process.env.POLICY_RESTART_BASE_DELAY_MS
+        ? parseInt(process.env.POLICY_RESTART_BASE_DELAY_MS, 10)
+        : 10000;
+    private readonly restartMaxDelayMs: number = process.env.POLICY_RESTART_MAX_DELAY_MS
+        ? parseInt(process.env.POLICY_RESTART_MAX_DELAY_MS, 10)
+        : 5 * 60 * 1000;
+
+    /**
+     * When this service last asked for a replica. Replaces the permanent boolean
+     * seal so a fleet that stays saturated can keep scaling.
+     * @private
+     */
+    private lastScaleTriggeredAt: number = 0;
+
+    /**
+     * Guards checkForRunNewInstance against overlapping runs.
+     * @private
+     */
+    private scaleCheckInFlight: boolean = false;
+
+    private readonly scaleCooldownMs: number = process.env.SCALE_COOLDOWN_MS
+        ? parseInt(process.env.SCALE_COOLDOWN_MS, 10)
+        : 60000;
+
+    /**
+     * Fleet headroom, in policy slots, below which a full service asks for another
+     * replica. Defaults to one service's worth of capacity.
+     * @private
+     */
+    private readonly scaleHeadroomSlots: number = process.env.SCALE_HEADROOM_SLOTS
+        ? parseInt(process.env.SCALE_HEADROOM_SLOTS, 10)
+        : 0;
+
+    /**
      * Start new policy-service triggered
      * @private
      */
-    private startNewPolicyServiceTriggered: boolean = false;
+    private get startNewPolicyServiceTriggered(): boolean {
+        return this.lastScaleTriggeredAt > 0
+            && (Date.now() - this.lastScaleTriggeredAt) < this.scaleCooldownMs;
+    }
+
+    private set startNewPolicyServiceTriggered(value: boolean) {
+        this.lastScaleTriggeredAt = value ? Date.now() : 0;
+    }
 
     constructor(private readonly logger: PinoLogger) {
         super();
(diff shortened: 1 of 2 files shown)


policy-service/src/helpers/policy-container.ts
```diff
@@ -269,26 +371,68 @@ export class PolicyContainer extends NatsService {
...
+
+            /*
+             * Only one service should scale per cooldown. With a permanent seal that
+             * was emergent - each service scaled at most once ever - but a re-arming
+             * seal makes a stampede reachable, so the winner is elected explicitly.
+             */
+            if (freeCheck.some(info => info.instanceId !== this.instanceId && info.startNewPolicyServiceTriggered)) {
+                return;
+            }
+            const candidates = freeCheck
```
dariy-rised, line 425: The election candidate list isn't filtered to peers that are actually saturated:

```ts
const candidates = freeCheck
    .filter(info => !info.startNewPolicyServiceTriggered)
    .sort((a, b) => a.instanceId.localeCompare(b.instanceId));
```

`freeCheck` includes every peer that replied, free or full. If a free peer's `instanceId` sorts first, it becomes `candidates[0]` and every saturated peer defers to it (`startNewPolicyServiceTriggered = true`, holding off for a full cooldown). But that free peer's own `checkForRunNewInstance()` returns at the very first line (`processCount < maxPolicyInstances`) - it never reaches this code, so it can never actually call `execFile`. As long as that peer's ID keeps sorting first and it stays non-saturated, the whole fleet defers to a "leader" that never acts, and scale-up stalls - the same symptom this PR is fixing, just via the new election path.

Suggest filtering candidates to peers that would actually act if elected, e.g. `freeCheck.filter(info => !info.free && !info.startNewPolicyServiceTriggered)`.

policy-service/src/helpers/policy-container.ts
```diff
@@ -326,12 +470,26 @@ export class PolicyContainer extends NatsService {
...
+            /*
+             * Node emits 'error' when a process cannot be SPAWNED (EAGAIN, ENOMEM, a
+             * bad process path) and 'exit' may never follow, so nothing else runs.
+             * `instance.process` is already assigned and runPolicyProcess skips any
+             * instance that has one, so the policy stayed wedged forever - still
+             * counted in processCount, holding a slot it never used, and with no
+             * responder for requests addressed to it.
+             *
+             * Clearing the handle is enough: the 1s sweep re-forks on its next tick,
+             * and the crash-loop guard below bounds the retries if it keeps failing.
```
dariy-rised:
```ts
// Clearing the handle is enough: the 1s sweep re-forks on its next tick,
// and the crash-loop guard below bounds the retries if it keeps failing.
```

but the crash-loop guard (`restartAttempts`, backoff, `maxRestartAttempts` ceiling) only lives in the `'exit'` handler's non-zero-code branch below - it's never touched here. Per the comment a few lines up, a process that fails to spawn may only ever emit `'error'` with no following `'exit'`. In that case this policy retries every ~1s tick indefinitely, with no backoff and no ceiling - the same "hammer the service forever" failure this PR set out to fix for the crash-loop case, just reachable through the spawn-failure path instead.

Suggest routing this through the same `restartAttempts`/backoff/ceiling logic used in the `'exit'` handler (or extracting it into a shared helper both handlers call).

policy-service/src/helpers/policy-container.ts
```diff
@@ -269,26 +371,68 @@ export class PolicyContainer extends NatsService {
...
 
-        execFile(this.runServiceScript, (error, _data) => {
-            if (error) {
-                this.logger.error(error, ['POLICY_SERVICE', this.runServiceScript], null);
+            /*
+             * Scale on fleet headroom rather than "any service has a slot". This one
+             * is already full; if what is left across the fleet is under a single
+             * service's worth, another replica is warranted.
+             */
+            const headroom = Math.max(1, this.scaleHeadroomSlots || this.maxPolicyInstances);
```
dariy-rised:
```ts
const headroom = Math.max(1, this.scaleHeadroomSlots || this.maxPolicyInstances);
```

`this.scaleHeadroomSlots` defaults to `0`, and `0` is falsy, so `scaleHeadroomSlots || this.maxPolicyInstances` always falls through to `maxPolicyInstances` whenever the value is `0` — including when an operator explicitly sets `SCALE_HEADROOM_SLOTS=0` to mean "scale as soon as there's zero free capacity, no buffer." That intent is silently discarded and treated the same as "unset."

Suggest distinguishing "unset" from "explicitly 0", e.g. checking `process.env.SCALE_HEADROOM_SLOTS === undefined` at the definition site rather than relying on `||` at the usage site.

dariy-rised (changes requested):

vshvets-bc:
All three were right, and two of them defeated the fix this PR exists for. Pushed in `712e1c5d`.

**A free peer could be elected and never act.** `freeCheck` carries every responder, so a peer with headroom could sort first and take the election while its own `checkForRunNewInstance()` returns at the first line - the fleet parks behind a leader that cannot spawn. Candidates are now `!info.free && !info.startNewPolicyServiceTriggered`, so only a peer that would actually reach the spawn can win.

**`SCALE_HEADROOM_SLOTS=0` was discarded.** Fixed at the definition site as you suggested: the field is `number | null`, `null` only when the variable is unset, and the usage resolves `null` to `maxPolicyInstances`. An explicit `0` now survives to `Math.max(1, ...)`, which is what turns "no buffer" into "scale once nothing is free" - `0` unclamped would have meant `totalFree >= 0` and never scaling at all.

**The spawn-failure path had no ceiling.** You're right that `restartAttempts` lived only in the `'exit'` non-zero branch. Extracted it into `scheduleRestart(policyId, instance, policyOwnerId)`, called from both handlers, so a spawn that keeps failing backs off and eventually releases its slot instead of being retried every ~1s forever. The comment that claimed the guard applied is gone.

Tests: 23 passing, 3 of them failing on `develop` before this commit - one per finding:
- `a free peer cannot be elected, however its id sorts`
- `an explicit SCALE_HEADROOM_SLOTS=0 is honoured, not read as unset` (plus a companion asserting it still scales at zero free)
- `A2: a spawn that keeps failing gives up instead of retrying forever`

`A: releases the instance when the process cannot be spawned` now waits for the backoff timer rather than asserting an inline release.

dariy-rised (approved):

Issue #6806, opened 2026-09-01
worker-service: a throw in runTask leaves the worker permanently marked busy and out of the pool


vshvets-bc:
## Problem

`runTask` in `worker-service/src/api/worker.ts` marks the worker busy on entry and clears that flag only with the **last statement of the happy path**:

```ts
const runTask = async (task, completeEvent = WorkerEvents.TASK_COMPLETE) => {
    this.isInUse = true;
    this.currentTaskId = task.id;
    ...
    const result = await this.processTaskWithTimeout(task);
    ...
    await this.publish(completeEvent, ...);      // unguarded
    await this.publish(WorkerEvents.WORKER_READY);
    this.isInUse = false;                        // only reached on success
}
```

There is no `try/finally`. The only `try` is a narrow one around result logging. A throw at any of the awaits in between leaves `isInUse` stuck at `true` for the lifetime of the process.

## Consequences

`GET_FREE_WORKERS` is gated on the flag, so a stuck worker stops answering:

```ts
this.subscribe(WorkerEvents.GET_FREE_WORKERS, async (msg) => {
    if (!this.isInUse) { this.publish(msg.replySubject, { ... }) }
});
```

and `WORKER_READY` is never re-published. The worker is silently removed from the pool while its liveness probe keeps passing — the pod looks healthy and does nothing.

The requester never receives a completion and waits out its own timeout.

Both call sites invoke `runTask` **without `await` and without `catch`**:

```ts
this.getMessages([this.replySubject, WorkerEvents.SEND_TASK_TO_WORKER].join('.'), async (task) => {
    if (!this.isInUse) {
        runTask(task);
        return new MessageResponse({ result: true })
```

so the rejection has nowhere to surface. `worker-service` registers no `unhandledRejection` or `uncaughtException` handler (`src/index.ts`, `src/app.ts`, `src/app-instance.ts` — none), so under Node's default `--unhandled-rejections=throw` the process exits instead. Either way the worker is lost with no error attributable to the task that caused it.

## Most likely trigger

Not a task failure — a **publish** failure on `TASK_COMPLETE`. `result.data` for `GET_FILE` and `ADD_FILE` carries file bodies, so an oversized result is rejected when publishing rather than when running. That path is entirely outside any `try`.

## Second, smaller defect

The result-logging catch calls `clearState()`:

```ts
} catch (error) {
    this.logger.error(error.message, [this.workerID, 'WORKER'], userId);
    this.clearState();      // sets isInUse = false
[...]

PR #6807, opened 2026-09-01
fix: 6806 release the worker on every exit path from runTask

vshvets-bc:
Fixes #6806.

## Problem

`runTask` marks the worker busy on entry and clears the flag only with the **last statement of the happy path**. There is no `try/finally` — the only `try` is a narrow one around result logging — so a throw at any await in between leaves `isInUse` stuck at `true` for the lifetime of the process.

`GET_FREE_WORKERS` is gated on that flag and `WORKER_READY` is never re-published, so the worker is silently removed from the pool while its liveness probe keeps passing. The requester never gets a completion and waits out its own timeout.

Both call sites invoke `runTask` **unawaited and uncaught**, and the service registers no `unhandledRejection` handler, so the rejection either vanishes or takes the process down.

The realistic trigger is a **publish** failure on `TASK_COMPLETE` rather than a task failure: `result.data` for `GET_FILE` and `ADD_FILE` carries file bodies, so an oversized result is rejected at publish time — a path entirely outside any `try`.

## Fix

- The body runs inside `try { … } finally { … }`, so `isInUse` is released on every exit path.
- `WORKER_READY` is re-published **from the `finally`**. Recovering the flag alone is not enough: without re-announcing, the worker still sits out of the pool.
- The completion publish is guarded. A result that cannot be transmitted is reported back on the same event, so the requester fails fast instead of waiting out its timeout, and the original error is logged.
- Drops `clearState()` from the result-logging catch. That released `isInUse` **mid-task**, so a worker still running one task could be handed another — a second, smaller defect on the same path.

No behaviour change on the happy path.

## Tests

`worker-service/tests/worker-runtask-release.test.mjs` — 6 cases driving the real `SEND_TASK_TO_WORKER` handler through `init()`, with the broker stubbed:

- releases `isInUse` after a normal task *(happy-path control)*
- releases `isInUse` when the task itself throws
- releases `isInUse` when publishing the result fails
- re-publishes `WORKER_READY` even when the task threw
- still answers `GET_FREE_WORKERS` after a failed task
- reports an error to the requester when the result cannot be delivered

**Five of the six fail against current `develop`**; the sixth is the happy-path control and passes on both, as intended.

Full `worker-service` suite after the change: **289 passing**.

## Note on scope

I also wrote a seventh case for the mid-task `clearState()` release. It passed against unmodified `develop` as well, so it was not actually discriminating — on `develop` the logger throw escapes that catch before the assertion point. Rather than ship a test that looks like it proves something it does not, I removed it. The fix still drops that `clearState()`, and the reasoning is in #6806.

## Not addressed here

Whether `TASK_COMPLETE` should route large `data` through `LargePayloadContainer` — which already exists for exactly this — is a separate change. This PR makes the failure survivable and attributable; it does not stop the oversized publish from failing in the first place.


vshvets-bc:
**Correction to the "most likely trigger" section — I got that part wrong.**

I claimed an oversized `TASK_COMPLETE` is rejected at publish time. It is not. `NatsService.publish` encodes through `ZipCodec`, which already offloads anything over the gate:

```ts
// common/src/mq/zip-codec.ts
const zipped = JSONCodec().encode(d);
const maxPayload = parseInt(process.env.MQ_MAX_PAYLOAD, 10);
const headerReserved = 32 * 1024;

if (Number.isInteger(maxPayload) && maxPayload <= (zipped.length + headerReserved)) {
    const directLink = new LargePayloadContainer().addObject(Buffer.from(zipped));
    return JSONCodec().encode({ directLink })
}
```

So a large `GET_FILE` / `ADD_FILE` result is already moved out of the message body and replaced with a small `{ directLink }` envelope. Payload size is not what makes that publish throw.

**One caveat.** The gate only applies when `MQ_MAX_PAYLOAD` parses as an integer. With it unset, nothing is ever offloaded and an oversized publish would fail at the broker — so the trigger I described is reachable, but only in a deployment that does not set it. It is set (1 MB) in the deployment I was looking at, so that was not the mechanism there.

**What this does not change.** The defect and the fix stand on their own: `isInUse` is released only on the happy path, and *any* throw between entry and that last statement strands the worker. Realistic sources are transport failures rather than payload size — a NATS connection drop mid-publish, or the logger call throwing (this fork calls `this.logger.*` without `await`, so a rejected logger promise surfaces differently, but a synchronous throw still escapes).

The point of the fix is that the worker should survive *whatever* throws, and not depend on enumerating the causes correctly. Which, as this correction shows, is exactly the assumption worth not making.

The tests are unaffected — they induce the failure by making the task and the publish throw directly, not by size.

worker-service/src/api/worker.ts
```diff
@@ -187,41 +187,84 @@ export class Worker extends NatsService {
...
+                /*
+                 * Re-announce before releasing. A worker that recovered its flag but
+                 * never re-published WORKER_READY would still sit out of the pool.
+                 */
+                try {
+                    await this.publish(WorkerEvents.WORKER_READY);
+                } catch (error) {
+                    this.logger.error(error.message, [this.workerID, 'WORKER'], userId);
+                }
+                this.clearState();
```
ihar-tsykala: ## `worker-service/src/api/worker.ts:256–267` — the new `finally`

`clearState()` is the last statement of the `finally`, standing after a `catch` that is not itself guarded. Nothing reaches it today, and I checked: everything `publish` can reject with — `JwtServicesValidator.sign`, `ZipCodec.encode`, `connection.publish` — throws an `Error` or a `NatsError`, so `error.message` is always safe to read. There is no failing case in the current code.

That is also the only thing holding it up. The one line that must always run sits behind a line that may throw, and its safety is borrowed from three functions in two other packages. The day any of them starts rejecting with something that has no `.message`, the `catch` throws, `clearState()` is skipped, and the worker stays marked busy for the lifetime of the process — the exact failure this change was written to remove, back again, silently, and with no test that would notice.

One line makes the release independent of all of that:

```ts
} finally {
    try {
        await this.publish(WorkerEvents.WORKER_READY);
    } catch (error) {
        this.logger.error(error?.message, [this.workerID, 'WORKER'], userId);
    } finally {
        this.clearState();
    }
}

**Review by ihar-tsykala — CHANGES_REQUESTED** [review]
Hi @vshvets-bc ,

## General comment

The `try / finally` shape is right, and re-publishing `WORKER_READY` from the `finally` rather than only recovering the flag is the part that actually puts the worker back in the pool. One note inline on that `finally`.

The four new comment blocks come to about thirty of the seventy-two added lines, and the nine-line preamble before the outer `try` reproduces #6806, which the description and the commit message already carry. Worth cutting back to the line or two that say why.

**Comment by vshvets-bc** [comment]
Both taken. Pushed in `8652bb55`.

**The release no longer borrows its safety from the catch body.** Your reading is right - nothing fails today, and that is exactly what makes it worth closing: `clearState()` was the one line that must always run, standing behind a line whose safety came from three functions in two other packages. Nested as you suggested:

```ts
} finally {
    try {
        await this.publish(WorkerEvents.WORKER_READY);
    } catch (error) {
        this.logger.error(error?.message, [this.workerID, 'WORKER'], userId);
    } finally {
        this.clearState();
    }
}
```

`error.message` is `error?.message` now too, so the nullish case cannot reach the nested finally in the first place.

**Test added** - `releases even if the WORKER_READY catch itself throws`. It fails on the unguarded shape and passes on the nested one; I checked both directions by reverting the source hunk and rebuilding.

Two things that made the first attempt useless, worth recording since they are easy to repeat:
- an `async` throwing logger stub proves nothing - the catch never awaits it, so it is just an unhandled rejection and `clearState()` still runs. It has to throw synchronously.
- the stub has to be installed *after* `init()`, which logs on its own path.

**Comments cut from about thirty lines to eight.** The nine-line preamble is now two, and the three blocks inside are one or two lines each; the test header went from eleven lines to three. Everything removed is in the description and #6806.

290 passing in worker-service.

ihar-tsykala (approved):

worker-service/src/api/worker.ts
```diff
@@ -187,41 +187,68 @@ export class Worker extends NatsService {
...
+                    try {
+                        await this.publish(completeEvent, {
+                            id: task.id,
+                            error: `Task result could not be delivered: ${error?.message}`
+                        });
+                    } catch (reportError) {
+                        this.logger.error(reportError.message, [this.workerID, 'WORKER'], userId);
+                    }
+                }
+            } catch (error) {
```
Pyatakov, line 236: **A throwing catch body still escapes `runTask` as an unhandled rejection**
The `finally` guarantees the release, but this catch body (and the one at line 247) can themselves throw, and then `runTask` rejects. Both call sites (lines 256 and 269) are fire-and-forget with no `.catch`, so that is an unhandled rejection and the worker process exits. Reproduced against this head: with `publish` rejecting on `WORKER_READY` and a logger whose `error()` throws, `clearState()` runs and then `unhandledRejection: logger unavailable` fires.Added test 5 doesn't catch this only because mocha installs its own `unhandledRejection` listener.

**Suggested fix:**
`runTask(task).catch(...)` at both call sites.

Pyatakov (changes requested):

vshvets-bc:
Right on both counts, including why my test missed it - mocha's own `unhandledRejection` listener meant asserting on worker state could never have caught this. Pushed in `e0a5a2b7`.

`.catch` at both call sites, as suggested:

```ts
runTask(task).catch((error) => this.logTaskFailure(error, task));
```

`logTaskFailure` reports through `console.error` rather than `this.logger`, deliberately: the rejection being handled here is, in the reproduction, a logger that throws, so routing it back through the logger would re-raise the same failure at the last point that can contain it.

**Test added that actually fails without the fix.** It registers its own `unhandledRejection` listener for the duration and asserts nothing arrives - additional listeners still fire alongside mocha's, so the escape is observable:

```
a throwing catch body does not escape as an unhandled rejection
```

Verified in both directions: with the `.catch` removed and rebuilt it is 290 passing / 1 failing; with it, **291 passing**.

worker-service/src/api/worker.ts
```diff
@@ -187,46 +187,75 @@ export class Worker extends NatsService {
...
-
-            await this.publish(WorkerEvents.WORKER_READY);
-            this.isInUse = false;
         }
 
         this.getMessages([this.replySubject, WorkerEvents.SEND_TASK_TO_WORKER].join('.'), async (task) => {
             if (!this.isInUse) {
-                runTask(task);
+                //fire-and-forget: without this a throw from runTask's own error handling
+                //is an unhandled rejection, which takes the process down
```
Pyatakov: Not needed, since `logTaskFailure` itself has its own comment. Please drop

worker-service/tests/worker-runtask-release.test.mjs
```diff
@@ -0,0 +1,173 @@
...
+    HederaSDKHelper.setTransactionResponseCallback = () => {};
+    restorers.push(() => { HederaSDKHelper.setTransactionResponseCallback = origTxCb; });
+
+    const worker = new Worker('w3c-key', 'w3c-proof', 'filebase-key', 'worker-release', fakeLogger);
+    worker.connection = {
+        subscribe() { return { async *[Symbol.asyncIterator]() {}, unsubscribe() {} }; },
+        publish() {},
+    };
+    worker.subscribe = (event, handler) => { captured.subscribe.set(event, handler); return { unsubscribe() {} }; };
+    worker.getMessages = (event, handler) => { captured.getMessages.set(event, handler); return { unsubscribe() {} }; };
```
Pyatakov:
```suggestion
    // isInUse is claimed by the beforeDecode hook, not by runTask, so the stub has to
    // keep it – otherwise the release assertions pass against a flag nobody ever set.
    worker.getMessages = (event, handler, noRespond, beforeDecode, onError) => {
        captured.getMessages.set(event, { handler, beforeDecode, onError });
        return { unsubscribe() {} };
    };
```

worker-service/tests/worker-runtask-release.test.mjs
```diff
@@ -0,0 +1,173 @@
...
+
+    return { worker, captured, restore() { while (restorers.length) { restorers.pop()(); } } };
+}
+
+// Drive one task through the real SEND_TASK_TO_WORKER handler and let the
+// unawaited runTask settle.
+async function dispatch(h, task = { id: 't-1', type: 'GET_FILE', data: {} }) {
+    const sendKey = [...h.captured.getMessages.keys()]
+        .find(k => k.endsWith(WorkerEvents.SEND_TASK_TO_WORKER));
+    await h.captured.getMessages.get(sendKey)(task);
```
Pyatakov:
```suggestion
    const { handler, beforeDecode } = h.captured.getMessages.get(sendKey);
    const early = beforeDecode?.();
    assert.equal(early, undefined, 'the worker must have been free to claim');
    assert.equal(h.worker.isInUse, true, 'the claim must be in place before the task runs');
    await handler(task);
```

vshvets-bc:
Both applied as suggested — and you were right that the suite was proving nothing. Pushed in `01cf6bdb`.

`develop` moved the claim into `claimIfFree` (beforeDecode) while this branch was open. My stub still had the two-argument `getMessages`, so the hook was never captured or called: `isInUse` stayed `false` for the whole test and every `assert.equal(h.worker.isInUse, false)` passed against a flag nobody had set.

The dispatch helper now asserts the claim **before** running the handler, so a harness that stops reflecting the runtime fails loudly instead of going quiet:

```js
const early = beforeDecode?.();
assert.equal(early, undefined, 'the worker must have been free to claim');
assert.equal(h.worker.isInUse, true, 'the claim must be in place before the task runs');
```

**Re-verified both fixes against the real claim**, since the previous runs did not:

| removed | result |
|---|---|
| the nested `finally` | 289 passing, **2 failing** |
| `.catch` at the call sites | 290 passing, **1 failing** |
| neither (branch as pushed) | **291 passing** |

One note on how I got here, in case it saves someone else the trip: `yarn build` in `worker-service` exits 0 while printing `command not found: tsc-alias`, so `tsc && tsc-alias` silently half-runs and the suite happily tests a stale `dist`. The first run after your comments looked green for that reason. The real failure was `TS2554: Expected 2-3 arguments, but got 5` — the new `getMessages` signature needed `common` rebuilt from the merged `develop` first.

Pyatakov (approved):

Issue #6808, opened 2026-09-01
Dynamic role update/delete return 500 "Invalid role" for any non-SR caller


nazlo-climission:
PUT/DELETE `/api/v1/permissions/roles/:id` return 500 `Invalid role` for any caller whose account role is not `STANDARD_REGISTRY`, even when they hold `PERMISSIONS_ROLE_UPDATE`/`DELETE`/`MANAGE`.

**Root cause**

Every `dynamic_role` row is owned by the tenant's STANDARD_REGISTRY DID (`owner.owner`), but `UPDATE_ROLE`/`DELETE_ROLE` in `auth-service/src/api/role-service.ts` scope their DB lookup to `owner.creator` (the caller's own DID) instead. For a non-SR caller these two fields differ, so the scoped lookup always misses and the handler throws `Invalid role`.

`CREATE_ROLE` has the same bug on the write side: it stamps new roles with `owner.creator`, so a role created by a non-SR user is invisible to the SR's own tenant-scoped listings and unmanageable by anyone but its creator.

`GET_ROLE` is unscoped, so the mismatch is never caught as a 404 - it lands in `InternalException` and surfaces as a 500 instead of an authorization error.

**Fix**

- Scope `CREATE_ROLE`/`UPDATE_ROLE`/`DELETE_ROLE` by `owner.owner` instead of `owner.creator`, matching the convention already used by the role-assignment paths.
- Return 403 Forbidden instead of 500 when the ownership check fails, since it's an authorization outcome, not an internal error.
- Update the Swagger examples on both routes to match.

**Files**

- `auth-service/src/api/role-service.ts`
- `api-gateway/src/api/service/permissions.ts`
- `interfaces/src/models/entity-owner.ts` (reference only)

PR #6809, opened 2026-09-01
fix: scope dynamic role create/update/delete by owner.owner, not owner.creator

nazlo-climission:
**Description**:

Fix ownership scoping on dynamic role create/update/delete so any caller other than the STANDARD_REGISTRY can actually manage roles they have permission for.

- Scope CREATE_ROLE/UPDATE_ROLE/DELETE_ROLE by owner.owner (the tenant's SR DID) instead of owner.creator (the caller's own DID), matching the convention already used by the role-assignment paths.
- Return 403 Forbidden instead of a 500 when the ownership check fails, since it's an authorization outcome, not an internal error; updated the Swagger examples to match.
- Fix setDefaultRole's owner filter to use user.parent || user.did instead of the raw user DID, so it doesn't silently depend on the same creator/owner bug.
- Add unit tests covering the ownership scoping for create/update/delete.

**Related issue(s)**:

Fixes #6808

**Notes for reviewer**:

tsc --noEmit clean on both edited files (pre-existing missing-package noise elsewhere is unrelated). 8/8 tests passing; reverting the role-service.ts change alone reproduces 3 of the new test failures, confirming they exercise the regression.

**Checklist**

### DIFF
--- a/auth-service/src/api/role-service.ts
+++ b/auth-service/src/api/role-service.ts
@@ -427,7 +427,7 @@ export class RoleService extends NatsService {
                     const roles = await entityRepository.find(DynamicRole, { id: { $in: userRoles } });
                     for (const role of roles) {
                         if (
-                            (role.owner && role.owner === owner.creator) ||
+                            (role.owner && role.owner === owner.owner) ||
                             (!role.owner && role.default)
                         ) {
                             roleMap.set(role.id, [owner.creator, role.name, role.uuid]);
--- a/auth-service/src/api/role-service.ts
+++ b/auth-service/src/api/role-service.ts
@@ -204,7 +204,7 @@ export class RoleService extends NatsService {
                     const { role, owner, restore } = msg;
                     delete role._id;
                     delete role.id;
-                    role.owner = owner.creator;
+                    role.owner = owner.owner;
                     role.permissions = ListPermissions.unique(role.permissions);
                     role.default = false;
                     role.readonly = false;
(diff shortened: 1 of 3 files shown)


auth-service/src/api/role-service.ts
dariy-rised: `CREATE_ROLE`/`UPDATE_ROLE`/`DELETE_ROLE` now scope by `owner.owner` (the tenant SR's DID) instead of `owner.creator` (the caller's own DID). That's the right convention, it already matches how `GET_ROLES` scopes things, and it's what was causing the 500 for non-SR callers. Checked the actual PR head commit, only these 3 lines change in `role-service.ts`.

But `UPDATE_USER_ROLE` / `DELEGATE_USER_ROLE` were missed

Same file, not touched by this PR. Assigning a role to a user still checks:

```ts
if (
    (role.owner && role.owner === owner.creator) ||
    (!role.owner && role.default)
) { ... } else { throw new Error('Role does not exist'); }
```

`owner.creator` is the caller's own DID, but `role.owner` is now always the SR's DID. So this only ever matches when the caller is the SR.

Fix would be the same change, `role.owner === owner.owner`, in both handlers.

And the new tests only cover create/update/delete in isolation, nothing exercises assigning a role to a user, so this wouldn't get caught.

dariy-rised (changes requested):

dariy-rised (approved):

dariy-rised (approved):
Confirmed: the follow-up fix in `eba7bf83d0` closes the assignment gap. Verified by re-running the scenario against the new code, non-SR create+assign now succeeds, and the pre-existing SR-owned-role assignment path is fixed too. Correction from me: DELEGATE_USER_ROLE didn't need the same change, it gates on the delegator's own held roles, not `role.owner`, so leaving it untouched was correct. Approving.

Issue #6810, opened 2026-09-01
UPLOAD_ARTIFACT does not check parent ownership, so artifacts can be attached to another user's draft policy


vshvets-bc:
## Problem

`UPLOAD_ARTIFACT` (`guardian-service/src/api/artifact.service.ts`) resolves its parent policy or tool **by id alone** and checks only that it is a draft. It never checks that the caller owns it.

```ts
const parent = await getParent(parentId);

if (!parent) {
    return new MessageError('There is no appropriate policy', 422);
}

const category: string = parent.type;
if (parent.type === 'policy') {
    if (parent.item.status !== PolicyStatus.DRAFT) {
        return new MessageError('Policy is not in DRAFT status', 422);
    }
}
```

`getParent` is `DatabaseServer.getPolicyById(parentId)` / `getToolById(parentId)` with no owner predicate. Status is checked; ownership is not.

## Why it matters

The artifact row is written with the **uploader** as `owner`:

```ts
const row = await DatabaseServer.saveArtifact({
    ...
    policyId: parentId,
    owner: owner.owner,
    category
} as any);
```

while `GET_ARTIFACTS` filters the listing by `filter.owner = owner.owner`. So the policy owner listing artifacts for their own policy **does not see the injected one**.

The consumers that assemble the policy package key on `policyId` alone with no owner filter, so the artifact is still pulled into that policy's export, comparison and publish.

The gateway route gates only on `Permissions.ARTIFACTS_FILE_CREATE`.

## This is an omission, not a design

`DELETE_ARTIFACT` in the same file scopes its lookup by owner:

```ts
const artifactToDelete = await DatabaseServer.getArtifact({
    id: artifactId,
    owner: owner.owner
});
```

Upload simply never got the same treatment.

## Suggested fix

Verify the resolved parent belongs to the caller before accepting the upload, covering both the policy and the tool branch.

Worth answering exactly as for a missing parent rather than with a distinct "not yours" message — a different response would confirm that an id exists in someone else's account.

PR to follow.

PR #6811, opened 2026-09-01
fix(guardian-service): check parent ownership before accepting an artifact upload

vshvets-bc:
Fixes #6810.

## Problem

`UPLOAD_ARTIFACT` resolves its parent policy or tool **by id alone** and checks only that it is a draft — never that the caller owns it. So an artifact can be attached to someone else's draft policy.

It is invisible to the victim: the row is written with the **uploader** as `owner`, and `GET_ARTIFACTS` filters the listing on `owner`, so the policy owner never sees it. The consumers that assemble the policy package key on `policyId` alone, so it is still pulled into that policy's export, comparison and publish.

`DELETE_ARTIFACT` in the same service already scopes its lookup by owner, which is what makes this an omission rather than a deliberate design.

## Fix

One ownership check on the resolved parent, covering both the policy and the tool branch.

It reuses the existing `'There is no appropriate policy'` 422 rather than a distinct message. Answering differently for *"exists, but not yours"* would confirm that an id exists in another user's account, which is a disclosure in its own right — so a refusal is indistinguishable from a miss.

The status gate is untouched and still applies to the owner's own non-draft policies.

## Tests

`guardian-service/tests/artifact-upload-owner.test.mjs` — 6 cases on the existing `_handler-harness.mjs`:

- accepts an upload from the policy owner *(control)*
- refuses an upload against another user's draft policy
- does not disclose that the policy exists — refusal is byte-identical to not-found
- refuses an upload against another user's tool
- accepts a tool upload from its owner *(control)*
- still rejects a non-draft policy for its own owner *(control — the status gate survives)*

**Three fail against current `develop`**; the other three are controls that pass either way.

## Note on running the suite

The three new cases pass in isolation and fail as expected without the fix. The full `guardian-service` suite could not be run in my environment — `tests/**` fails to load on `SyntaxError: The requested module '@hiero-ledger/sdk' does not provide an export named 'MirrorNodeAccountBalanceQuery'`. I confirmed that error reproduces on a clean checkout with my changes stashed, so it is pre-existing and unrelated to this branch, but it does mean I have not exercised the whole suite locally.

### DIFF
--- a/guardian-service/src/api/artifact.service.ts
+++ b/guardian-service/src/api/artifact.service.ts
@@ -55,6 +55,21 @@ export async function artifactAPI(logger: PinoLogger): Promise<void> {
                 return new MessageError('There is no appropriate policy', 422);
             }
 
+            /*
+             * The parent was resolved by id alone, so verify the caller owns it.
+             * Without this an artifact can be attached to another user's draft:
+             * the row is written with the uploader as `owner`, so it is hidden
+             * from the policy owner's listing (GET_ARTIFACTS filters on owner)
+             * while export, comparison and publish key on `policyId` alone and
+             * pull it in. DELETE_ARTIFACT already scopes its lookup by owner.
+             *
+             * Answers exactly as for a missing parent: telling the caller the id
+             * exists but belongs to someone else is itself a disclosure.
+             */
+            if (parent.item?.owner !== owner.owner) {
+                return new MessageError('There is no appropriate policy', 422);
+            }
+
             const category: string = parent.type;
             if (parent.type === 'policy') {
                 if (parent.item.status !== PolicyStatus.DRAFT) {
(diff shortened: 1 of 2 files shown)


ihar-tsykala (changes requested):
Hi @vshvets-bc ,

## General comment

The ownership check is in the right place — before both the policy and the tool branch, and before anything is written — and the new tests do fail without it.

Nothing behaves wrongly. Could you trim the block comment above the check, and the header of `artifact-upload-owner.test.mjs`, to the line or two saying why the refusal reuses the not-found message — the rest is already in the pull request description and the commit message.

ihar-tsykala (approved):

vshvets-bc:
Trimmed. Pushed in `f44f2afc`.

The block above the check is now the one line that is not evident from the code - why the refusal reuses the not-found message:

```ts
// answers exactly as for a missing parent: saying the id exists but belongs
// to someone else is itself a disclosure
```

The `artifact-upload-owner.test.mjs` header went from eleven lines to three, keeping the same point. The rest - the `owner` column on the row, `GET_ARTIFACTS` filtering the listing, export/comparison/publish keying on `policyId`, and `DELETE_ARTIFACT` already scoping by owner - is in the pull request description and #6810.

No behaviour change: 6 passing, unchanged.

Issue #6817, opened 2026-09-02
Profile restore invalidates the cache on task dispatch instead of completion, so the profile stays stale for up to 10 minutes


vshvets-bc:
### Summary

`PUT /profiles/restore/:username` and `PUT /profiles/restore/topics/:username` clear the profile cache when the background task is **accepted**, not when it **finishes**. A `GET /profiles/:username` served while the restore is still running re-fills the cache with the pre-restore profile, and the user keeps seeing it for up to the 600 s TTL.

### Detail

`api-gateway/src/api/service/profile.ts`, `restoreUserProfile`:

```ts
RunFunctionAsync<ServiceError>(async () => {
    const guardians = new Guardians();
    await guardians.restoreUserProfileCommonAsync(user, username, profile, task);

    await this.cacheService.invalidate(getCacheKey([req.url, ...invalidedCacheTags], user))
    await this.invalidateAccountsCache(user.id);
}, ...);
```

`restoreUserProfileCommonAsync` is a single `sendMessage` (`api-gateway/src/helpers/guardians.ts`):

```ts
return await this.sendMessage(MessageAPI.RESTORE_USER_PROFILE_COMMON_ASYNC, { user, username, profile, task });
```

It resolves when guardian-service **accepts** the task and returns the `NewTask` — that is the point of the endpoint answering `202` with a task id to poll. So the `await` buys nothing here: the invalidation runs milliseconds after dispatch, while the restore is still resolving the DID and scanning topics.

`restoreTopic` (`PUT /profiles/restore/topics/:username`) has the same shape.

### Why this looks like an oversight rather than a decision

The sibling endpoint in the same file already does it correctly. `setUserProfileAsync` (`PUT /profiles/push/:username`) registers a completion callback:

```ts
taskManager.registerCallback(task, async () => {
    await this.cacheService.invalidate(
        getCacheKey([req.url, ...invalidedCacheTags], user)
    );
    await this.invalidateAccountsCache(user.id);
});
```

Three adjacent async profile flows: one invalidates on completion, two on dispatch.

### Steps to reproduce

1. Start a profile restore. The endpoint returns a task id and the cache entry is cleared immediately.
2. While the task runs, load anything that issues `GET /profiles/:username` (a page load or component refresh). The cache is repopulated with the profile as it still is — un-restored.
3. `GET /profiles/:username` is `@UseCache()` with the default 600 s TTL (`api-gateway/src/constants/cache.ts`).
[...]

PR #6818, opened 2026-09-02
fix(api-gateway): invalidate the profile cache when a restore completes, not when it is dispatched

vshvets-bc:
Fixes #6817.

## Problem

`restoreUserProfile` and `restoreTopic` cleared the profile cache inside `RunFunctionAsync`, right after `restoreUserProfileCommonAsync` / `getAllUserTopicsAsync` resolved:

```ts
RunFunctionAsync<ServiceError>(async () => {
    const guardians = new Guardians();
    await guardians.restoreUserProfileCommonAsync(user, username, profile, task);

    await this.cacheService.invalidate(getCacheKey([req.url, ...invalidedCacheTags], user))
    await this.invalidateAccountsCache(user.id);
}, ...);
```

Those helpers are a single `sendMessage` that resolves when guardian-service **accepts** the task and returns the `NewTask` — which is why these endpoints answer `202` with a task id to poll. So the `await` buys nothing: the invalidation ran milliseconds after dispatch, while the restore was still resolving the DID and scanning topics.

Any `GET /profiles/:username` served in that window repopulated the cache with the **pre-restore** profile, and that route is `@UseCache()` with the default 600 s TTL — so the user kept seeing the old profile for up to ten minutes after the task reported success.

## Fix

Both handlers now register a completion callback, which is what `setUserProfileAsync` in the same file already does:

```ts
taskManager.registerCallback(task, async () => {
    await this.cacheService.invalidate(getCacheKey([req.url, ...invalidedCacheTags], user));
    await this.invalidateAccountsCache(user.id);
});
```

`TaskManager.registerCallback` and its completion-time invocation already exist, including the guard against a task completing twice, so this adds no new machinery — it just brings two of three adjacent async profile flows in line with the third.

The error arm is deliberately unchanged: a failed **dispatch** is known immediately, so invalidating there is still correct.

## Note for reviewers

`setUserProfileAsync` invalidates two tags — `profiles/:username` **and** `accounts/session` — while both restore handlers invalidate only `profiles/:username` (`restoreUserProfile` additionally calls `invalidateAccountsCache`, `restoreTopic` does neither). Whether restore should also clear the session tag is a separate question, so this PR keeps the existing tag sets and only changes *when* they are cleared. Happy to widen it if you would prefer.

### DIFF
--- a/api-gateway/src/api/service/profile.ts
+++ b/api-gateway/src/api/service/profile.ts
@@ -364,12 +364,15 @@ export class ProfileApi {
         const username: string = user.username;
 
         const invalidedCacheTags = [`/${PREFIXES.PROFILES}/${username}`];
+
+        taskManager.registerCallback(task, async () => {
+            await this.cacheService.invalidate(getCacheKey([req.url, ...invalidedCacheTags], user));
+            await this.invalidateAccountsCache(user.id);
+        });
+
         RunFunctionAsync<ServiceError>(async () => {
             const guardians = new Guardians();
             await guardians.restoreUserProfileCommonAsync(user, username, profile, task);
-
-            await this.cacheService.invalidate(getCacheKey([req.url, ...invalidedCacheTags], user))
-            await this.invalidateAccountsCache(user.id);
         }, async (error) => {
             await this.logger.error(error, ['API_GATEWAY'], user.id);
             taskManager.addError(task.taskId, { code: error.code || 500, message: error.message });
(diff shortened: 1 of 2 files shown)


api-gateway/src/api/service/profile.ts
Pyatakov: `addError` calls `handleCallback`, which runs the callback registered on line 368, so the `cacheService.invalidate` here  invalidates the same key a second time on failure. Dead code now; drop it.

Pyatakov: Same leftover – this line duplicates the callback registered on line 447.

Pyatakov (changes requested):

vshvets-bc:
Both dropped. Pushed in `2e663359`.

You're right that `addError` reaches `handleCallback` (`task-manager.ts:328`), so the callback registered a few lines above already invalidates on the failure path - and `handleCallback` deletes the callback before running it, so it cannot fire twice either way.

Added `tests/task-manager-completion-callback.test.js` while removing the lines, since the deletion now rests entirely on that behaviour:
- the callback runs when the task **fails**, not only when it succeeds
- it runs **once** when a result and an error both arrive

2 passing.

Pyatakov (approved):

PR #6832, opened 2026-09-04
feat: 6594 table field columns configuration

ihar-tsykala:
**Description**:

A Table field could be added to a schema but could not be described there. Its columns appeared only after somebody uploaded a spreadsheet, and the file's first row silently became the column keys. A schema author therefore could not write a stable formula, script, or API integration against those columns before a document existed.

The field panel now provides a Define columns switch. Each declared column has a display name and a stable key generated from that name, and columns can be added, removed down to one, and reordered. The configuration survives schema save and reload, controls the document form, and is carried with the stored table reference. AutoCalculate, Custom Logic, and Math Advanced JavaScript use the same table helpers and declared keys, including the Math editor's Testing flow. The Indexer displays the same column names as Guardian.

The documents endpoint can expand table values on request and apply row paging and column filtering without changing stored documents. Tables without declared columns keep their legacy behavior and use the uploaded file's first row as the header. Column data types, per-column validation and defaults, and the General formula builder remain unchanged.

Fixes #6594

**Notes for reviewer**:

**Checklist**


frontend/src/app/modules/policy-engine/dialogs/math-editor-dialog/math-model/table-hydration.ts
```diff
@@ -0,0 +1,162 @@
...
+        );
+    }
+
+    const blob = await firstValueFrom(deps.artifactService.getFileBlob(fileId));
+    return await deps.gzipService.gunzipToText(blob);
+}
+
+async function loadRows(table: any, deps: ITableHydrationDeps): Promise<any> {
+    const csvText = await readCsvText(table, deps);
+    const declaredKeys = Array.isArray(table.columnKeys) && table.columnKeys.length > 0
```
Pyatakov: `loadRows` passes `table.columnKeys` as declared keys for *any* table that has them, including legacy tables whose keys are the auto-generated `C1..Cn`. `parseCsvToTable` then strips row 0 as a header, and since `isDeclared()` is false the caller runs `reKeyLegacyTable`, which strips a second row and uses its values as keys. The pipeline was ran with the repo's papaparse plus the new csv.service logic: `Year,CO2 / 2021,10 / 2022,20 / 2023,30` comes back as `columnKeys: ['2021','10']` with two rows. The formula gets tested against silently wrong data.

**Suggested fix:**
`const declaredKeys = isDeclared(table) ? table.columnKeys : undefined;`

policy-service/src/policy-engine/helpers/table-field.ts
```diff
@@ -213,42 +62,14 @@ function defineHidden<T extends object, K extends string>(
...
-
 /**
  * Recursively traverses any object/array and hydrates all table occurrences.
  */
 export async function hydrateTablesInObject(
     root: unknown,
     loadFileText: TableFileLoader,
-    delimiter: string = ','
+    delimiter: string = ',',
+    shouldHydrate?: TableHydrationFilter
```
Pyatakov: Added with no caller anywhere

api-gateway/src/api/service/policy.ts
```diff
@@ -2824,6 +2856,12 @@ export class PolicyApi {
                 type,
                 pageIndex,
                 pageSize,
+                {
+                    expand: String(expandTables)?.toLowerCase() === 'true',
```
Pyatakov: `String(...)?.toLowerCase()`; `String()` is never nullish.

frontend/src/app/services/csv.service.ts
```diff
@@ -6,32 +6,47 @@ export interface TableData {
...
+        storedColumnKeys?: string[]
+    ): TableData {
+        const storedKeys: string[] = Array.isArray(storedColumnKeys) ? storedColumnKeys : [];
+        const hasStoredKeys: boolean = storedKeys.length > 0;
+
         const parsed = Papa.parse<string[]>(csvText, {
             header: false,
             delimiter,
-            skipEmptyLines: true,
+            skipEmptyLines: !hasStoredKeys,
```
Pyatakov, line 34: Verified with papaparse: `"a,b\r\n1,2\r\n"` parses to `[['a','b'],['1','2'],['']]`, so any declared-column CSV ending in a newline gains an all-empty trailing row in grid, preview and re-serialised file. Dropping the header doesn't need empty lines kept – leave it `true`.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1740,6 +1760,197 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
+    }
+
+    public setTableColumnValue(index: number, key: 'name' | 'key', value: string): void {
+        const columns = this.selectedField?.tableColumns;
+        if (!Array.isArray(columns) || index < 0 || index >= columns.length) { return; }
+        const column = columns[index];
+        const derivedKey = key === 'name' && !this.isTableColumnKeyUnlocked(column);
+        column[key] = value;
+        if (derivedKey) {
+            column.key = SchemasConfigurationComponent.toColumnKey(value);
```
Pyatakov, line 1819:
```typescript
return !!column.key && column.key !== SchemasConfigurationComponent.toColumnKey(column.name);
```

The line above returns `false` (key is "locked") whenever the stored `key` matches the derived slug of the `name`. Since editor-created columns follow this pattern by design, renaming them silently rewrites the key:
```typescript
// setTableColumnValue (line 1816-1819):
const derivedKey = key === 'name' && !this.isTableColumnKeyUnlocked(column);
if (derivedKey) {
    column.key = SchemasConfigurationComponent.toColumnKey(value);  // Silent rewrite
}
```
When editing an existing schema where `key='year'` and `name='Year'`, `isTableColumnKeyUnlocked` returns `false`, so changing the name to `'Reporting year'` silently rewrites the key to `'reporting_year'`.

Pyatakov (changes requested):

frontend/src/app/modules/policy-engine/dialogs/math-editor-dialog/math-model/table-hydration.ts
ihar-tsykala: It was fixed. Could you check. Thanks!

policy-service/src/policy-engine/helpers/table-field.ts
ihar-tsykala: It was fixed. Could you check. Thanks!

api-gateway/src/api/service/policy.ts
ihar-tsykala: It was fixed. Could you check. Thanks!

frontend/src/app/services/csv.service.ts
ihar-tsykala: It was fixed. Could you check. Empty rows are data — dropping them broke single-column tables. Only the newline artefact is dropped now. Thanks!

frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
ihar-tsykala: This works as expected.

The key follows the column name only while the user lets it. Pressing the edit (pencil) button turns that off, and from then on the key stays as it is, whatever the name becomes. The same button also lets the user change only the key and leave the name alone.

So the user has full control over keys and over the formulas built on them.

A published schema cannot be edited, so the only thing that can break here is the author's own draft.

Pyatakov (approved):

Issue #6835, opened 2026-09-04
Policy can be started twice and hosted by two policy-service processes, causing intermittent "Block Unavailable"


vshvets-bc:
### Problem description

A policy can end up **running on two policy-service processes at once**. Both join the same NATS queue group for that policy, so block requests are split between two instances that hold divergent in-memory block state. Roughly half of them fail with `Block Unavailable` (503/422), and the UI renders forms and lists empty at random. Refreshing often "fixes" it, because the refresh happens to land on the other instance.

There are two independent ways to get there.

**1. `guardian-service` can issue two `GENERATE_POLICY` messages for one policy.**

`PolicyEngine.generateModel` decides whether to start a policy from `checkIfPolicyAlive`, a short NATS ping:

```ts
// guardian-service/src/policy-engine/policy-engine.ts
const exist = await new GuardiansService().checkIfPolicyAlive(policyId);
if (!exist) {
    const r = await this.sendMessageWithTimeout<any>(PolicyEvents.GENERATE_POLICY, 1000, { ... });
```

A policy takes tens of seconds from fork to answering that ping. Any second caller arriving inside that window sees "not alive" and sends its own `GENERATE_POLICY`, which a **different** pod accepts. Nothing serialises the callers, and every start path goes through this method: startup, dry-run, publish, `regenerateModel`, and `restore-data-from-hedera`.

Two things make the window easy to hit:

* startup starts everything at once — `await Promise.all(policies.map(async (policy) => { await this.generateModel(...) }))`, so a user action during startup races it;
* the capacity retry is unbounded — `await new Promise(resolve => setTimeout(resolve, 10000)); return this.generateModel(policyId, enableMock);` — so two callers that both find the fleet full will keep retrying until each lands a confirm, possibly on different pods.

**2. `policy-service` can fork a second child for a policy it already hosts.**

```ts
// policy-service/src/helpers/policy-container.ts
public addPolicy(config: IPolicyStartOptions): boolean {
    if (this.processCount >= this.maxPolicyInstances) { ... return false }
    this.container.set(config.policyId, { options: config, process: null });
    return true;
}
```

There is no check for a policy this pod already hosts. `runPolicyProcess` guards with `if (instance.process) { return; }`, but `addPolicy` has just replaced the entry with a fresh one whose `process` is `null`, so the guard passes and a **second child is forked**. The handle to the first is dropped while that process keeps running and stays subscribed to the policy's subjects — it can no longer be stopped, and it is not counted.

[...]

PR #6836, opened 2026-09-04
fix: 6835 prevent a policy from being started twice and hosted by two processes

vshvets-bc:
Fixes #6835

A policy could end up running on two policy-service processes at once. Both join the same NATS queue group for that policy, so block requests are split between two instances holding divergent in-memory state and roughly half fail with `Block Unavailable`. This closes both routes into that state.

### 1. `guardian-service`: serialise starts of the same policy

`generateModel` decides whether to start a policy from `checkIfPolicyAlive`, a short ping, while a policy takes tens of seconds from fork to answering it. A second caller inside that window saw "not alive" and sent its own `GENERATE_POLICY`, which a different pod accepted.

Every start path funnels through `generateModel` — startup, dry-run, publish, `regenerateModel`, `restore-data-from-hedera` — so the guard belongs there. It now keeps an `inFlightStarts` map and hands a concurrent caller the attempt already running. The entry is released when that attempt settles, on success and on failure alike, so it is not a cache and a failed start does not wedge it shut.

The body moved to a private `startModel`. The capacity retry re-enters `startModel` rather than `generateModel` — going through the guarded entry point would make it join itself and never settle.

### 2. `policy-service`: do not re-add a policy the pod already hosts

`addPolicy` overwrote an existing container entry with a fresh `{ options, process: null }`. `runPolicyProcess` only skips an instance that already has a `process`, so replacing the entry defeated that guard: a repeated `GENERATE_POLICY` forked a second child and dropped the handle to the first, which kept running and stayed subscribed — it could no longer be stopped, and it was not counted.

It now returns early when the policy is already in the container. That check runs *before* the capacity test, because re-adding a hosted policy consumes no slot and must not be refused as if the pod were full.

### Tests

Nine new unit tests, all failing before the change and passing after.

`guardian-service/tests/unit/policy-engine-generate-dedupe.test.mjs`
* concurrent starts of one policy produce a single `GENERATE_POLICY`, and both callers get the same result
* different policies are not deduped against each other
* the guard is released once a start settles, so a later start is still allowed
* the guard is released when a start fails, so the next attempt retries
* a caller that joins an in-flight start receives its failure too

`policy-service/tests/unit-tests/helpers/policy-container-add-policy.test.mjs`
* a policy the pod is not hosting is accepted
* a live child handle survives a repeated `GENERATE_POLICY`
* re-adding a hosted policy does not consume a slot, so it is not refused when the pod is full
* a genuinely new policy is still refused once the pod is full

### Verification

Run against `develop` (`4063d4725`):

| suite | baseline | with this PR |
|---|---|---|
| `guardian-service` `tests/unit` | 1822 passing, 11 failing | 1827 passing, 11 failing |
| `policy-service` `tests/unit-tests` | 3263 passing, 0 failing | 3267 passing, 0 failing |

The 11 `guardian-service` failures are present on unmodified `develop` and are all schema-template / package-publish tests, unrelated to policy startup. `policy-service` has an intermittent 2s-timeout flake in `custom-logic-worker-error.test.mjs` that fires on some runs on `develop` too.

`tslint` is clean on both changed files.

### DIFF
--- a/guardian-service/src/policy-engine/policy-engine.ts
+++ b/guardian-service/src/policy-engine/policy-engine.ts
@@ -238,6 +238,12 @@ export class PolicyEngine extends NatsService {
     private readonly policyReadyCallbacks: Map<string, (data: any, error?: any) => void> = new Map();
 
     /**
+     * Starts currently in flight, keyed by policy id
+     * @private
+     */
+    private readonly inFlightStarts: Map<string, Promise<any>> = new Map();
+
+    /**
      * Policy initialization errors container
      * @private
      */
@@ -2453,6 +2459,28 @@ export class PolicyEngine extends NatsService {
      * @param policyId
      */
     public async generateModel(policyId: string, enableMock: boolean): Promise<any> {
+        const inFlight = this.inFlightStarts.get(policyId);
+        if (inFlight) {
+            return inFlight;
+        }
+
+        const start = this.startModel(policyId, enableMock)
+            // Released before the caller settles, so a `.then` that starts the policy
+            // again cannot observe a stale entry. `finally` runs too late for that.
+            .then(
+                (result) => { this.inFlightStarts.delete(policyId); return result; },
+                (error) => { this.inFlightStarts.delete(policyId); throw error; }
+            );
+        this.inFlightStarts.set(policyId, start);
+        return start;
+    }
+
+    /**
+     * Start the policy process. Callers go through generateModel, which serialises
+     * concurrent starts of the same policy.
+     * @param policyId
+     */
+    private async startModel(policyId: string, enableMock: boolean): Promise<any> {
         const policy = await DatabaseServer.getPolicyById(policyId);
         if (!policy || (typeof policy !== 'object')) {
             throw new Error('Policy was not exist');
@@ -2493,7 +2521,7 @@ export class PolicyEngine extends NatsService {
             } else {
                 await new Promise(resolve => setTimeout(resolve, 10000));
 
-                return this.generateModel(policyId, enableMock);
+                return this.startModel(policyId, enableMock);
             }
         } else {
             return Promise.resolve();
--- a/guardian-service/src/policy-engine/policy-engine.ts
+++ b/guardian-service/src/policy-engine/policy-engine.ts
@@ -244,6 +244,12 @@ export class PolicyEngine extends NatsService {
     private readonly inFlightStarts: Map<string, Promise<any>> = new Map();
 
     /**
+     * Ceiling on one start attempt, after which the entry is released so a later
+     * caller can retry rather than await a promise that will never settle.
+     */
+    private static readonly START_TIMEOUT_MS = 15 * 60 * 1000;
+
+    /**
      * Policy initialization errors container
      * @private
      */
@@ -2450,8 +2456,19 @@ export class PolicyEngine extends NatsService {
      * @param policyOwnerId
      */
     public async destroyModel(policyId: string, policyOwnerId: string | null): Promise<void> {
+        //a start in flight can no longer complete: the child exits with code 0 and no
+        //ready event follows, so the entry has to go with it
+        this.clearInFlightStart(policyId);
         PolicyServiceChannelsContainer.deletePolicyServiceChannel(policyId);
         new GuardiansService().sendPolicyMessage(PolicyEvents.DELETE_POLICY, policyId, { policyOwnerId });
+    }
+
+    /**
+     * Drop a recorded start attempt, so the next caller starts afresh.
+     * @param policyId
+     */
+    private clearInFlightStart(policyId: string): void {
+        this.inFlightStarts.delete(policyId);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


guardian-service/src/policy-engine/policy-engine.ts
```diff
@@ -2453,6 +2459,28 @@ export class PolicyEngine extends NatsService {
...
+        }
+
+        const start = this.startModel(policyId, enableMock)
+            // Released before the caller settles, so a `.then` that starts the policy
+            // again cannot observe a stale entry. `finally` runs too late for that.
+            .then(
+                (result) => { this.inFlightStarts.delete(policyId); return result; },
+                (error) => { this.inFlightStarts.delete(policyId); throw error; }
+            );
+        this.inFlightStarts.set(policyId, start);
```
Pyatakov, line 2504: The `inFlightStarts` entry is only cleared by the two handlers attached to the start promise, and that promise can never settle in some real paths: `startModel`'s confirmed branch waits on `policyReadyCallbacks` with no timeout, and POLICY_READY never arrives when
- policy-service gives up after `maxRestartAttempts` – `scheduleRestart` deletes the container entry and emits nothing (`policy-service/src/helpers/policy-container.ts:550`), or
- `destroyModel` runs while a start is in flight – the child exits with code 0, the entry is deleted, no ready event follows.

After that, every later `generateModel` for that policy returns the dead promise and hangs. `PolicyEngine` is `@Singleton`, so the map is process-wide. Before this change the first caller hung but a *second* caller re-ran the whole start and recovered; this PR removes that recovery, so the policy is unstartable until guardian-service restarts.

**Suggested fix:**
Bound the stored attempt (race it against a rejecting timeout that clears the entry) and clear `inFlightStarts` in `destroyModel`/`destroyModelForMigration`.

policy-service/src/helpers/policy-container.ts
```diff
@@ -283,6 +283,14 @@ export class PolicyContainer extends NatsService {
      * @param config
      */
     public addPolicy(config: IPolicyStartOptions): boolean {
+        // Already hosted here: keep the existing entry. Replacing it would hand
+        // runPolicyProcess a `process: null` instance and defeat its guard, forking a
+        // second child while the first keeps running and stays subscribed. Checked
+        // before the capacity test because re-adding a hosted policy costs no slot.
+        if (this.container.has(config.policyId)) {
```
Pyatakov: Keeping the existing entry is right, but the incoming `config` is discarded while the reply is still `confirmed: true`. A GENERATE_POLICY asking for different `enableMock`/`skipRegistration` on an already-hosted policy is served by the child forked with the old options, and guardian-service parks on POLICY_READY as if its request were honored.
**Suggest**
Logging the mismatch and returning `false` when `config` differs from `instance.options`, so the request routes to a pod that will honor it.

policy-service/src/helpers/policy-container.ts
```diff
@@ -283,6 +283,14 @@ export class PolicyContainer extends NatsService {
      * @param config
      */
     public addPolicy(config: IPolicyStartOptions): boolean {
+        // Already hosted here: keep the existing entry. Replacing it would hand
+        // runPolicyProcess a `process: null` instance and defeat its guard, forking a
+        // second child while the first keeps running and stays subscribed. Checked
+        // before the capacity test because re-adding a hosted policy costs no slot.
```
Pyatakov: Keep only relevant information without unnecessary verbose

Pyatakov (changes requested):

vshvets-bc:
Both are real, and the first is the more serious - you are right that this PR removed the accidental recovery that made the old behaviour survivable. Pushed in `de7a7918`.

**1. The stored attempt is bounded now.** `generateModel` races `startModel` against a timeout that rejects and clears the entry, and `destroyModel` / `destroyModelForMigration` clear it too, for exactly the path you describe: the child exits 0 and no ready event follows.

```ts
const start = Promise.race([this.startModel(policyId, enableMock), bound]).then(
    (result) => { clearTimeout(expiry); this.clearInFlightStart(policyId); return result; },
    (error)  => { clearTimeout(expiry); this.clearInFlightStart(policyId); throw error; }
);
```

`START_TIMEOUT_MS` is 15 minutes - well past a legitimate cold start, so it only ever fires on an attempt that cannot complete. A caller that hits it gets a rejection naming the policy rather than hanging, and the next caller starts afresh.

**2. A mismatched request is declined.** `addPolicy` compares `enableMock` / `skipRegistration` against the options the child was actually forked with, logs the mismatch and returns `false`, so the request routes to a pod that will honour it instead of being served by the wrong process while the caller waits on POLICY_READY.

**Tests:**
- `policy-engine-generate-dedupe.test.mjs` — **7 passing**, adding `a start that never settles is released, so a later caller can retry` and `destroying a policy drops a start still in flight`
- `policy-container-add-policy.test.mjs` — **6 passing**, adding `declines a request whose start options differ from the running child` and a companion that a matching repeat is still claimed

Comments compacted in both files.

Pyatakov (approved):

Issue #6849, opened 2026-09-07
sendToGuardianBlock resolves a client-supplied draftId without an owner check, so one participant can overwrite and take over another's draft


vshvets-bc:
### Summary

`sendToGuardianBlock` resolves a draft by an id taken straight from the block payload, scoped by policy but not by owner, and then copies every field of the incoming document onto that row — including `owner`. One participant in a policy can overwrite another participant's draft and become its owner.

### Detail

The id is client supplied. `request-vc-document-block.ts`:

```ts
const draft = data.draft;
const draftId = data.draftId;
```

`data` is the unmodified block payload, and it is stamped onto the outgoing document:

```ts
item.draft = draft;
item.draftId = draftId;
```

`send-to-guardian-block.ts`, `getVCRecord()` then resolves it:

```ts
if (document.draft || document.draftId) {
    old = await ref.databaseServer.getVcDocument({
        id: { $eq: document?.draftId },
        policyId: { $eq: ref.policyId },
        draft: { $eq: true }
    });
}
```

Scoped by `policyId` and `draft`. There is no `owner` predicate.

`updateVCRecord()` merges and persists what it finds:

```ts
old = this.mapDocument(old, document);
old = await PolicyUtils.updateVC(ref, old, userId);
```

and `mapDocument()` copies everything except the identifiers:

```ts
private mapDocument(old: IPolicyDocument, doc: IPolicyDocument): IPolicyDocument {
    for (const key in doc) {
        if (key !== 'id' && key !== '_id' && typeof doc[key] !== 'function') {
            old[key] = doc[key];
        }
    }
    return old;
}
```

`owner` is not excluded, so the found row's owner is replaced by the submitter's.

### Steps to reproduce

1. Users A and B are both participants in policy P. B has a draft VC.
2. A obtains B's draft row id. Ids reach clients as strings — `documents-source-addon.ts` projects `$toString: '$_id'`, and any grid not configured `onlyOwnDocuments` lists other participants' drafts.
3. A submits an ordinary VC through the request block with `draftId` set to B's row.
4. The send block finds B's draft (same policy, `draft: true`), overwrites its content with A's, and sets `owner` to A.

B's draft is destroyed and the row now belongs to A. No error is raised and nothing in the record distinguishes it from an ordinary edit.

### Expected

A draft can only be resumed by the user who created it, and no incoming document can reassign an existing row's owner.

### Note

[...]

PR #6850, opened 2026-09-07
fix(policy-service): scope the draft lookup to its owner

vshvets-bc:
Fixes #6849.

## Problem

`getVCRecord()` resolved a draft by an id taken **straight from the block payload**, scoped by `policyId` and `draft: true` but not by owner:

```ts
old = await ref.databaseServer.getVcDocument({
    id: { $eq: document?.draftId },
    policyId: { $eq: ref.policyId },
    draft: { $eq: true }
});
```

`mapDocument()` then copies every field of the incoming document onto the row it finds, excluding only `id` and `_id` — so **`owner` was copied too**.

A participant in a policy could pass another participant's `draftId`, overwrite that draft's content and become its owner, with no error raised and nothing in the record distinguishing it from an ordinary edit. Draft ids are not secret: `documents-source-addon.ts` projects them to clients as `$toString: '$_id'`, and any grid not configured `onlyOwnDocuments` lists other participants' drafts.

## Fix

Two lines, two layers.

**The lookup is scoped to the submitter.** `document.owner` is the submitter's DID — `PolicyUtils.createVC` sets `owner: owner.did` from the authenticated user — so this restricts the row to whoever is submitting.

**`mapDocument` no longer copies `owner`.** No incoming document should reassign an existing row's owner, whichever branch found it. This also covers the `hash` branch, which is scoped by policy and hash alone and can likewise land on another user's row.

A miss falls through to the existing insert path, storing the submission as the submitter's own document — so a stale or unknown `draftId` costs the user nothing.

## Note on testing

I did not add tests here, because `policy-service` has no existing spec that exercises this block's runtime path in this repository — the nearest, `send-to-guardian-block.test.mjs`, covers only the validator. Happy to add coverage in whatever shape you prefer.

The equivalent change is covered downstream by four specs that assert the **query filter** rather than the outcome — worth mentioning because the old code behaves correctly for the legitimate case, so a test that merely resumes one's own draft passes against the defect. Against the unpatched source, 3 of those 4 fail; the one that passes is the deliberate "the owner can still resume their own draft" guard.

Verified `tsc --noEmit` passes for `policy-service` with this change.

### DIFF
--- a/policy-service/src/policy-engine/blocks/send-to-guardian-block.ts
+++ b/policy-service/src/policy-engine/blocks/send-to-guardian-block.ts
@@ -61,7 +61,7 @@ export class SendToGuardianBlock {
      */
     private mapDocument(old: IPolicyDocument, doc: IPolicyDocument): IPolicyDocument {
         for (const key in doc) {
-            if (key !== 'id' && key !== '_id' && typeof doc[key] !== 'function') {
+            if (key !== 'id' && key !== '_id' && key !== 'owner' && typeof doc[key] !== 'function') {
                 old[key] = doc[key];
             }
         }
@@ -80,6 +80,7 @@ export class SendToGuardianBlock {
             old = await ref.databaseServer.getVcDocument({
                 id: { $eq: document?.draftId },
                 policyId: { $eq: ref.policyId },
+                owner: { $eq: document?.owner },
                 draft: { $eq: true }
             });
         }
--- a/policy-service/src/policy-engine/blocks/send-to-guardian-block.ts
+++ b/policy-service/src/policy-engine/blocks/send-to-guardian-block.ts
@@ -61,7 +61,7 @@ export class SendToGuardianBlock {
      */
     private mapDocument(old: IPolicyDocument, doc: IPolicyDocument): IPolicyDocument {
         for (const key in doc) {
-            if (key !== 'id' && key !== '_id' && key !== 'owner' && typeof doc[key] !== 'function') {
+            if (key !== 'id' && key !== '_id' && typeof doc[key] !== 'function') {
                 old[key] = doc[key];
             }
         }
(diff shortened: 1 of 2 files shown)


policy-service/src/policy-engine/blocks/send-to-guardian-block.ts
```diff
@@ -61,7 +61,7 @@ export class SendToGuardianBlock {
      */
     private mapDocument(old: IPolicyDocument, doc: IPolicyDocument): IPolicyDocument {
         for (const key in doc) {
-            if (key !== 'id' && key !== '_id' && typeof doc[key] !== 'function') {
+            if (key !== 'id' && key !== '_id' && key !== 'owner' && typeof doc[key] !== 'function') {
```
Pyatakov: Excluding `owner` from `mapDocument` breaks `setRelationshipsBlock`'s `changeOwner` option. That block assigns `doc.owner = owner` (`policy-engine/blocks/set-relationships-block.ts:100, :125`) and feeds the document into `sendToGuardianBlock`. Reassigning the owner doesn't change the credential content, so `toCredentialHash()` is unchanged and `getVCRecord`'s hash branch still finds the existing record; `updateVCRecord` goes through `mapDocument`, which now drops the new `owner`, and returns `old`, so the reassignment is neither persisted nor propagated downstream. `mapDocument` is also shared by the approval, DID and VP paths, making the change far wider than the draft bug.

The new owner filter already guarantees a draft belongs to the same owner, reverting this line is the suggested resolution.

Pyatakov (changes requested):

vshvets-bc:
Reverted, as suggested. Pushed in `0ed9b21c`.

Your reasoning holds all the way through: `toCredentialHash()` is unchanged by a reassignment, `getVCRecord`'s hash branch still finds the existing record, and `mapDocument` dropping `owner` would have made `updateVCRecord` return `old` - so `setRelationshipsBlock`'s `changeOwner` (`:100`, `:125`) would have been neither persisted nor propagated. And `mapDocument` is shared by the approval, DID and VP paths, so the blast radius was far wider than the draft bug.

The owner filter on the draft lookup is enough on its own, and it is the narrower change.

Added `send-to-guardian-draft-owner.test.mjs` (4 passing), including two that pin exactly what the revert restores:
- `asks for the draft by owner as well as id and policy` - fails without the filter
- `leaves the hash path alone`
- `mapDocument still carries a reassigned owner` - the `changeOwner` case
- `mapDocument still refuses to copy the identity fields`

Pyatakov (approved):

Issue #6855, opened 2026-09-07
Excel formula range expansion is unbounded: a crafted .xlsx aborts guardian-service on the preview path


vshvets-bc:
### Summary

`Expression.parseRange` expands a cell range into one array entry per row, with the row numbers taken from the formula text of the uploaded `.xlsx` and no bound of any kind. A crafted file exhausts the heap and aborts guardian-service. Because a heap abort is not an exception, the `try/catch` that surrounds the parse cannot contain it.

### Detail

`common/src/xlsx/models/expression.ts`:

```ts
private parseRange(start: string, end: string): string[] {
    try {
        const _start = start.match(/(?<col>([A-Z]|[a-z])+)(?<row>(\d)+)/);
        const _end = end.match(/(?<col>([A-Z]|[a-z])+)(?<row>(\d)+)/);
        const result: string[] = [];
        ...
        const max = Math.max(startRow, endRow);
        const min = Math.min(startRow, endRow);
        for (let i = min; i <= max; i++) {
            result.push(`${startCol}${i}`);
        }
        return result;
    } catch (error) {
        throw new Error('Invalid range');
    }
}
```

`startRow` and `endRow` are `parseInt` of digits captured from the formula string. Nothing compares them against the worksheet's dimensions, against Excel's own 1,048,576-row ceiling, or against any limit.

`=SUM(A1:A99999999)` allocates 99,999,999 short strings — roughly 40-60 bytes each once V8 overhead is counted, so several GB for one range. A formula may contain several ranges and a sheet many formulas.

### Why the existing catch does not help

Both the inner `try` and the caller's `try { expression.parse(); ... }` in `common/src/xlsx/generate-blocks.ts` catch JavaScript exceptions. Exceeding the heap is not one — V8 aborts the process — so the guard that looks like it contains bad input contains only *well-formed* bad input.

### Reachability

The formula text comes from the uploaded sheet (`xlsx-to-json.ts`):

```ts
const formulae = worksheet.getFormulae(table.getCol(Dictionary.ANSWER), row);
if (formulae) {
    field.formulae = formulae;
```

and is parsed during import (`generate-blocks.ts`):

```ts
expressions.push(new Expression(field.name, field.formulae));
...
expression.parse();
```

[...]

PR #6856, opened 2026-09-07
fix(common): bound Excel formula range expansion

vshvets-bc:
Fixes #6855.

## Problem

`parseRange` expands a cell range into one string per row, with the row numbers taken from the **formula text of the uploaded file** and no bound of any kind:

```ts
const max = Math.max(startRow, endRow);
const min = Math.min(startRow, endRow);
for (let i = min; i <= max; i++) {
    result.push(`${startCol}${i}`);
}
```

`=SUM(A1:A99999999)` asks for 100 million short strings — several GB once V8 overhead is counted — and a sheet may carry many such formulas.

**The existing catch cannot help.** Exceeding the heap aborts the process rather than raising, so neither the `try` inside `parseRange` nor the caller's `try { expression.parse() }` contains it. The path is reachable on **preview**, before the user has committed to importing anything.

## Fix

Two caps, checked **before** anything is allocated:

- `MAX_RANGE_CELLS = 10_000` for a single range
- `MAX_EXPRESSION_CELLS = 50_000` across all ranges in one formula

The caller in `generate-blocks.ts` already records a failed parse as an `XlsxResult` validation error and continues, so an over-large range now reports as **invalid input** rather than terminating the service. The refusal keeps its own message instead of being rewritten to `'Invalid range'` by the blanket catch, so it doesn't look like a syntax error.

## The caps are a judgement call

Excel's 1,048,576-row ceiling is an upper bound, not a safe one, and real schema sheets reference field rows in the tens — so 10,000 is generous while staying orders of magnitude below anything dangerous. But if any legitimate workbook uses larger ranges, this rejects it as invalid input. Happy to set them wherever you prefer, or make them configurable.

## Tests

Eight cases in `common/tests/unit-tests/xlsx-expression/`, alongside the existing expression specs: ordinary ranges unchanged, exactly-at-cap allowed, one-beyond refused, the crafted range refused *without allocating*, the message explains why, a genuinely malformed range still reports `Invalid range`, the total cap trips across several individually-legal ranges, and a normal multi-range formula is untouched.

**They deliberately do not reproduce the crash.** A test that attempted the 100-million allocation would take the CI runner down with it, so they assert the bound instead — that the refusal happens before expansion, and fast.

Verified against this branch: 8 pass, and the 36 pre-existing expression tests still pass.

## Also worth checking, not in this PR

`xlsx-to-json.ts` iterates worksheets, rows and columns from the file with no ceiling on worksheet count either — same class of exposure on a different path. Left out deliberately rather than widening this change.

### DIFF
--- a/common/src/xlsx/models/expression.ts
+++ b/common/src/xlsx/models/expression.ts
@@ -1,5 +1,23 @@
 import * as mathjs from 'mathjs';
 
+/**
+ * Largest span a single range (A1:A9999) may expand to.
+ *
+ * parseRange builds one string per row, so an unbounded range from an uploaded
+ * file is an allocation the process cannot survive: `=SUM(A1:A99999999)` asks for
+ * 100 million strings. That is an OOM abort rather than an exception, so the
+ * try/catch around the parse cannot contain it.
+ *
+ * Excel's own ceiling of 1,048,576 rows is an upper bound, not a safe one.
+ */
+export const MAX_RANGE_CELLS = 10_000;
+
+/** Total cells one expression may reference across all of its ranges. */
+export const MAX_EXPRESSION_CELLS = 50_000;
+
+/** Raised when a formula asks for more cells than the caps above allow. */
+export class RangeTooLargeError extends Error { }
+
 export class Expression {
     public readonly name: string;
     public readonly formulae: string;
@@ -10,12 +28,15 @@ export class Expression {
     public validated: boolean;
     public transformed: string;
 
+    private cellCount: number;
+
     constructor(name: string, formulae: string) {
         this.name = name;
         this.formulae = formulae;
         this.symbols = new Set<string>();
         this.functions = new Map<string, string[]>();
         this.ranges = new Map<string, string[]>();
+        this.cellCount = 0;
     }
 
     public parse(): void {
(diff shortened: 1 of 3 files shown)


common/src/xlsx/models/expression.ts
```diff
@@ -1,5 +1,23 @@
...
+/**
+ * Largest span a single range (A1:A9999) may expand to.
+ *
+ * parseRange builds one string per row, so an unbounded range from an uploaded
+ * file is an allocation the process cannot survive: `=SUM(A1:A99999999)` asks for
+ * 100 million strings. That is an OOM abort rather than an exception, so the
+ * try/catch around the parse cannot contain it.
+ *
+ * Excel's own ceiling of 1,048,576 rows is an upper bound, not a safe one.
+ */
```
Pyatakov, line 6: Compact comment

common/src/xlsx/models/expression.ts
```diff
@@ -72,11 +93,28 @@ export class Expression {
...
+                );
+            }
+
             for (let i = min; i <= max; i++) {
                 result.push(`${startCol}${i}`);
             }
             return result;
         } catch (error) {
+            if (error instanceof RangeTooLargeError) {
+                throw error;
```
Pyatakov, line 110: Currently when `expression.parse()` throws a `RangeTooLargeError`, the catch block (`common/src/xlsx/generate-blocks.ts:L246-254`) discards it and reports a generic message:

```ts
} catch (error) {
    expression.validated = false;
    xlsxResult.addError({
        type: 'error',
        text: `Failed to parse formula (${expression.name}=${expression.formulae}).`,
        message: `Failed to parse formula (${expression.name}=${expression.formulae}).`,
        worksheet: xlsxSchema.worksheet.name
    }, null);
}
```

It would be nice to preserve it by appending `${error.message}` to both the `text` and `message` fields:

```ts
text: `Failed to parse formula (${expression.name}=${expression.formulae}). ${error.message}`,
message: `Failed to parse formula (${expression.name}=${expression.formulae}). ${error.message}`,
```

This way, when someone uploads a file with `=SUM(A1:A99999999)`, they'll see the helpful message `"Range A1:A99999999 spans 99999999 cells, more than the 10000 allowed."` instead of just `"Failed to parse formula."`, which makes the test's assertion "says why, rather than reporting a syntax error" actually hold for end users, not just in the unit test.

Pyatakov (changes requested):

vshvets-bc:
Done. Pushed in `bb2f42e5`.

You're right that the assertion held only inside the unit test - `generate-blocks.ts` discarded the error and reported the generic sentence, so nobody uploading a file ever saw which range was at fault. The cause is appended to both fields now:

```ts
const reason = `Failed to parse formula (${expression.name}=${expression.formulae}). ${error?.message ?? ''}`.trim();
```

`error?.message` with a trim, so a parse failure carrying no message still reads as a clean sentence rather than trailing whitespace.

Added `the reason survives into what generate-blocks reports`, asserting the composed message names `A1:A99999999` and the cap - so the claim is now pinned where it matters and not only at the throw site. 9 passing.

Comments compacted: the `MAX_RANGE_CELLS` block is 3 lines, the test header 6.

Pyatakov (approved):

Issue #6862, opened 2026-09-07
Profile policy keys: a second key for the same policy silently overwrites the first one's secret


vshvets-bc:
### Summary

The policy signing keys on the profile page are stored in two places that can disagree: one Mongo row per creation, but a single vault slot per `did` + `messageId`. Creating a second key for the same policy silently destroys the first one's secret.

A second, separate problem in the same area: deleting a key removes the row and leaves its secret in the vault. That one is noted at the end but not addressed by the PR, because it needs a decision rather than a patch.

### Defect: a second key for the same policy overwrites the first

`guardian-service/src/api/profile.service.ts`, `GENERATE_USER_KEYS`:

```ts
const item = await DatabaseServer.saveKey({
    messageId,
    owner: user.did
});
if (!key) {
    key = PrivateKey.generate().toString();
}
const wallet = new Wallet();
await wallet.setUserKey(
    user.did,
    KeyType.MESSAGE_KEY,
    `${user.did}#${item.messageId}`,
    key,
    msg?.user?.id
);
```

`saveKey` inserts unconditionally, with no check for an existing row for the same `messageId` and owner, so a second call adds a second row. The vault entry, however, is addressed by `${user.did}#${item.messageId}` and **does not include the row id**, so both rows point at one slot and the second `setUserKey` overwrites what the first stored.

The result is a keys list showing two entries for one policy where only the newer secret still exists. Anything signed or encrypted with the first key can no longer be resolved through the profile, and nothing in the UI says the older entry is dead.

`messageId` is free text in the create dialog, so reaching this needs no unusual client — entering the same policy message id twice is enough.

Note that the documented import flow is not affected: a recipient importing a shared key is a different DID, so their vault address differs. The collision is one user creating twice for the same policy.

### Steps to reproduce

1. On the profile page, create a policy key for message id `M`.
2. Create another key for the same message id `M`.
3. The list shows two entries. Only the second resolves; the first key's secret is gone.

### Expected

Either the second create is refused, or it updates the existing row — but the first secret must not be destroyed silently, since it cannot be recovered.

### Fix

PR follows: refuse the create when the owner already holds a key for that `messageId`. Refusing rather than replacing, because the previous secret is unrecoverable; rotation is delete-then-create, which frees the row and lets the next create take the slot.

### Separate problem, not addressed in the PR

[...]

PR #6863, opened 2026-09-07
fix(guardian-service): refuse a duplicate policy key instead of destroying the first

vshvets-bc:
Fixes #6862.

## Problem

`GENERATE_USER_KEYS` inserts a row per creation, but the vault slot it fills is addressed by `${user.did}#${item.messageId}` and carries **no row id**:

```ts
const item = await DatabaseServer.saveKey({ messageId, owner: user.did });
...
await wallet.setUserKey(user.did, KeyType.MESSAGE_KEY, `${user.did}#${item.messageId}`, key, ...);
```

A second create for the same policy therefore added a second row while `setUserKey` **overwrote the first row's secret**. The keys list showed two entries for one policy, only the newer one resolved, and nothing said the older was dead.

`messageId` is free text in the create dialog, so entering the same policy id twice is enough. (The documented import flow is unaffected — a recipient importing a shared key is a different DID, so their vault address differs.)

## Fix

Refuse the create when the owner already holds a key for that `messageId`, rather than replacing it silently, because the previous secret cannot be recovered. Rotation is delete-then-create, which frees the row and lets the next create take the slot.

## Tests

Four cases in `guardian-service/tests/profile-policy-keys.test.mjs`, using the existing `_handler-harness`.

They assert the **lookup filter**, not only the refusal — scoping to `messageId` *and* `owner` is what keeps one user's key from blocking another's, and a key for a different policy from blocking this one. A test that only checked "the second create failed" would pass against a filter scoped too broadly.

Verified against this branch: 4 pass. Against the unpatched source **2 fail** — the other two are the deliberate guards that ordinary creates still work.

## A separate problem I deliberately did not fix

`DELETE_USER_KEYS` removes the Mongo row and nothing else, so a key the user believes they revoked is still resolvable by anything addressing the vault directly.

There is no delete path to call: `Wallet` exposes only `getKey`/`setKey`, because `SecretManagerBase` declares only `getSecrets`/`setSecrets`, and none of the backends under `common/src/secret-manager/` mentions delete, destroy or purge.

Adding one means extending that interface across all of them, where "delete" means materially different things:

| backend | semantics |
|---|---|
| Vault KV v2 | `delete` soft-deletes a version; `destroy` is permanent |
| AWS Secrets Manager | schedules deletion with a minimum recovery window |
| Azure Key Vault | soft-deletes, then needs a separate purge |
| GCP Secret Manager | destroys a version |

Getting it wrong leaves the secret recoverable — not a fix — or destroys it irrecoverably. That reads as a maintainer decision about secret lifecycle rather than something to settle inside this change, so it is described in the issue and left out here. Happy to implement whichever semantics you prefer.

### DIFF
--- a/guardian-service/src/api/profile.service.ts
+++ b/guardian-service/src/api/profile.service.ts
@@ -631,23 +631,15 @@ export function profileAPI(logger: PinoLogger) {
                 const { messageId, user } = msg;
                 let { key } = msg;
 
-                /*
-                 * One row per creation, but a single vault slot per
-                 * `${did}#${messageId}`: the address below carries no row id. A
-                 * second create for the same policy would add a second row while
-                 * setUserKey overwrote the first row's secret, leaving a keys list
-                 * where the older entry is dead and nothing says so.
-                 *
-                 * Refused rather than silently replaced, because the previous
-                 * secret cannot be recovered. Rotation is delete-then-create.
-                 */
+                // one vault slot per `did#messageId`, so a second create would overwrite
+                // the first row's secret with no way to recover it
                 const existing = await DatabaseServer.getKeys({
                     messageId,
                     owner: user.did
                 });
                 if (existing?.length) {
                     return new MessageError(
-                        `A key for this policy already exists. Delete it before creating another.`,
+                        'A key for this policy already exists. Delete it before creating another.',
                         409
                     );
                 }
--- a/guardian-service/src/api/profile.service.ts
+++ b/guardian-service/src/api/profile.service.ts
@@ -630,6 +630,28 @@ export function profileAPI(logger: PinoLogger) {
                 }
                 const { messageId, user } = msg;
                 let { key } = msg;
+
+                /*
+                 * One row per creation, but a single vault slot per
+                 * `${did}#${messageId}`: the address below carries no row id. A
+                 * second create for the same policy would add a second row while
+                 * setUserKey overwrote the first row's secret, leaving a keys list
+                 * where the older entry is dead and nothing says so.
+                 *
+                 * Refused rather than silently replaced, because the previous
+                 * secret cannot be recovered. Rotation is delete-then-create.
+                 */
+                const existing = await DatabaseServer.getKeys({
+                    messageId,
+                    owner: user.did
+                });
+                if (existing?.length) {
+                    return new MessageError(
+                        `A key for this policy already exists. Delete it before creating another.`,
+                        409
+                    );
+                }
+
                 const item = await DatabaseServer.saveKey({
                     messageId,
                     owner: user.did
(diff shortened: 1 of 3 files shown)


guardian-service/src/api/profile.service.ts
```diff
@@ -630,6 +630,28 @@ export function profileAPI(logger: PinoLogger) {
...
+                /*
+                 * One row per creation, but a single vault slot per
+                 * `${did}#${messageId}`: the address below carries no row id. A
+                 * second create for the same policy would add a second row while
+                 * setUserKey overwrote the first row's secret, leaving a keys list
+                 * where the older entry is dead and nothing says so.
+                 *
+                 * Refused rather than silently replaced, because the previous
+                 * secret cannot be recovered. Rotation is delete-then-create.
+                 */
```
Pyatakov: Compact comment

Pyatakov (changes requested):

guardian-service/src/api/profile.service.ts
```diff
@@ -630,6 +630,28 @@ export function profileAPI(logger: PinoLogger) {
...
+                const existing = await DatabaseServer.getKeys({
+                    messageId,
+                    owner: user.did
+                });
+                if (existing?.length) {
+                    return new MessageError(
+                        `A key for this policy already exists. Delete it before creating another.`,
+                        409
```
Pyatakov, line 643: The route in `api-gateway/src/api/service/profile.ts:697` declares only 422/500; add `@ApiConflictResponse`, and its description still presents the import flow as unconditional.

guardian-service/src/api/profile.service.ts
```diff
@@ -630,6 +630,28 @@ export function profileAPI(logger: PinoLogger) {
...
+                 * Refused rather than silently replaced, because the previous
+                 * secret cannot be recovered. Rotation is delete-then-create.
+                 */
+                const existing = await DatabaseServer.getKeys({
+                    messageId,
+                    owner: user.did
+                });
+                if (existing?.length) {
+                    return new MessageError(
+                        `A key for this policy already exists. Delete it before creating another.`,
```
Pyatakov: Template literal with no interpolation; a plain quoted string serves better here.

vshvets-bc:
All three done. Pushed in `63ff71aa`.

**`@ApiConflictResponse` added** to `POST /profiles/keys`, with `ConflictErrorDTO` and the real message as the example, matching the shape already used in `account.ts:150`.

**The description no longer reads as unconditional.** It now ends with the refusal and the reason:

> Either way the call is refused with **409** when this DID already holds a key for that `messageId`: one vault slot is addressed per `did#messageId`, so a second key would overwrite the first with no way to recover it. Delete the existing key before creating another.

**Template literal → plain string.**

Comments compacted too: the block above the check is two lines, the test header four. 4 passing.

Pyatakov (approved):

PR #6868, opened 2026-09-08
feat: 6734 Allow template authors to highlight customizable areas

dariy-rised:
**Description**:
* Added schema template guidelines support.

**Related issue(s)**:

Resloves [#6734](https://github.com/hashgraph/guardian/issues/6734)


frontend/src/app/views/schemas-configuration/schemas-configuration.component.ts
```diff
@@ -1024,6 +1032,30 @@ export class SchemasConfigurationComponent implements OnInit, OnDestroy {
...
 
+    public setSelectedSchemaGuidelines(guidelines: string): void {
+        if (this.isTemplateReadonly) {
+            return;
+        }
+        const config = this.ensureSelectedSchemaConfig();
+        if (!config) {
+            return;
+        }
+        config.guidelines = guidelines || '';
```
Pyatakov: Clearing the textarea persists `guidelines: ""` rather than removing the key, so an emptied note travels into the stored config, into every policy snapshot, and into `templateStateHash` (which drives "template updated" detection). Suggested a `delete` on blank.

frontend/src/app/views/schemas-configuration/schemas-configuration.component.html
```diff
@@ -385,13 +385,32 @@
                                     </span>
                                     <span class="st-toggle" [class.active]="canAddCustomFieldsToSelectedSchema"></span>
                                 </button>
+                                <div class="st-guidelines-editor">
+                                    <label class="st-guidelines-label">Guidelines</label>
```
Pyatakov: The `Guidelines` label is a sibling of the textarea with no `for`, so it doesn't focus the field and the control is announced unlabelled. Same at line 717.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6890, opened 2026-09-10
feat: 6712 Template schema order

dariy-rised:
**Description**:
* Added the ability to feature schemas using a template
* Added ordering by featured schemas first

**Related issue(s)**:

Resolves [#6712](https://github.com/hashgraph/guardian/issues/6712)

### DIFF
--- a/common/src/entity/schema.ts
+++ b/common/src/entity/schema.ts
@@ -24,6 +24,10 @@ import { BaseEntity } from '../models/index.js';
     properties: ['defs'],
     name: 'defs_index',
 })
+@Index({
+    properties: ['templateFeatured', 'createDate'],
+    name: 'templateFeatured_createDate_index',
+})
 export class Schema extends BaseEntity implements ISchema {
     /**
      * Schema uuid
@@ -180,6 +184,14 @@ export class Schema extends BaseEntity implements ISchema {
         index: true
     })
     templateSchemaId?: string;
+
+    /**
+     * Denormalized copy of config.schemas[id].featured from the owning schema template,
+     * kept in sync by schema-template.service.ts. SchemaTemplate.config itself is not
+     * queryable (GridFS-backed), so this is what grid sort / cross-context display use.
+     */
+    @Property({ nullable: true })
+    templateFeatured?: boolean;
 
     /**
      * Schema code version
--- a/common/src/entity/schema.ts
+++ b/common/src/entity/schema.ts
@@ -25,9 +25,9 @@ import { BaseEntity } from '../models/index.js';
     name: 'defs_index',
 })
 @Index({
-    properties: ['templateFeatured', 'createDate'],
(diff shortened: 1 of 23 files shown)


guardian-service/package.json
```diff
@@ -67,8 +67,8 @@
         "lint": "tslint --config ../tslint.json --project .",
         "start": "node dist/index.js",
         "test": "mocha tests/**/*.test.mjs --reporter mocha-junit-reporter --reporter-options mochaFile=../test_results/guardian-service.xml --timeout 10000 --exit",
-        "test:local": "mocha tests/**/*.test.mjs --exit",
-        "test:stability": "mocha tests/stability.test.mjs --exit"
+        "test:local": "mocha tests/**/*.test.mjs --timeout 10000 --exit",
+        "test:stability": "mocha tests/stability.test.mjs  --exit"
```
Pyatakov: double space in test:stability, the line's only change from base

docs/guardian/workspace/schema-templates/README.md
```diff
@@ -54,6 +55,9 @@ The configuration controls:
   * `schemaSettingsLocked: false` allows changes. `true` prevents them.
 * **Can add custom fields**: lets policy developers add custom fields to the schema.
   * `customFieldsLocked`: `false` allows custom fields, while `true` prevents them.
+* **Featured**: marks a template schema so it stands out in schema lists.
+  * `featured: true` shows a filled star next to the schema and sorts it first on the main Schemas grid. `false` (the default) does neither.
```
Pyatakov: "sorts it first on the main Schemas grid" overstates it; the comparator sorts by topic first

common/src/entity/schema.ts
```diff
@@ -24,6 +24,10 @@ import { BaseEntity } from '../models/index.js';
     properties: ['defs'],
     name: 'defs_index',
 })
+@Index({
+    properties: ['templateFeatured', 'createDate'],
+    name: 'templateFeatured_createDate_index',
+})
```
Pyatakov: This index holds only the sort fields, so it can give `GET_SCHEMAS_V2` the order but cannot narrow its filter (`owner`, `readonly`, `system`, usually `category`/`topicId`).
Suggest:

```suggestion
@Index({
     properties: ['owner', 'category', 'templateFeatured', 'createDate'],
     name: 'owner_category_templateFeatured_createDate_index',
 })
 ```

 Otherwise the planner picks between walking this index in order and fetching/discarding every non-matching doc, or narrowing on `topicId` and doing a blocking in-memory sort, so the index is maintained on every write without removing either cost. Dropping it and leaving the sort unindexed (as `createDate` already was) is also fine.

Pyatakov (changes requested):

Pyatakov (approved):

PR #6908, opened 2026-09-11
chore: update schema excel template with the missing field types

ihar-tsykala:
**Description**:

Fixes #

**Checklist**


swidnikk:
Couple things I've noticed

# Enum worksheet: remove the Schema Name Column
The Enums sheet's "Schema name" column is dead weight — readSharedEnumSheet/resolveSharedEnumColumns in common/src/xlsx/xlsx-to-json.ts match headers by name (Enum Name, Loaded to IPFS, Value) and never read a "Schema name" column; enums are matched globally by name, not scoped per schema. Keeping it in the template implies authors should fill it in, but it's just unused overhead. Suggest dropping it from the template/README so the example matches what the parser actually consumes: Enum Name | Loaded to IPFS | Value.

# Update the guidance for Enum's 
In the Enums tab, the three columns are: Enum Name | Loaded to IPFS | Value

For the first value of each group: fill in Enum Name, Loaded to IPFS. Leave those three columns blank on subsequent value rows for the same group. 

Dropdown list — values defined in the Enums tab. The field's Parameter must exactly match the "Enum Name" column.

Parameter column: Type-dependent: enum name (Enum)

image


# Description Examples
Suggest to update all description fields to represent what might be presented to the user on a form such as the following, 
image

ihar-tsykala:
Hey @swidnikk, could you check it again please?

swidnikk:
@ihar-tsykala looks great

swidnikk (approved):

PR #6922, opened 2026-09-15
feat: 6843 Path Editing, Improve Mathblock Output

dariy-rised:
**Description**:
* Added path picker for math block outputs

**Related issue(s)**:

Resolves [#6843](https://github.com/hashgraph/guardian/issues/6843)


frontend/src/app/modules/policy-engine/dialogs/math-editor-dialog/math-editor-dialog.component.html
```diff
@@ -604,21 +604,41 @@
                         <div class="formula-body" [attr.error]="!item.validField">
                           <div class="guardian-input-container">
                             <label class="form-label">Output</label>
-                            <div class="formula-link" (click)="onOutputLink(item)">
-                              @if (item.field) {
-                                <div class="formula-link-value">
-                                  <div class="formula-link-value__entity">{{getEntityName('output', null)}}:</div>
-                                  <div class="formula-link-value__item">{{getFieldName('output', null, item.field)}}</div>
-                                </div>
+                            <div class="path-input-wrapper">
```
Pyatakov: The output picker is a verbatim copy of the input picker at lines 146-180 (only the callback and the `'output'` argument differ); worth extracting into an `ng-template` so the two can't drift.

Pyatakov (changes requested):

Pyatakov (approved):

Issue #6925, opened 2026-09-17
TopicListener replays and re-executes the entire topic on every restart; resume counter advances before the handler succeeds


vshvets-bc:
## Summary

Two defects in `TopicListener` (`common/src/topic-listener/topic-listener.ts`) mean a listener
restart **replays and re-executes its entire topic history**, and the resume counter can claim
messages were processed when they were not.

Found while investigating a production listener that stopped delivering messages for seven weeks.
That particular wedge was caused by a downstream modification and is not reproducible on
`develop`, but these two are present here and reproduce independently.

## 1. A restarted listener replays the whole topic

`_start()` decides whether to send a resume position using a **truthiness** test:

```ts
if (this._startNumber) {
    options.index = this._startNumber;
}
```

and `topic-listener-service/src/api/listener-service.ts` defaults a missing index:

```ts
if (index === null) {
    index = -1;
}
```

So any listener whose `_startNumber` is falsy asks for `-1` and the service replays from the
start of the topic. Two ways to hit it:

- a fresh process, where `_startNumber` is `null` — i.e. **every restart**
- a listener legitimately sitting at index `0`

The replayed messages are not skipped, they are **re-executed**. `sendData` compares
`index > this._startNumber`, and with `_startNumber === null` that coerces to `index > 0`, which
is true for every message in the topic. So each restart re-runs every handler the listener has
ever run.

For an actions topic this means duplicated side effects on every restart, and a load spike
proportional to topic history.

**Fix:** `if (typeof this._startNumber === 'number')`.

## 2. The resume counter advances before the handler succeeds

```ts
if (index > this._startNumber) {
    this._startNumber = index;          // assigned first
    await this._observable(message);    // may throw, or never settle
}
await TopicListener.channel.publish(`${CONFIRM}.${this._listenerId}`, index);
```

`_startNumber` is the position a restart resumes from and the value the `>` test above compares
against, but it is committed before the handler runs. If the handler throws, the outer `catch`
swallows it and returns `false` — **skipping the confirm** — while the counter keeps the bumped
value. If the handler never settles, nothing after the `await` runs at all.

Either way the listener records a message as processed that was not. On the next delivery the
`>` test is false, so the message is silently skipped rather than retried.

**Fix:** assign after the handler resolves, and wrap the call in an inner `try/catch` so a failed
message still publishes its confirm instead of relying on the next delivery to unblock the
sender's single-message window.

## Scope

[...]

PR #6926, opened 2026-09-17
fix(common): resume a topic listener instead of replaying its whole topic

vshvets-bc:
Fixes #6925.

Two defects in `TopicListener`, both reachable on any listener restart.

## 1. A restart replays and re-executes the whole topic

`_start()` used a truthiness test to decide whether to send a resume position:

```ts
if (this._startNumber) { options.index = this._startNumber; }
```

A falsy `_startNumber` is dropped, and `listener-service.ts` defaults the missing index to `-1`,
replaying from the start of the topic. That covers every fresh process (`_startNumber` is `null`)
and any listener legitimately at index `0`.

The replayed messages are **re-executed**, not skipped: `sendData` tests
`index > this._startNumber`, and with a null counter that coerces to `index > 0` — true for every
message. So each restart re-ran every handler the listener had ever run.

`typeof this._startNumber === 'number'` instead.

## 2. The resume counter advanced before the handler succeeded

```ts
if (index > this._startNumber) {
    this._startNumber = index;          // committed first
    await this._observable(message);    // may throw, or never settle
}
await TopicListener.channel.publish(...);
```

`_startNumber` is what a restart resumes from and what the `>` test compares against. If the
handler throws, the outer `catch` returns `false` and skips the confirm while the counter keeps
the bumped value; if it never settles, nothing after the `await` runs. Either way the listener
records a message as processed that was not, and the next delivery silently skips it.

Now assigned after the handler resolves, with an inner `try/catch` so a failed message still
publishes its confirm rather than relying on redelivery to release the sender's single-message
window.

## Notes

- **No change on the success path** — same handler, run once, same confirm.
- Both changes are in `common`, so every service constructing a `TopicListener` benefits.
- Found while investigating a production listener that stopped delivering for seven weeks. That
  specific wedge came from a downstream modification and is **not** reproducible on `develop`;
  these two are, and are independent of it.


common/src/topic-listener/topic-listener.ts
```diff
@@ -133,7 +141,9 @@ export class TopicListener {
             if (this._name) {
                 options.name = this._name;
             }
-            if (this._startNumber) {
+            // typeof, not truthiness: index 0 is a valid position, and a dropped index
+            // makes the service default to -1 and replay the whole topic.
```
Pyatakov:
```suggestion
            // typeof, not truthiness, so a start position of 0 is still sent. A listener
            // with no start position (null) sends none and resumes from the service's
            // persisted index.
```

common/src/topic-listener/topic-listener.ts
```diff
@@ -103,8 +103,16 @@ export class TopicListener {
...
+                try {
+                    await this._observable(message);
+                    // Advance only on success: a handler that throws or never settles has
+                    // not processed this message, and the counter is what a restart resumes
+                    // from and what the `index > _startNumber` test above compares against.
+                    this._startNumber = index;
+                } catch (error) {
+                    // Confirm below regardless, so one failed message cannot hold the
+                    // sender's single-message window shut.
+                }
```
Pyatakov, line 112:
```suggestion
                try {
                    await this._observable(message);
                    this._startNumber = index;
                } catch (error) {
                    // The message is confirmed below and dropped by the service: it is not retried.
                    await this.sendError(error);
                }
```

Pyatakov (changes requested):

Pyatakov (approved):

Issue #6921, opened 2026-09-14
Add a conditions lock option to Schema Templates


dariy-rised:
### Problem description
Template config has no lock for conditions.

### Requirements
- Add `conditionsLocked` toggle (config + UI), allowing all condition edits; adding a new field in a branch still needs `customFieldsLocked=false`.
- Per-field lock toggles for fields inside conditions, like regular fields.
- Fix template custom conditions/fields updating.

### Definition of done
`conditionsLocked` allows edit the Conditions tab as specified, and template updates preserve or flag policy-authored condition work instead of silently dropping it.

### Acceptance criteria
- [ ] `conditionsLocked` toggles per schema and persists.
- [ ] Locked = no condition add/edit/remove; unlocked + `customFieldsLocked` = edit only, no new fields.
- [ ] Per-field lock toggles work inside conditions.
- [ ] Resolving condition changes conflicts when updating template.
- [ ] Tests cover the above in the existing `schema-template-*` unit test files.

PR #6929, opened 2026-09-17
feat: 6921 Template conditions

dariy-rised:
**Description**:
  * Add a per-schema conditions lock toggle to Schema Templates
  * Add conflict resolution for condition changes on template update
  * Show template config (lock toggle) changes in the update diff

**Related issue(s)**:

Resolves [#6921](https://github.com/hashgraph/guardian/issues/6921)


frontend/src/app/views/schemas-configuration/schemas-configuration.component.html
```diff
@@ -1948,20 +1961,25 @@
...
+                                [disabled]="!canChangeConditionsForSelectedSchema"
                                 (click)="setConditionOperator(cond, 'OR')">IF ANY</button>
                         </div>
                         <span class="sc-cond-type-note">
                             @if (getIfOperator(cond) === 'SINGLE') { — single condition }
                             @else if (getIfOperator(cond) === 'AND') { — all must match }
                             @else { — any must match }
                         </span>
-                        <button class="sc-cond-block-del" title="Remove condition" (click)="removeConditionAt(ci)">
+                        <button class="sc-cond-block-del" title="Remove condition"
```
Pyatakov: `removeConditionAt` now returns early when a THEN/ELSE field is template-locked, but this button is only disabled on `!canChangeConditionsForSelectedSchema`, so the trash icon looks active and nothing happens on click. Disable it (and set a title) using the same check, e.g. a `conditionHasLockedField(cond)` helper shared with `removeConditionAt`.

guardian-service/src/api/schema-template.service.ts
```diff
@@ -1555,14 +1897,67 @@ function applySchemaDocumentSettings(document: any, name: string, description: s
...
+    const sourceConditions = sourceParsed.conditions || [];
+    const placements = analyzeConditionFieldPlacements(previousConditions, sourceConditions, custom);
+
+    // dropFieldNames also excludes the field from the root-properties merge below, so an
+    // orphaned condition the user didn't choose to keep is dropped entirely, not left
+    // behind as a detached root field.
+    const dropFieldNames = new Set<string>();
+    const carryOverConditionIndices = new Set<number>();
+    const matchedIndexByOldIndex = new Map<number, number>();
+    if (!schemaConfig.customFieldsLocked) {
```
Pyatakov: This block only checks `customFieldsLocked`; `conditionsLocked` is never read during the update. 

Scenario: a policy adds its own condition (no template trigger), then the template's next version turns on `conditionsLocked`. On update the condition is carried over by `carryOverCustomCondition` (and policy-added cross-schema targets on matched conditions by `restoreConditionCrossTargets`). The policy then ends up with a non-template condition that the UI and `validateTemplateSchemaUpdateByConfig` both block from being edited or removed.

If locked means "the conditions are the template's", skip the carry-over/cross-target restore when `schemaConfig.conditionsLocked` is set (custom fields in matched branches can still be restored), and apply the same gate to the preview block at line 1655 so the dialog doesn't offer "Keep as custom" for a condition that will be locked. If carrying them over is intended, it's worth a comment here, because the result can't be edited afterwards.

guardian-service/src/api/schema-template.service.ts
```diff
@@ -1582,7 +1977,33 @@ function preparePolicySchemaUpdate(
...
+                placement.matchedIndex,
+                placement.branch,
+                placement.field.name
+            );
+        }
+    }
+    // Unconditional per matched pair, not per field placement - a condition can hold a
+    // policy-added cross-schema target with no custom field alongside it at all.
+    for (const [oldIndex, newIndex] of matchedIndexByOldIndex) {
+        restoreConditionCrossTargets(target.document, previousDocument, oldIndex, newIndex);
```
Pyatakov: `restoreConditionCrossTargets` copies every cross-schema target from the old policy document into the rebuilt one. The old document also holds the targets that came from the previous template version, so a target the template author deletes comes back.

Repro: v1 has `IF type = "solar" THEN require Address.zip`. The policy applies v1, then v2 is published with that requirement removed. After Update, `zip` is still required, and because every later update restores from the previous policy document, no template version can remove it.

Only policy-added targets should be carried over. `context.snapshot` already holds the template schema the policy was built from, so compare against the snapshot condition with the same signature and restore only targets that are not in it. The doc comment above calls the unconditional merge safe. Please update it either way.

guardian-service/src/api/schema.service.ts
```diff
@@ -70,6 +70,12 @@ function getSchemaFields(schema: ISchema): SchemaField[] {
     return new Schema(schema, true).fields || [];
 }
 
+function getConditionsHash(schema: ISchema): string {
+    return SchemaHelper.stableStringify(
+        SchemaHelper.cloneSchemaRuntimeValue(new Schema(schema, true).conditions || [])
+    );
+}
```
Pyatakov: `getConditionsHash` stringifies the parsed conditions, and those carry the full `SchemaField` objects (`ifCondition.field`, `thenFields`, `elseFields`) – title, description, required, enum and so on. With `conditionsLocked: true`, any edit to a field that is a trigger or sits in a THEN/ELSE branch changes the hash, even when the field itself is unlocked (`locked: false`) or is a custom field that `customFieldsLocked: false` allows. The save then fails with "Conditions for ... are locked" although no condition was touched.

This also defeats the "per-field lock toggles work inside conditions" requirement in #6921 whenever conditions are locked. Hash the condition's structure instead of its fields – combinator, each predicate's field identity (`templateFieldId` or name) + `fieldValue`, branch field names, and target paths.

Suggested:
```ts
function getConditionsHash(schema: ISchema): string {
    const id = (f: any) => f?.templateFieldId || f?.name;
    const preds = (ic: any) => ic?.AND ?? ic?.OR ?? (ic ? [ic] : []);
    return SchemaHelper.stableStringify((new Schema(schema, true).conditions || []).map((c: any) => ({
        op: c.ifCondition?.AND ? 'AND' : c.ifCondition?.OR ? 'OR' : 'SINGLE',
        if: preds(c.ifCondition).map((p: any) => [id(p.field), p.fieldPath, p.fieldValue]),
        then: (c.thenFields || []).map(id),
        else: (c.elseFields || []).map(id),
        thenTargets: (c.thenTargets || []).map((t: any) => t.fieldPath),
        elseTargets: (c.elseTargets || []).map((t: any) => t.fieldPath)
    })));
}
```

Pyatakov (changes requested):

guardian-service/src/api/schema-template.service.ts
dariy-rised: > (custom fields in matched branches can still be restored)

If we have `template_1` with `condition_1` and `conditionsLocked = false`, and policy added custom field to that condition `custom_field_1` and then we want to update template by `template_2` where we have same `condition_1` but `conditionsLocked = true`, do we want to just delete `custom_field_1` and show it in preview window, or what do you mean about restoring custom fields in matched branches, put them to schema root fields?

Pyatakov: Not the root, moving it there would make the field always visible and change the form's behavior. I meant keeping it in condition_1's branch, which is what the code does now, but you're right that it doesn't fit: under `conditionsLocked` the field would then be stuck – it can't be removed from the branch or deleted without changing the conditions, which the lock rejects.

I'd go with your first option: when `conditionsLocked` becomes true, drop policy-added fields and cross-schema targets from conditions and list them in the preview as removals, the same way `customFieldsLocked` handles custom fields. The template author turning on the lock is the signal that the conditions are now the template's. The same rule then covers fully custom conditions and the keep-as-custom offer for orphaned ones.

PR #6936, opened 2026-09-18
feat: 6687 schema conditions array fields

dariy-rised:


common/src/xlsx/json-to-xlsx.ts
```diff
@@ -652,20 +652,25 @@ export class JsonToXlsx {
                 throw new Error(`Condition refers to unknown field "${sub.field?.name}".`);
             }
             const v = valueToFormula(sub.fieldValue);
+            const isArrayField = !!(sub.field?.isArray && !sub.field?.isRef);
+
+            if (sub.comparator === 'contains' && isArrayField) {
+                return `ISNUMBER(FIND(","&${v}&",", ","&${f.name}&","))`;
```
Pyatakov, line 658: The `contains` export formula `ISNUMBER(FIND(...))` only matches cells joined by bare commas. The import side (`xlsxToPresetArray`) also accepts `a, b` and quoted items like `"x,y"`. With `a, b` in the cell, `contains 'b'` is FALSE in the workbook but true in the imported schema. Also, the `equals` export on array fields is still `EXACT(Gn,v)`, which is only true for a one-element cell, not "every element equals".

interfaces/src/helpers/schema-helper.ts
```diff
@@ -706,6 +706,91 @@ export class SchemaHelper {
         }
     }
 
+    /**
+     * Loose scalar equality shared by server-side condition validation and (via
+     * `testPredicateValue`) array comparators. Numeric strings compare numerically; everything
+     * else falls back to a trimmed string comparison. Deliberately has no date-parsing branch -
+     * the form's `equalsLoosely` has one this does not, and unifying them is a separate decision
+     * (see docs/array-condition-operators-design.md, section 3.6), not something to fold in here.
```
Pyatakov, line 714: The JSDoc points to `docs/array-condition-operators-design.md`, which doesn't exist in the repo.

indexer-interfaces/src/interfaces/schema/schema-condition.interface.ts
```diff
@@ -16,6 +16,13 @@ export interface SchemaCondition {
          * field value
          */
         fieldValue: string;
+        /**
+         * comparator - absent means 'equals'. Explicit 'equals' on an array field means
+         * "each element equals" (the array analogue of '='), distinguished from an absent
+         * comparator (a legacy predicate predating array comparators, which stays dead/
+         * never-matches, unmigrated).
+         */
+        comparator?: 'contains' | 'equals';
```
Pyatakov, line 25: Comment still describes the discarded design, where an array predicate with no comparator "stays dead". That contradicts the `interfaces` doc and what the code actually does.

Pyatakov (changes requested):

Pyatakov (approved):

Issue #4531, opened 2025-01-08
Not able to revert back after selecting a role in the policy workflow


saharshkhicha18:
### Problem description
In the policy workflow's `Roles` block, if the user selects wrong role by mistake then there is no way for the user to go back and change the role. I have tried implementing a few ways in which the assigned role can be reverted but it does not seem to work.

### Step to reproduce
1. Go to a policy which has `Roles` block configured
<img width="1397" alt="image" src="https://github.com/user-attachments/assets/00b0bbc2-27ae-4abe-ab7d-458ea05edc1f" />

2. Run/Dryrun the policy and select a role
<img width="955" alt="image" src="https://github.com/user-attachments/assets/a921375e-47d2-4008-b0e1-fd9e5400631b" />

3. Once selected you will see the next screen and there is no way to revert the role
<img width="1593" alt="image" src="https://github.com/user-attachments/assets/46840909-ca13-4c1d-8203-7072dca957db" />

### Expected behavior
Ideally I want the user the have an option for them to go back and be able to select a different user role. Is it possible with the current infrastructure and if so then how can it be implemented?

PR #6941, opened 2026-09-20
Fix/4531 role reselection

Eclipse1704:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Eclipse1704:
This PR addresses #4531 by allowing users to return to the role-selection state after choosing a role by mistake.

The implementation reuses Guardian’s existing active/inactive group mechanism instead of deleting historical USER_ROLE records/VCs, and includes regression tests for clearing and reactivating role selections.

This is submitted for the #4531 bounty. Happy to address any maintainer feedback or additional test requirements.

Pyatakov:
@Eclipse1704 Thanks for the PR. The bounty program is currently inactive, so this contribution isn't eligible for a bounty.

The fix also doesn't solve #4531. In role-only policies, "Change role" only marks the user's existing role records inactive and doesn't remove them. When the user picks a role again, the Roles block creates a new record with a new UUID and publishes another USER_ROLE VC, and the old record stays. The user can then switch back to it from the group picker. So instead of reverting a wrong choice, a user can collect every role in the policy (for example Project_Proponent and VVB) and switch between them, which would let them validate their own project. The `getGroupTemplates` check you removed from `selectGroup` was what prevented this for policies without groups. A correct fix needs to replace the previous role, by removing or revoking the old record and its VC, rather than adding a new one next to it.

